Issue #315, opened 2013-01-22
Thrift tests fails randomly on Travis-CI


munkie:
Sometimes thrift tests fails on Travis-CI with foloowing messages:
```
1) Elastica\Test\Transport\ThriftTest::testSearchRequest with data set #0 (array('localhost', 9500, 'Thrift'))
Elastica\Exception\ClientException: No enabled connection

/home/travis/builds/ruflin/Elastica/lib/Elastica/Client.php:323
/home/travis/builds/ruflin/Elastica/lib/Elastica/Client.php:449
/home/travis/builds/ruflin/Elastica/lib/Elastica/Client.php:467
/home/travis/builds/ruflin/Elastica/lib/Elastica/Index.php:403
/home/travis/builds/ruflin/Elastica/lib/Elastica/Index.php:132
/home/travis/builds/ruflin/Elastica/lib/Elastica/Index.php:183
/home/travis/builds/ruflin/Elastica/test/lib/Elastica/Test/Transport/ThriftTest.php:40
/home/travis/builds/ruflin/Elastica/vendor/phpunit/phpunit/PHPUnit/Framework/TestCase.php:969
/home/travis/builds/ruflin/Elastica/vendor/phpunit/phpunit/PHPUnit/Framework/TestCase.php:824
/home/travis/builds/ruflin/Elastica/vendor/phpunit/phpunit/PHPUnit/Framework/TestResult.php:648
/home/travis/builds/ruflin/Elastica/vendor/phpunit/phpunit/PHPUnit/Framework/TestCase.php:769

2) Elastica\Test\Transport\ThriftTest::testSearchRequest with data set #1 (array('localhost', 9500, 'Thrift', array(false, 10000, 20000)))
Elastica\Exception\ClientException: No enabled connection

/home/travis/builds/ruflin/Elastica/lib/Elastica/Client.php:323
/home/travis/builds/ruflin/Elastica/lib/Elastica/Client.php:449
/home/travis/builds/ruflin/Elastica/lib/Elastica/Client.php:467
/home/travis/builds/ruflin/Elastica/lib/Elastica/Index.php:403
/home/travis/builds/ruflin/Elastica/lib/Elastica/Index.php:132
/home/travis/builds/ruflin/Elastica/lib/Elastica/Index.php:183
/home/travis/builds/ruflin/Elastica/test/lib/Elastica/Test/Transport/ThriftTest.php:40
[...]

PR #337, opened 2013-02-23
Connection exceptions in Client exception

munkie:
As we discussed in #315, i've added connection exceptions to client exception so there was possibility to get more verbose information why request has failed.

Also added abstract *getDsn()* method to *AbstractTransport* and implemented it in all transports. The purpose if this method to return human readable info about transport and connection it uses like: protocol, host, port and path. For Http transport it would look like this **http://localhost:9200/path** and for Thrift protocol like this: **thrift://localhost:9500**. It is used in client connection exception message for easier understanding what connection caused error.

### DIFF
--- a/lib/Elastica/Client.php
+++ b/lib/Elastica/Client.php
@@ -586,28 +586,36 @@ class Client
      */
     public function request($path, $method = Request::GET, $data = array(), array $query = array())
     {
-        $connection = $this->getConnection();
-        try {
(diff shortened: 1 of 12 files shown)


ruflin:
@munkie Long time no see :-) Quite and old PR, pitty we didn't get it in on time. I should probably have merged it an modify the changefile myself.

munkie:
@ruflin Was checking pending pull requests list, and found this one)

PR #342, opened 2013-02-26
Fix server cluster configuration

gigablah:
This fixes the `Elastica\Client::_initConnections` method so that the following code in the documentation works:

```php
$elasticaClient = new \Elastica\Client(array(
    'servers' => array(
        array('host' => 'localhost', 'port' => 9200)
        array('host' => 'localhost', 'port' => 9201)
    )
));
```

Before, the code was referencing an uninitialized variable (`$_config`), and it only assumed a single connection.


ruflin:
Closed as old documentation was removed and connections array should be used.

PR #723, opened 2014-11-13
Timeout considered in milliseconds

zedfmario:
## Goal
Allowing timeouts in milliseconds

## Need
For a high performance system, I needed a connection to elasticsearch, that couldn't last more that 250ms. If that case occurs, there's a fallback in memory. 

## Problem
There's no option for curlopt_timeout_ms developed.

## Solution
A new parameter on the connection that determines whether the timeout given should be considered as milliseconds or not.

## Tests
I run the tests and everything worked fine. 

## Feedback
Please, contact me in case of any doubt.

### DIFF
--- a/changes.txt
+++ b/changes.txt
@@ -1,4 +1,5 @@
 CHANGES
+
 2014-11-13
 - New timeInMillis config parameter in Elastica\Client, default false. Timeout could be considered as millisecods from now on.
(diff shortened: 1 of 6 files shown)


ruflin:
@skymeyer Any update here?

ruflin:
Closing this one as no updates for more then a year.

PR #724, opened 2014-11-16
QueryBuilder

webdevsHub:
Hi, after the ScanAndScoll-Iterator I wrote a QueryBuilder for Elastica!
# Goals
- full DSL support
- version check (is that command available in my elasticsearch installation?)
- convenience (autocomplete, links to doc, good exception messages)
- extendable (add/overwrite parts of the DSL)
# Implementation

I took the idea from [predis](https://github.com/nrk/predis), because they have a similar situation:
1. Every redis command is represented by a class in predis (like in Elastica: every Query/Filter/... has a class).
2. They have Profile-Classes (I called them Version) which contains all supported commands of a Version.
3. To execute a command via Client-Class predis uses __call() magic function to check whether or not that command is supported in configured version.

My implementation is different in some ways:
## DSL part classes

The existent command classes in Elastica can be used for QueryBuilder because they have a fluent interface. So I only had to create wrapper classes to easily access/construct these classes.

I created one class for each elasticsearch DSL part (Query, Filter, Aggregation, Suggesters) which constructs every existent Elastica-Object with a convenient method definition. You can find this classes in `Elastica\QueryBuilder\DSL`.

This is the heart of QueryBuilder and all other stuff is optional. If you don't want the Version-check you could just use this classes:

```
$queryDSL = new \Elastica\QueryBuilder\DSL\Query();
$queryDSL->match('field', 'value'); // returns Elastica\Query\Match
```

I added every available method according to elasticsearch documentation to the DSL classes. In case a method is not implemented in Elastica (e.g. query "custom_filters_score") you get a `Elastica\Exception\NotImplementedException`.

Most of the DSL methods just pipe the required Constructor arguments to the class, but I changed some method definitions for convenience reasons (e.g. `Elastica\QueryBuilder\DSL\Query::match()`)

There is only one single Filter, that causes problems: `Elastica\QueryBuilder\DSL\Filter::geo_shape()` 2 implementations exists: GeoShapeProvided, GeoShapePreIndexed. Maybe you have an idea to solve this?!
## QueryBuilder

`Elastica\QueryBuilder` takes a `Elastica\QueryBuilder\Version` object and registers the default DSL objects. You can access this DSL object via `__call()` (names defined by `Elastica\QueryBuilder\DSL::getType()`) or via `$qb->query()` `$qb->filter()` `$qb->agg()` `$qb->suggest()` convenience methods.

Add or overwrite a DSL object with `Elastica\QueryBuilder::addDSL()`
### Facade

Version check is done by `Elastica\QueryBuilder\Facade`. It wraps a DSL with a Version object and acts like a whitelist firewall. Note that the QueryBuilder convenience methods have a "wrong" @return annotation to get correct autocomplete.

The exception messages in Facade are very important, because they provide exact information about what is not working. For this reason I wrote tests for the exception messages.
## Tests

I'm not sure if we should test DSL and Version classes, because they are basicly configurations. The actual work is done by the returned objects and they are tested pretty well. I would like to discuss this!

I added a showcase for QueryBuider usage in `Elastica/Test/QueryBuilderTest::testQueryBuilder()`.
# Conclusion

I'm open to critics, even if you want major changes. Lets discuss the QueryBuilder topic without any restrictions, because I think a QueryBuilder is a very important functionallity for a Client library. We can just merge my current implementation or see it as a first proposal and continue working on it.

### DIFF
--- a/changes.txt
+++ b/changes.txt
@@ -7,7 +7,7 @@ CHANGES
 - Avoid remove previously added params when adding a suggest to the query #726
 
 2014-11-16
-- Added Elastica\QueryBuilder
+- Added Elastica\QueryBuilder #724
 
 2014-11-13
 - fixed reserved words in queries which composed of upper case letters (Util::replaceBooleanWords) #722


jdeniau:
Hi @webdevsHub,

This seems really nice, but can you document an example of a "complex" query ?

Thank you

webdevsHub:
@jdeniau please take a look at the new documentation: http://elastica.io/getting-started/search-documents.html#section-query

jdeniau:
Oops sorry I did not see that !

Thank you

ewgRa:
On my mind goal "version check" must be achieved not by implementing "Version", but by code assistant from IDE.

Solution like Version allows understand that you use not supported feature only on runtime level. So, there is no way to prevent misuse of Elastica on development time.

Lets imagine we in future and Elasticsearch during a time deprecate 584 queries. When I type $qb->query()-> - I will see all 584 queries and must understand which one I can use or not?

ewgRa:
For example sooner functionality will be dropped from Elasticsearch and from Elastica. From QueryBuilder it will be also dropped anyway.
And we return to situation when Version kind of strange thing - ability to use this or another functionality defined by code base.

Good case now - deprecation. We need to show user that query is deprecated now, but still supported by ES. 

If developer will use Version140 - he will be wondering - why he have deprecation message, because it is deprecated in 1.7.0. It give unexpected behavior.

ewgRa:
Since each Elastica release declare with which one Elasticsearch version it compatible - can you show real case of usage Version?

webdevsHub:
Let me first try do understand your question and issues before getting into discussion:

I added that version check because of the very fast changing Elasticsearch API (mostly in between 0.9 and 1.*) and because there are so many different versions used by developers out there.

The old way to handle that was: just use the corresponding Elastica version released for a specific Elasticsearch version. But those old versions are no longer maintained so you do not get bug fixes or new features.

Version check is intended to find misuse because you construct QueryBuilder with a specific ES version and it throws exceptions if you use something not supported in that version. Now you can use latest release of Elastica with old Elasticsearch DSL.

You are right: that exception is thrown on runtime level + `@deprecated` might be confusing if you do not read annotation message and DSL classes stacks up "deprecated waste".

That approach will no longer work when Elasticsearch change all query names (like your worst case with 584 deprecated queries) or when a new version is completly incompatible with old ones (like a query JSON structure change).

@ewgRa Have I understood that properly?

ewgRa:
@webdevsHub Thank for detailed explanations and yes, you focus on my concerns here.

My point is:

1. Version not prevent misuse. You can understand misuse only at runtime, it is not helpful. QueryBuilder must help, but now it is suggest wrong directions.

2. Problems with deprecated and removed functionality.

webdevsHub:
So you want to remove all version stuff and change DSL classes to reflect current API?

ewgRa:
I can suggest to do this, if another developers will agree with me.

ruflin:
Thanks for the details @webdevsHub @ewgRa . I would pose the question a little bit differently: Does it with Elastica 3.0.0 still make sense to keep the versions or is it just an overhead we have to maintain. My opinion is that we should have probably removed all the old versions with the release of 3.0.0 but unfortunately I missed it. As Elastica is decoupled from the elasticsearch versioning, major changes are minor or major version numbers. So when we break the QueryBuilder, there is probably also a new release out. And functions are mainly only added, rarely removed.

@webdevsHub What is your opinion on the versions? Are they still needed?

webdevsHub:
ewgRa pointed out that all this version stuff does not work + create problems. So lets just delete that. The only solution to ewgRa 2. problem would be to create DSL classes for each version in different repositories and let users replace DSL classes via composer; or create a IDE plugin :)

Removing version system + change DSL classes to current ES API (which includes useful deprecation annotations) is the best option in my opinion

webdevsHub:
I can create a pull request to remove version stuff. About DSL classes change: I am not up to date with ES API 2.* (I am still working only on 1.*), is there any overview of what is new, deprecated, ... without having to compare different ES doc HTML?

ruflin:
@webdevsHub A PR would be great. I haven't seen a very simple comparison yet. Just go with what we have now. More functions can always be added.

Issue #492, opened 2013-10-31
FunctionScore: `scale_weight` is not a supported by elasticsearch


Xylakant:
The `scale_weight` parameter does not exist for function_score in the final implementation: https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Query/FunctionScore.php#L96. The decay is now only controlled by the `decay` parameter.

PR #735, opened 2014-12-19
Implement the `weight` in the function score query

jdeniau:
The [weight](http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html#_weight) now replace the `boost_factor` since the 1.4.0.

I also fixed https://github.com/ruflin/Elastica/issues/492 by replacing the `scale_weight` parameter by the `weight`


ruflin:
Merged. Thx :-)

PR #737, opened 2014-12-23
#736: Additional Request Body Options for Percolator

### LINKED ISSUE #736: Percolator: Support "additional request body options" per the ElasticSearch official documentation
The [documentation](http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/search-percolate.html) for Percolator discusses some additional request body options that can be passed when making a percolate request. Currently, there is no easy way to manage this in Elastica.

In Percolate, we have this:
```php
public function matchDoc(Document $doc, $query = null, $type = 'type', $params = array())
{
    $path = $this->_index->getName() . '/' . $type . '/_percolate';
    $data = array('doc' => $doc->getData());
    return $this->_percolate($path, $query, $data, $params);
}
```

The additional request body options are: `filter`, `query`, `size`, `track_scores`, `sort`, `facets`, `aggs` and `highlight`.

I'm working on a pull-request which processes the `$params` array to extract relevant request body options and add them to the `$data` array, but would be interested to know if anyone has a better solution in mind.

ck99:
Fixes #736

### DIFF
--- a/lib/Elastica/Percolator.php
+++ b/lib/Elastica/Percolator.php
@@ -12,6 +12,26 @@ namespace Elastica;
  */
 class Percolator
 {
+    const EXTRA_FILTER         = 'filter';
+    const EXTRA_QUERY          = 'query';
+    const EXTRA_SIZE           = 'size';
+    const EXTRA_TRACK_SCORES   = 'track_scores';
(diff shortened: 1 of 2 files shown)


ruflin:
@ck99 Sorry for the delay. Now it is finally merged. Thx.

PR #742, opened 2015-01-05
New vagrantbox

im-denisenko:
- Vagrantfile was updated to use ubuntu/precise32 as default box and ansible as provision system
- Travis setup was also rewrited because this is very similar process as creation of new vagrant box


ruflin:
Merged. Is there an easy option to start multiple instances of ES that work in a cluster?

im-denisenko:
Easiest one is create two config files and add custom logic into [/etc/init.d/elasticsearch](https://github.com/elasticsearch/elasticsearch/blob/master/src/deb/init.d/elasticsearch), i guess.
I can fix it later, if i'll figure out how this bash magic works.

ruflin:
Sounds good. Actually that all the tests work properly at least 3 instances are needed, as one is shut down to test if shutting down works.

im-denisenko:
How about add something like this after every shutdown call?
```php
system('sudo service elasticsearch restart');
sleep(10);
```
It's quite ugly, but provides a way to restore all shutdowned nodes at the end of each test, so:
- two nodes will be enough, 
- all runs of phpunit will produce same result (good for virtualbox)
- Elastica\Cluster::shutdown can be tested too

ruflin:
There are two reasons for the multiple instances:
* Make sure that the tests also work properly with multiple instances (which was not the case in the beginning) to represent a real cluster
* Testing the shutdown

For the first part we only need at least 3 instances. For the second part, in case we restart the service with the above command, I'm not sure how we handle the waiting for the starting of the service. Perhaps we should run the "shutdown" tests in a second independent test set, means in case something goes wrong it does not jeopardise the rest of the tests.

PR #743, opened 2015-01-07
Fix issue 718

im-denisenko:


im-denisenko:
Btw, I noticed a lot of random failed builds with two reasons for most of them:
- OutOfMemory exception (when almost all tests fails)
- Timeout (phpunit stops after 50-100 executed tests)

OutOfMemory probably fixed in #742. Problem is, at [this line](https://github.com/ruflin/Elastica/blob/master/test/bin/run_elasticsearch.sh#L17) three instances of ES starts, while phpunit is actually uses only one (at least I didn't found as 9201 and 9202 ports are used in the tests). Also, explicit memory limit can be added to ES config (I'm not sure which exactly option and value should be used for that).

Timeout is probably about memory too, but maybe some tests should be refactored?

ruflin:
The multiple instances are needed for the tests. The cluster instances connect automatically to each other. That is one of my first concerns when I looked at #742 that I couldn't find the multiple instances.

I will look at both of your pull requests as soon as possible. Thx for the contributions.

macnibblet:
I could really use this, any chance it can get merged soon ?

ruflin:
Merged.

@im-denisenko Thx.

PR #745, opened 2015-01-10
Create .gitattributes with basic export-ignores

rmruano:
Files and directories with the attribute export-ignore won’t be added to archive files. This will prevent composer from downloading unnecessary files (i.e: tests) when --prefer-dist is set.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+/test export-ignore
+.gitattributes export-ignore
+.gitignore export-ignore
+.travis.yml export-ignore
+.coveralls.yml export-ignore


fprochazka:
:+1:

im-denisenko:
Documentation about how to use `.gitattributes` for ignoring some parts of package was removed from composer in https://github.com/composer/composer/commit/b34220edc8ed738bce3f682248622976543d6ec4 and this was discussed here https://github.com/symfony/symfony/issues/6605. TLDR: now this practice considered bad.

@ruflin This file is incomplete at this moment. Should we keep this updated or perhaps remove it too?

rmruano:
@im-denisenko I don't think that removing it from the documentation or that symfony reverted it means this practice is considered bad and should be discouraged but I can understand it because some projects are excluding them and others not which could cause confusion. IMHO, I prefer not including the test suite in the builds because you still have `--prefer-source` if you need them. Anyway, I don't have any problem with reverting this if @ruflin consider they should be included in the builds.

ruflin:
The reason I like to include the tests is because I often refer to the tests if someone looks for examples. This can be handy for the engineering. 

For the other files: Is there any benefit not having them in the package?

rmruano:
.gitattributes excludes only affects builds, clones and composer
--prefer-source would still retrieve all files as usual without loosing the
advantage of having a clean package for production.

If you remove tests from it (.gitattributes) there's no point in keeping
the excludes for just a few files, the file can be safely removed then.

ruflin:
If a build is created or it is deployed to production, tests etc. are not needed. But if someone works on the projects and the library is cloned with composer, test are part of the documentation. 

The question I ask myself is what should be the default? If .gitattributes is not set, there is no way to exclude it for production packages as far as I know (please correct me here if I'm wrong). If the .gitattribute exists, at least the option to include it exist. The only problem I have here is that most engineers will never to that as they don't know about the flag.

PR #747, opened 2015-01-10
Add ignore_unavailable as a valid option to search

tmatei:
Could we add the ignore_unavailable option as a valid option to search?

http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/multi-index.html#multi-index


ruflin:
Sure. Can you add a small integration test here?

tmatei:
Added tests, can you merge it?

ruflin:
Ok merged it. Travis seems to have some issues at the moment :-(

tmatei:
Thanks! I noticed unit tests take a very long time to run. Is this normal?

ruflin:
It is not only unit tests, but also integration test. In total they should not take more then about 7 minutes (per environment). I also just merged https://github.com/ruflin/Elastica/pull/742 and hope with this the tests work again.

PR #748, opened 2015-01-13
Set correct script attributes in AbstractSimpleAggregation

matthiasmullie:
Existing code would not set 'lang' (making it impossible to e.g. use groovy in those scripts)
It would also always add 'params', even if there aren't any


ruflin:
Can you add a small test to make sure this problem does not occur again?

ruflin:
Thx, merged.

PR #749, opened 2015-01-13
Use instanceof instead of is_a

XWB:
The `is_a` method is deprecated in favour of the instanceof operator.


ruflin:
Thx. We should have spotted that much earlier.

PR #751, opened 2015-01-14
A Little Housekeeping

XWB:
This bundle can use a little cleanup. I used https://github.com/FriendsOfPHP/PHP-CS-Fixer to resolve most issues.

### DIFF
--- a/lib/Elastica/Bulk/Action/AbstractDocument.php
+++ b/lib/Elastica/Bulk/Action/AbstractDocument.php
@@ -23,7 +23,7 @@ abstract class AbstractDocument extends Action
     }
 
     /**
-     * @param \Elastica\Document $document
+     * @param  \Elastica\Document                     $document
      * @return \Elastica\Bulk\Action\AbstractDocument
      */
     public function setDocument(Document $document)
@@ -38,7 +38,7 @@ abstract class AbstractDocument extends Action
     }
 
     /**
-     * @param \Elastica\Script $script
+     * @param  \Elastica\Script                       $script
      * @return \Elastica\Bulk\Action\AbstractDocument
      */
     public function setScript(Script $script)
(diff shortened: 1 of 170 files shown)


lib/Elastica/Bulk/Action/AbstractDocument.php
```diff
@@ -100,7 +96,7 @@ public function getScript()
             return $this->_data;
         }
 
-        return null;
+        return;
```
ruflin: I assume this is done automatically by the cleanup script. I'm not a fan of "empty" return, but if we have an automated way to cleanup things, I can also "life" with this version.

ruflin:
@XWB I like almost all changes except the return null; and ' . ' without space. It also changes some of the standards we used before but I prefer to have an automatic script that fixes the coding issues and it aligned with a standard. We should include this in our build file if possible.

XWB:
> I assume this is done automatically by the cleanup script.

Correct, though we can tweak the script a bit. The following options are available:

```
php php-cs-fixer.phar fix /path/to/project --level=psr0
php php-cs-fixer.phar fix /path/to/project --level=psr1
php php-cs-fixer.phar fix /path/to/project --level=psr2
php php-cs-fixer.phar fix /path/to/project --level=symfony
```

According to http://cs.sensiolabs.org/, the level `symfony` includes:

```
* concat_without_spaces [symfony] Concatenation should be used
without spaces.
* empty_return [symfony] A return statement wishing
to return nothing should be simply "return".
```

The Symfony fixer is not be required for `ruflin/elastica` so we could ignore it in future clean-ups.

XWB:
@ruflin I didn't expect you to merge this PR so quickly :P

If you want, you could revert the commit (or remove it and force update the master branch) and I can send a new PR without the Symfony code standard (see comment above).

ruflin:
As the Symfony bundle is one of the biggest Elastica users, I don't think we need to revert anything. I prefer to have a reliable checker for the styles. What happens if you now apply only the first three? Would that lead to any changes?

PR #753, opened 2015-01-14
Fix travis build

im-denisenko:
Right now some builds ([1](https://travis-ci.org/ruflin/Elastica/jobs/46968655), [2](https://travis-ci.org/ruflin/Elastica/jobs/46968653), [3](https://travis-ci.org/ruflin/Elastica/jobs/46819813), [4](https://travis-ci.org/ruflin/Elastica/jobs/46968652), [5](https://travis-ci.org/ruflin/Elastica/jobs/46968648), etc) failed on step when ansible trying to import ES repository gpg key.
I'm don't sure what exactly going wrong (from virtualbox import works fine, so maybe network settings or something travis-specific), but possible solution is to retrieve key by url instead of keyserver.

### DIFF
--- a/ansible/roles/elasticsearch/tasks/main.yml
+++ b/ansible/roles/elasticsearch/tasks/main.yml
@@ -29,7 +29,7 @@
 - name: import elasticsearch gpg key
   apt_key: >
     id=D88E42B4
-    keyserver=pgp.mit.edu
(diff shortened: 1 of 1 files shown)


ruflin:
Ok. Lets see if this works more stable.

PR #757, opened 2015-01-23
Added Elastica\Query\Regexp

gnit:


webdevsHub:
Nice! One step further in the "full language support"-mission :clap:  

Suggestion: you could implement the flags mentioned in documentation (ANYSTRING, COMPLEMENT, EMPTY, INTERSECTION, INTERVAL, or NONE) as constants and provide a method to add one or more flags to the query (or as an additional argument).

Could you please add that new class to QueryBuilder in https://github.com/ruflin/Elastica/blob/master/lib/Elastica/QueryBuilder/DSL/Query.php#L396 with same method arguments as in the construct (+ return, param annotations in DocBlock Comment of the method).

gnit:
QueryBuilder is now OK

I will add Flags implementation soon

ruflin:
Thx.

PR #738, opened 2014-12-23
Elasticsearch 1.4.2

jdeniau:
I tried to update to 1.4.2 (based on your 1.4.0 branch).

I now have 8 failing tests:
7 with this message : 
```
Failed asserting that exception of type "Elastica\Exception\ResponseException" matches expected exception "\Elastica\Exception\RuntimeException". Message was: "ClusterBlockException[blocked by: [FORBIDDEN/5/index read-only (api)];]" 
```

The last one is that ```Thread.sleep(100); ``` is [not authorized anymore now by ES](https://github.com/elasticsearch/elasticsearch/blob/master/src/test/java/org/elasticsearch/script/GroovySandboxScriptTests.java#L82) 

I do not know how to solve those problems :/

PR #759, opened 2015-01-24
Fix cluster setup/Update php version

im-denisenko:
Hello.

From comments of #742, summary of changes:
- Fix multi-instance setup in build script. Now all cluster-specific tests not skipped.
- Move shutdown tests to separate testsuite.
- Split phpunit.xml to two configs, so in virtualbox phpunit by default will not shutdown cluster and generate clover and coverage reports

After that, I failed to realise what it's enough for PR. So, from comments of #738, here is further summary :)
- Php version in composer bumped to 5.4.0
- Guzzle included in require-dev, for Transport\Guzzle tests
- Fix bugs in Transport\Guzzle, one is incorrect handling of "403 Forbidden" response, second is incorrect proxy setup
- Fix tests for Transport\Guzzle and Transport\Http


im-denisenko:
Also, after cluster fix, problem with travis builds like [this one](https://travis-ci.org/ruflin/Elastica/jobs/48128207) is actual again. 
Near every fifth build will be with 80% failed tests without any visible reason.
I thought this is because of memory, but [it's not](https://github.com/im-denisenko/Elastica/commit/b3293892ca47e294780d6874e3d53dd51506537d).
I have no idea how to fix it.

// please restart build

test/phpunit.xml.dist
```diff
@@ -1,33 +0,0 @@
-<phpunit
```
ruflin: Why did you rmeove the phpunit.xml.dist file? The nice part about is that it can be locally overwritten and the .dist stays still in place.

ruflin:
I restarted the build. Lets see what the outcome will be.

ruflin:
At least most tests passed :+1:

ruflin:
I merged it and we can move forward from here.

ruflin:
It seems we are missing the memcache php library: https://travis-ci.org/ruflin/Elastica/jobs/48128202#L466

im-denisenko:
Build of master branch was suddenly failed after success with phpunit because of [github rate limit](https://travis-ci.org/ruflin/Elastica/jobs/48387592#L5033), lol :)
I'll check memcache and api rate in a few days.

test/phpunit.xml.dist
im-denisenko: Didn't even knew about this feature - my main intention was to split up configs.
File can be safely renamed back to .dist, hovewer.

ruflin:
@im-denisenko I think here is a solution for the github rate limit problem: https://github.com/composer/composer/issues/1314

Seems like @fprochazka is also aware of this one ;-)

im-denisenko:
25 references to there - such popular issue.
Thanks, anyway. It is also suddenly closes one ticket on my current job position, hah.

PR #761, opened 2015-01-27
Fix composer in travis

im-denisenko:


ruflin:
Thanks. Not sure why one of the builds failed. But it seems to solve the Github issue. Lets see what happens to the follow up builds.

ruflin:
@im-denisenko Currently the last build always fails because of memory issues. I remember you played around with the memory? https://travis-ci.org/ruflin/Elastica/jobs/48605333

im-denisenko:
I still can't figure out, what happens in builds with OutOfMemory error, even can't reproduce it in virtualbox. Only thing i'm sure that first error always happens in aggregation tests. If aggregation tests passes, build will be successfull.

I suggest to create a new issue for that, to mark problem as "known". Maybe, someone who already know solution, will see it.

ruflin:
So increasing the elasticsearch memory didn't help here?

im-denisenko:
Don't think so. First time I [reference this problem](https://github.com/ruflin/Elastica/pull/743#issuecomment-68983772), memory limit had default value 1g (btw, travis machine has 3g of memory, so 3 instances was too much for them). Now it has value [256m](https://github.com/ruflin/Elastica/blob/master/ansible/roles/elasticsearch/templates/elasticsearch.service#L60) per instance, and in most builds it's enough.
Also, as I said, I can't reproduce this in virtualbox, which has [1g of memory](https://github.com/ruflin/Elastica/blob/master/Vagrantfile#L15) total, that much less, than in travis.

ruflin:
But if you set it to 256, the hard limit is 256. Perhaps we should try 512? The strange thing is that always the last build fails. Either it is connected to the PHP version or the build VM's are somehow connected.

im-denisenko:
I still think this is random and not related to memory limit. And it's not always last build. [Here](https://travis-ci.org/im-denisenko/Elastica/builds/48665942) was failed both first and sixth, but [here](https://travis-ci.org/ruflin/Elastica/builds/48605323) was sixth only. Both builds based on the same commit.
I can try to fix it somehow, but with speed "one try per hour" it'll take a very long time.

Just for history, my assumptions about it:
- maybe it's really just a memory problem
- maybe each test should delete created indices in tearDown
- something wrong with aggregation tests
- ES instances should use separate directories for data files
- something wrong with JVM params
- something wrong with ulimit
- maybe tests runs so fast and little sleep between them can help

ruflin:
Nice list. The problem is that when the tests run too long, quite often there is not output / error messages that can be analysed. The best way would be to try to reproduce it locally. 

The failed tests take really long, often exceed the maximum time. So this means indirectly, that the response is really slow which could mean es has problem on writing files. If I look at your list, ulimit jumps into my eyes. Perhaps we should try to set it at least to 65k or higher.

PR #762, opened 2015-01-27
Add missing interface to ElasticsearchException

im-denisenko:
This addition provides 100% assurance in behavior of code below (for example, used in [Monolog](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/ElasticSearchHandler.php#L122)):
```php
try {
    // do something
} catch (Elastica\Exception\ExceptionInterface $ex) {
    // ignore all possible exceptions from elastica,
    // but let all other exceptions be thrown
}
```

### DIFF
--- a/changes.txt
+++ b/changes.txt
@@ -1,5 +1,8 @@
 CHANGES
 
+2015-01-27
+- Exception\ElasticsearchException now can be catched like all other exceptions as Exception\ExceptionInterface
+
 2015-01-25
 - Release v1.4.2.0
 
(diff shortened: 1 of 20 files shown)


ruflin:
Merged. Of course some of the builds failed because of https://github.com/ruflin/Elastica/pull/759#issuecomment-71705711 ;-)

ruflin:
BTW: With your recent pull requests you increased the coverage from 81% to at least 83% = Awesome

im-denisenko:
yeah, awesome :)

PR #766, opened 2015-01-31
Fix missing memcache library

im-denisenko:
As said in https://github.com/ruflin/Elastica/pull/759#issuecomment-71524030, we are missing memcache. I dont know if there are exist someone who would actually use this transport due to limitation on key length for GET requests (to be honest, for me this plugin and transport seems useful like NullTransport), but I fixed this issue.

- added installation of memcache library
- removed installation of memcached server
- added installation of transport-memcached plugin
- added basic tests for crud operations using memcached transport
- memcached transport fixed


im-denisenko:
As always, build failed.

lib/Elastica/Transport/Memcache.php
```diff
@@ -79,4 +94,17 @@ public function exec(Request $request, array $params)
...
         return $response;
     }
+
+    /**
+     * Check if key that will be used dont exceed 250 symbols
+     *
+     * @throws Elastica\Exception\Connection\MemcacheException If key is too long
+     * @param  string                                          $key
+     */
+    private function checkKeyLength($key)
```
ruflin: Can you use _checkKeyLength here as the function is private?

ruflin:
Nice, all tests except the two incomplete ones are passing. And of course a random build is failing ...

lib/Elastica/Transport/Memcache.php
im-denisenko: Sure, fixed.
Speaking of that, this library convention is slightly different than that [PSR-2](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md) says:
> Method names SHOULD NOT be prefixed with a single underscore to indicate protected or private visibility.
> Property names SHOULD NOT be prefixed with a single underscore to indicate protected or private visibility.

ruflin:
Almost 84% ;-)

ruflin:
@im-denisenko Concerning the convention difference to PSR-2. I'm aware of that. That is the only thing I don't like about PSR-2. The good thing is it is only "SHOULD NOT" and not "MUST". In case we decide to change it, it would have to be changed across the whole library.

PR #767, opened 2015-02-02
reset php 5.3 tests on travis

jdeniau:
This PR is due to the discussion about PHP 5.3 [here](https://github.com/ruflin/Elastica/pull/764#issuecomment-72378573).

The short array syntax has been remove [in this PR](https://github.com/ruflin/Elastica/pull/737/files) by the way so it seems as simple as that. :)

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -1,19 +1,19 @@
 <?php
 
 $finder = Symfony\CS\Finder\DefaultFinder::create()
-    ->in(['lib', 'test']);
+    ->in(array('lib', 'test'));
 
 $config = Symfony\CS\Config\Config::create()
     ->setUsingCache(true)
     ->level(Symfony\CS\FixerInterface::SYMFONY_LEVEL)
-    ->fixers([
+    ->fixers(array(
         // [contrib] Multi-line whitespace before closing semicolon are prohibited.
         'multiline_spaces_before_semicolon',
         // [contrib] There should be no blank lines before a namespace declaration.
         'no_blank_lines_before_namespace',
         // [contrib] Ordering use statements.
         'ordered_use',
-    ])
+    ))
     ->finder($finder);
 
 return $config;
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ install:
 before_script:
   - mkdir -p build/logs
   - echo "extension=memcache.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
+  - if [ $ES_COMPOSER_NODEV == "no" ] && ! $(php -v | head -1 | grep -q 'PHP 5.3'); then composer require "guzzlehttp/guzzle" "4.2.*" --dev --no-ansi --no-progress --no-interaction; fi
 
 script:
   - phpunit -c test/phpunit-travis.xml
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,14 @@
 language: php
 
 php:
+  - 5.3
   - 5.4
   - 5.5
   - 5.6
+
+matrix:
+    allow_failures:
+        - php: 5.3
 
 env:
   matrix:
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,6 @@ install:
 before_script:
   - mkdir -p build/logs
   - echo "extension=memcache.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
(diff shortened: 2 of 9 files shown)


im-denisenko:
Please fix this files: https://github.com/ruflin/Elastica/commit/7b56732f7853143003357ab10fd8724a02a9184a https://github.com/ruflin/Elastica/commit/e689375e9f6a0bf0a926ae18fa9f53237524cfd5

Also, this packages can't be installed on 5.3: https://github.com/ruflin/Elastica/commit/7b56732f7853143003357ab10fd8724a02a9184a https://github.com/ruflin/Elastica/commit/a6ec47e0bcdff1f07c586f12015471fa922a9d88

im-denisenko:
It is hard to downgrade guzzle dependency to "3.8.1"? It requires php ">=5.3.3", so will work with any version, and Transport/Guzzle will always be tested.

im-denisenko:
It is hard to downgrade guzzle dependency to "3.8.1"? It requires php ">=5.3.3", so will work with any version, and Transport/Guzzle will always be tested.

Or maybe install guzzle manually (not via `composer.json`) in `.travis.yml`, if current build its not 5.3? Like that:
```yaml
before_script:
  - mkdir -p build/logs
  - echo "extension=memcache.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
  - [ "$ES_COMPOSER_NODEV" == "no" ] && composer require "guzzlehttp/guzzle" "4.2.*" --dev --no-ansi --no-progress --no-interaction || true
```

ruflin:
Sorry for the late answer. So if I understand the problem right: Guzzle is PHP 5.4 only, but there is an old version that isn't. The question is now on how to use Guzzle PHP 5.3 for one build, and 5.4 for all the other builds? What if we just say Guzzle Client is not supported for PHP 5.3, we don't include it and we skip the tests?

jdeniau:
You are right.

If we can do that, it will be the better we can do, this way, we do not try to hack the composer process on PHP 5.3.

My knowledge of phpunit on how to skip test is limited though, and I do not know how to do this.

im-denisenko:
@jdeniau for skipping test, you just call `self::markTestSkipped` method. That @ruflin suggests - it's done already, just drop guzzle from require-dev and pr can be merged.

But then GuzzleTransportTest will always be skipped, and any errors in this transport will be undetectable via travis builds. And this is ~2% of coverage, if coveralls not kidding us :)

I checked logs from [this build](https://travis-ci.org/ruflin/Elastica/builds/49324365). More likely, it's permission issue (my fail I forgot about that, [again](https://github.com/im-denisenko/Elastica/commit/5cfcc25707569f2758852fd209ad5c61dbc58fa1)). It can be fixed, if `sudo composer require` will be used instead of `composer require` in the command I wrote above.

ruflin:
@jdeniau What I meant is more that
```
if (php < 5.4) {
   self::markTestSkipped
}
```

So only for PHP 5.3 it is skipped.

ruflin:
@jdeniau @im-denisenko Any updates here?

im-denisenko:
There are no way to check php version after phpunit was started, if guzzle is installed.
Guzzle 4.2 uses [file autoloading](https://github.com/guzzle/guzzle/blob/4.2.3/composer.json#L31) and in their `src/functions.php` used short array syntax.
So, all 5.3 builds will fail immediately after start, [like here](https://travis-ci.org/ruflin/Elastica/jobs/49429591#L422).

Therefore, php version should be checked *before* phpunit start. It could be done with snippet I wrote above, but with sudo (because `vendor/` directory owner is root):
```yml
before_script:
  - mkdir -p build/logs
  - echo "extension=memcache.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
  - if [ $ES_COMPOSER_NODEV == "no" ] && ! $(php -v | head -1 | grep -q 'PHP 5.3'); then sudo composer require "guzzlehttp/guzzle" "4.2.*" --dev --no-ansi --no-progress --no-interaction; fi
```

ruflin:
Based on this pull request I opened a branch with the above script inside. Lets see if it passes: https://github.com/ruflin/Elastica/compare/jdeniau-php-53-es-142

ruflin:
@im-denisenko Doesn't seem to work as expected: https://travis-ci.org/ruflin/Elastica/jobs/50792723

im-denisenko:
@ruflin It's because guzzle still in require-dev. I change this there #779

jdeniau:
Thank you @im-denisenko for the last commits ;)

ruflin:
Thanks @im-denisenko @jdeniau This is now in master.

PR #768, opened 2015-02-02
Allow scalar index

bakura10:
Hi,

Numbers are a valid index name, so the fix should also be applied here :).

Thanks!


ruflin:
@bakura10 Can you add a brief test for this to make sure it is never reverted back?

bakura10:
Sure, I've made a proper PR in #769

PR #769, opened 2015-02-02
Allow scalar type

bakura10:
Replace #768

### DIFF
--- a/test/lib/Elastica/Test/SearchTest.php
+++ b/test/lib/Elastica/Test/SearchTest.php
@@ -146,6 +146,8 @@ class SearchTest extends BaseTest
         $search = new Search($client);
 
         $search->addIndex(1);
+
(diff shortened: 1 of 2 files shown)


test/lib/Elastica/Test/SearchTest.php
```diff
@@ -140,6 +140,14 @@ public function testAddIndexInvalid()
         $search->addIndex(new \stdClass());
     }
 
+    public function testAddNumericIndex()
+    {
+        $client = $this->_getClient();
+        $search = new Search($client);
+
+        $search->addIndex(1);
```
ruflin: I would suggest to also add an assert, that the "index" was successfully added. For example you could use getIndexes() and check if it is part of it.

bakura10:
Done :)

ruflin:
Merged. Thx.

PR #773, opened 2015-02-10
Add support for filters aggregation

bakura10:
Hi,

This adds support for the new Filters aggregation of Elasticsearch 1.4, with tests. As per Elasticsearch, it supports both anonymous and named filters.

Once this is merged, would it be possible to have a new point release tagged please? :)


ruflin:
Thx. I will try to make a release this weekend.

bakura10:
Thanks a lot!

PR #774, opened 2015-02-10
Fix IDE doc for better auto-completion

bakura10:
On modern IDE (at least on PHPStorm), returning static allows to let the IDE resolve on real-time based on the caller type. This is necessary if we still want to have nice autocompletion with the fluent interface. I've only fixed that for param and aggregations, I'll do another PR in the future if I find other cases :)

### DIFF
--- a/lib/Elastica/Aggregation/AbstractAggregation.php
+++ b/lib/Elastica/Aggregation/AbstractAggregation.php
@@ -58,7 +58,7 @@ abstract class AbstractAggregation extends Param
      * Add a sub-aggregation
      * @param  AbstractAggregation                  $aggregation
      * @throws \Elastica\Exception\InvalidException
-     * @return AbstractAggregation
+     * @return static
      */
     public function addAggregation(AbstractAggregation $aggregation)
     {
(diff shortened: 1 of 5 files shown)


im-denisenko:
I always thought that `@return static` or `@return self` should be used for static methods like Elastica\Search::create, but in case of fluent interface `@return $this` is more semantic, isn't it? This differences also described in proposed [psr-5](https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#keyword) (13, 14, 15).

// my ST3 cannot into autocomplete, so maybe i'm just wrong :)

bakura10:
I just tried and I can confirm that $this work as well :). I like static because it is coherent with how late-state binding works, but I'm open to change that to $this. Let's let @ruflin decide :D.

ruflin:
I wasn't even aware that PHPStorm is now able to use static or $this for auto complete. I quite like "$this" as it describes very well what it actually is. In which IDE's is this going to work?

bakura10:
I'm pretty sure NetBeans take advantage of that!

im-denisenko:
Eclipse confirmed also

bakura10:
I've updated to modify static to $this!

ruflin:
Merged. Thx.

PR #775, opened 2015-02-11
Fix out of memory

im-denisenko:
This PR fixes OutOfMemory exception (which indeed must be named like OutOfThreadLimit, because problem is not related to memory at all) in travis builds. Last 132 builds of last commit (22 runs * 6 php versions) [were successfull](https://travis-ci.org/im-denisenko/Elastica/builds), except 3, that were failed in `provision.sh` on step `apt-get install` or `pip install` due to network issues.

Summary of changes:
* ES threadpool size was limited by using `processors: 1` setting
* Thread stack size was limited (if you have 100M for all threads, and each thread consumes 10M, you can start 10 threads. With 1M per thread, you can start 100. 1024 is minimal value at which ES instance can be started)
* Third instance of ES was removed, because two is enough for all tests, including nodeShutdown and clusterShutdown
* Memory limit was fixed (parameter for `ulimit -l` must be either an integer 256 or "unlimited", but string "256M" was used)
* Modification of `vm.max_map_count` was removed, because we can't use sysctl inside travis machine
* `_createIndex` now will generate unique name for created index, if name was not provided as first argument
* Test cluster now accepts connections only from localhost, because I saw from some logs, how node with random name (not "Silver Fox" nor "Skywalker") had joined to the cluster.
* I/O operations decreased by adding `index.store.type: memory` setting
* `Test\Base::tearDown` now performs deletion of all indices and caches between tests


ruflin:
Nice, thanks. What would be the consequence of still having 3 instances for testing? Would we run into any limitations because of the criterias you mentioned above?

im-denisenko:
Yep, there will be problems.

I was trying a lot not do this, but at the end I failed to get stable 3-nodes cluster. Third node always either dies immediately after start or in the middle of test suite, regardless of used settings. It dies even if phpunit will sleep between tests and run them very slow.

Anyway, I'm sure, that run multiple nodes on machine with such low performance it's not an usual way to use elasticsearch, so it's good thing at least two instances can be started.

ruflin:
Ok, good to know. So I assume the issues appear on travis and not on the local machine?

ruflin:
BTW: The only issue I have with the new tests is that they take quite a bit longer. I assume this is because of the removal of the indices. But here stability goes over speed.

im-denisenko:
Yes, I still can't reproduce it in the virtualbox. I think, it's could be because of virtualization system difference. In travis this is openvz, and in virtualbox this is em... virtualboxvz? I'm not good at this, so I will stop making further assumptions :)

ruflin:
I'm happy it works now ;-) Thx a lot.

PR #779, opened 2015-02-15
Fix 5.3 build

im-denisenko:


composer.json
```diff
@@ -18,12 +18,11 @@
     "require-dev": {
         "munkie/elasticsearch-thrift-php": "1.4.*",
         "phpunit/phpunit": "4.1.*",
-        "guzzlehttp/guzzle": "4.2.*",
         "fabpot/php-cs-fixer": "~1.4"
     },
     "suggest": {
         "munkie/elasticsearch-thrift-php": "Allow using thrift transport",
-        "guzzlehttp/guzzle": "Allow using guzzle 4.x as the http transport",
+        "guzzlehttp/guzzle": "Allow using guzzle 4.x as the http transport (Requires php 5.4)",
```
merk: This change means that travis will never be testing the guzzle transport

merk: Never mind. Missed the above condition :)

PR #780, opened 2015-02-15
Feature: support for scripted metric aggrations

ukautz:
Scripted metric aggregation is still marked as experimental (as of ES 1.4.3) but I need it and thought maybe you want to merge it to your great package (love it!)


ruflin:
Merged. Would be nice if you could keep an eye on it in case some changes are made to the functionality.

ukautz:
Thanks for merging! Will do since I'm forced to using scripted metrics until ES 2.0.

PR #787, opened 2015-02-20
Add Elastica\Query\Image

Jmoati:
Add support of https://github.com/kzwang/elasticsearch-image


im-denisenko:
Is this plugin alive? I see they support ES 1.1.0, and last commit was half a year ago. 
PR with support for 1.4.2 was opened month ago and still has no comments. 
Elastica supports ES 1.4.3 right now, how would you use it with outdated plugin?

Jmoati:
I use https://github.com/ifgh/elasticsearch-image which is a 1.4.2 fork.
I think that it will work with 1.4.2 too.

test/lib/Elastica/Test/Query/ImageTest.php
```diff
@@ -0,0 +1,45 @@
...
+    public function testToArrayFromImage()
+    {
+        $field = "image";
+
+        $query = new Image();
+        $query->setFieldFeature($field, 'CEDD');
+        $query->setFieldHash($field, 'BIT_SAMPLING');
+        $query->setFieldBoost($field, 100);
+
+        $query->setFieldImage($field, __DIR__ . '/../../../../data/test.jpg');
```
ruflin: To have a more readable path, check here: https://github.com/ruflin/Elastica/blob/master/test/lib/Elastica/Test/IndexTest.php#L124

lib/Elastica/Query/Image.php
```diff
@@ -0,0 +1,180 @@
...
+        return $this->setFieldParam($field, 'hash', $hash);
+    }
+
+    /**
+     * Set field image value
+     *
+     * @param  string $field
+     * @param  string $path File will be base64_encode
+     *
+     * @return Image
```
ruflin: I assume this should also be \Elastica\Query\Image

ruflin:
To make sure it works with the current version, we should add an integration tests, which means we must add the plugin to our setup process. Like this we can also be sure it works with future versions.

Also we should add it then to the dependencies in the README file: https://github.com/ruflin/Elastica

@Jmoati If you need some help to get the setup done etc. we are here to help.

im-denisenko:
More concrete steps how to add plugin to build:
- Add variable ES_IMAGE_PLUGIN_VER [here](https://github.com/ruflin/Elastica/blob/master/ansible/es-playbook.yml#L13)
- Add plugin installation [here](https://github.com/ruflin/Elastica/blob/master/ansible/roles/elasticsearch/tasks/main.yml#L67)
- Add plugin name [here](https://github.com/ruflin/Elastica/blob/master/ansible/roles/elasticsearch/templates/config-default.yml#L10)
- Add plugin to table of dependencies [here](https://github.com/ruflin/Elastica/blob/master/README.markdown)
- Implement tests
- Add **BIG BOLD** docblock that this query is not standard functionality of ES. Like [here](https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Document.php#L159)

lib/Elastica/Query/Image.php
im-denisenko: I think, since #774, here (and everywhere) should be `@return $this`

ruflin: :+1:

test/lib/Elastica/Test/Query/ImageTest.php
Jmoati: Thx.

lib/Elastica/Query/Image.php
Jmoati: I will replace by $this.

Jmoati:
I will try this as soon as i can.

Jmoati:
Can you please check.

ansible/roles/elasticsearch/tasks/main.yml
```diff
@@ -45,6 +45,11 @@
     state=present
     update_cache=no
 
+- name: install image plugin
+  command: >
+    creates=/usr/share/elasticsearch/plugins/image
+    /usr/share/elasticsearch/bin/plugin --url https://github.com/SibaTokyo/elasticsearch-image/releases/download/{{ ES_IMAGE_PLUGIN_VER }}/elasticsearch-image-1.4.0.zip -install image
```
im-denisenko: I guess "1.4.0" at the end of url should be variable too.

im-denisenko:
All classes in `Elastica\Query` namespace can be instantiated by `Elastica\QueryBuilder`. You can add this query to querybuilder [here](https://github.com/ruflin/Elastica/blob/master/lib/Elastica/QueryBuilder/Version/Version140.php) and [here](https://github.com/ruflin/Elastica/blob/master/lib/Elastica/QueryBuilder/DSL/Query.php).

But this query isn't part of elasticsearch, so I'm not sure should it be added to querybuilder or not. 
Guess @ruflin or @webdevsHub knows better.

ansible/roles/elasticsearch/tasks/main.yml
Jmoati: You are right.

Jmoati:
Maybe, the PR can be accepted before I implement it in QueryBuilder if necessary?

ansible/roles/elasticsearch/tasks/main.yml
Jmoati: It's fixed.

ruflin:
Merged. Yes we can add it to the query builder in a second step.

Jmoati:
Thx.

webdevsHub:
I would prefer not to add it to Query DSL class because it is not an official elasticsearch query. Furthermore it will be difficult to update the Version classes with custom queries.

However we could implement a new "plugin" DSL in QueryBuilder and put all the custom plugin stuff in there

```
$qb = new QueryBuilder();
$qb->plugin()->image(/*...*/);

// or
$qb->elastica()->image(/*...*/);
```

ruflin:
@webdevsHub I like the "plugin DSL" idea.

webdevsHub:
I will look into that during weekends. I want to add some comments to the
QB unit tests anyway.

Are there any other plugin related queries/aggs already implemented in
Elastica that should be added to the new QB part?

ruflin:
All the plugins are listed here in the README: https://github.com/ruflin/Elastica

Geocluster and attachment are the other two.

PR #788, opened 2015-02-21
Fix fluent interface inconsistency

im-denisenko:
- All methods that supposed to be fluent now return instance from here they were called.
- Some tests were added.
- `@return` docblock updated across entire library. Finishes work started in #774.
- `@param` and `@return` realigned using php-cs-fixer.

### DIFF
--- a/lib/Elastica/Bulk/Action/AbstractDocument.php
+++ b/lib/Elastica/Bulk/Action/AbstractDocument.php
@@ -23,7 +23,7 @@ abstract class AbstractDocument extends Action
     }
 
     /**
-     * @param  \Elastica\Document                     $document
+     * @param  \Elastica\Document $document
      * @return $this
      */
     public function setDocument(Document $document)
@@ -38,7 +38,7 @@ abstract class AbstractDocument extends Action
     }
 
     /**
-     * @param  \Elastica\Script                       $script
+     * @param  \Elastica\Script $script
      * @return $this
      */
     public function setScript(Script $script)
@@ -56,8 +56,9 @@ abstract class AbstractDocument extends Action
     }
 
     /**
-     * @param  \Elastica\Script|\Elastica\Document    $data
      * @throws \InvalidArgumentException
+     *
+     * @param  \Elastica\Script|\Elastica\Document $data
      * @return $this
      */
     public function setData($data)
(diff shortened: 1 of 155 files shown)


lib/Elastica/Bulk/Action/AbstractDocument.php
```diff
@@ -114,9 +115,9 @@ public function getData()
     abstract protected function _getMetadata(AbstractUpdateAction $source);
 
     /**
-     * @param  \Elastica\Document|\Elastica\Script    $data
-     * @param  string                                 $opType
-     * @return \Elastica\Bulk\Action\AbstractDocument
+     * @param  \Elastica\Document|\Elastica\Script $data
+     * @param  string                              $opType
+     * @return static
```
ruflin: I like that :-)

lib/Elastica/Connection.php
```diff
@@ -254,9 +256,10 @@ public function getConfig($key = '')
     }
 
     /**
-     * @param  \Elastica\Connection|array $params Params to create a connection
      * @throws Exception\InvalidException
-     * @return \Elastica\Connection
+     *
+     * @param  \Elastica\Connection|array $params Params to create a connection
+     * @return self
```
ruflin: :+1:

lib/Elastica/Query/Fuzzy.php
```diff
@@ -64,7 +63,7 @@ public function setFieldOption($param, $value)
         $keyArray = array_keys($params);
         $params[$keyArray[0]][$param] = $value;
 
-        return $this->setparam($keyArray[0], $params[$keyArray[0]]);
```
ruflin: Good spot ...

ruflin:
Awesome. Thanks.

ruflin:
@im-denisenko BTW: I think you will like this: https://scrutinizer-ci.com/g/ruflin/Elastica/?branch=master

im-denisenko:
I quickly looked at reported issues, and yeah, it's awesome :)

Though, many of them seems to be false-positive due to incorrect phpdoc comments (this tool gets variable types from docblocks, as I can guess)

Speaking of that, I see 1527 warnings every time I trying to run phpdoc. I already wanted to fix that, so now there yet another reason to do this.

ruflin:
:-)

PR #795, opened 2015-03-04
[RFR] Escape new symbols in Util::escapeTerm (fixes #794)

### LINKED ISSUE #794: Extra characters to be escaped
As of ElasticSearchs documentation (http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#_reserved_characters) the characters '<' and '>' needs to be escaped. However, Lucene does not require this. I've checked out the Util class code and its not escaping those characters.

Yesterday, I had an issue on production server due to a '<' symbol in a query that was not escaped. How are we dealing with this scenario?

I've created my own escape function that adds these two symbols for now. I've also created a PR.

topikito:
As of Elastic Search documentation, "<" and ">" needs to be escaped.

### DIFF
--- a/lib/Elastica/Util.php
+++ b/lib/Elastica/Util.php
@@ -10,6 +10,7 @@ namespace Elastica;
  * @author Nicolas Ruflin <spam@ruflin.com>
  * @author Thibault Duplessis <thibault.duplessis@gmail.com>
  * @author Oleg Zinchenko <olegz@default-value.com>
+ * @author Roberto Nygaard <roberto@nygaard.es>
  */
 class Util
 {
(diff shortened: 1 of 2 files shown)


topikito:
http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#_reserved_characters

ruflin:
It makes sense to go with the elasticsearch reserved characters. I restarted the travis build as I'm not sure why it completely failed.

topikito:
All green :) I've described in more detail the isse on issue #794

ruflin:
Merged. Thx.

PR #798, opened 2015-03-12
fixes #796

### LINKED ISSUE #796: QueryBuilder not generating the right match query
Hello,

It seems that the QueryBuilder, when asked to generate a match() query, doesn't generate the right DSL.

```php
query->setQuery(
    $qb->query()->match("description", "processor");
);
```
will generate the following Elasticsearch JSON:
```json
{
    "query":{
        "match":{
            "field": {
                "description-en_US-print":"processor"
            }
        }
    }
}
```
Which is the wrong syntax, as the match query should be:
```json
{
    "query":{
        "match":{
            "description-en_US-print":"processor"
        }
    }
}
```
as defined here: http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/query-dsl-match-query.html#query-dsl-match-query

The problem seems to come from the QueryBuilder code:
https://github.com/ruflin/Elastica/blob/master/lib/Elastica/QueryBuilder/DSL/Query.php#L67

When replacing the following line:
```php
$match->setFieldParam('field', $field, $value);
```
by
```php
$match->setParam($field, $value);
```
the resulting syntax is correct and the query works well.

What do you think guys ? Is there something I'm doing wrongly or there's a problem with the QueryBuilder DSL Query part for the match query ?

Thanks !

webdevsHub:
technically it is an BC break, but that short match construction actually never worked as intended. Should I add a notice somewhere?

### DIFF
--- a/changes.txt
+++ b/changes.txt
@@ -1,5 +1,8 @@
 CHANGES
 
+2015-03-12
+- Fixed short match construction in query DSL #796
+
 2015-02-22
 - Added Elastica\Query\Image
 
--- a/lib/Elastica/QueryBuilder/DSL/Query.php
+++ b/lib/Elastica/QueryBuilder/DSL/Query.php
@@ -64,7 +64,7 @@ class Query implements DSL
     {
         if ($field !== null && $value !== null) {
             $match = new Match();
-            $match->setFieldParam('field', $field, $value);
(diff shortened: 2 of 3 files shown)


ruflin:
@webdevsHub As it never worked before, I don't think we should treat it as a BC break.

PR #799, opened 2015-03-13
Fix incorrect documentation URL in phpDoc

borNfreee:
Old URL throws 404 error.
New correct URL is http://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-geo-point-type.html


im-denisenko:
There are 303 other uses of "elasticsearch.org" across all library. Perhaps you could check them too?

You could locate them using:
```
find {lib,test} -type f -name '*.php'  | xargs grep -n 'elasticsearch.org'
```

ruflin:
Agree, would be nice to have it directly changed everywhere. @borNfreee Can you add this change?

PR #806, opened 2015-03-28
Add Elastica\Facet\DateHistogram::setFactor()

oohnoitz:
Implements [Factor](http://www.elastic.co/guide/en/elasticsearch/reference/current/search-facets-date-histogram-facet.html#_factor) for DateHistogram.

### DIFF
--- a/test/lib/Elastica/Test/Facet/DateHistogramTest.php
+++ b/test/lib/Elastica/Test/Facet/DateHistogramTest.php
@@ -28,22 +28,23 @@ class DateHistogramTest extends BaseTest
 
         $mapping = new Mapping($type, array(
                 'name' => array('type' => 'string', 'store' => 'no'),
-                'dtmPosted' => array('type' => 'date', 'store' => 'no', 'format' => 'yyyy-MM-dd HH:mm:ss'),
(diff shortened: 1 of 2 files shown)


test/lib/Elastica/Test/Facet/DateHistogramTest.php
```diff
@@ -28,22 +28,23 @@ public function testTest()
 
         $mapping = new Mapping($type, array(
                 'name' => array('type' => 'string', 'store' => 'no'),
-                'dtmPosted' => array('type' => 'date', 'store' => 'no', 'format' => 'yyyy-MM-dd HH:mm:ss'),
+                'dtmPosted' => array('type' => 'long', 'store' => 'no'),
```
ruflin: Wouldn't it make sense to keep this test as it is and add a new one? Liek this we now that both options work. Or what is the reason you modified this test?

oohnoitz: Yeah. I'll keep the original test and add a new one instead to check that `factor` works.

ruflin:
Merged. thx.

PR #814, opened 2015-04-10
MLT query against documents

nurikabe:
Support for "More Like This" from documents.  (Goes back to an old conversation from 2011/2012: https://groups.google.com/forum/#!topic/elastica-php-client/1l9s80v0KMo for the original conversation)


ruflin:
Thx. Merged.

PR #816, opened 2015-04-15
deleteByQuery() implemented in Elastica\Index

rmruano:
I've noticed that `deleteByQuery()` is only available at `Elastica\Type`, for convenience I've implemented `Elastica\Index::deleteByQuery()` and I've added the same tests that `Elastica\Type` has but with a few modifications to validate that documents from all types of the index are successfully deleted.

### DIFF
--- a/changes.txt
+++ b/changes.txt
@@ -1,5 +1,8 @@
 CHANGES
 
+2015-04-15
+- deleteByQuery() implemented in Elastica\Index
+
 2015-03-29
 - Added Elastica\Suggest\Completion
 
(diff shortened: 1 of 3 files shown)


ruflin:
Thx

tmatei:
@ruflin Any way you can bump the version?

ruflin:
@tmatei Not sure what you mean. There were already 2 releases with the changes above.

tmatei:
@ruflin sorry my bad, I was still on the 1.4.x, didn't notice you moved to 2.x

Issue #825, opened 2015-04-27
The library is incompatible with PHP 7 because of reserved class names


stof:
``Bool`` is reserved in PHP 7. See https://travis-ci.org/Seldaek/monolog/jobs/60151678 for example

You can have a look at https://github.com/symfony/symfony/pull/14228 to see how Symfony will handle it with a BC layer on PHP 5

ruflin:
I also planned to look into this "soonish". Did you already try to run it with PHP 7? We should also create a list with the things that must be adjusted.

stof:
Well, Monolog runs its own testsuite against PHP 7 on Travis, and it has an integration with Elastica which broke the testsuite. See the link above.

The easy way to spot PHP 7 incompatibilities would be to enable testing on PHP 7 on Travis.

ruflin:
Thanks for the pull request. I wasn't aware that Travis already supports PHP 7.

ruflin:
It is probably best to follow the suggestion of @im-denisenko concerning the naming in #826. @stof Is Null also affected? 

What we could do to keep it BC is to set BoolFilter as the base class and use it across all the client library. But keep the Bool class which extends Bool*. So it will never actively used in the library itself but "could" be used before PHP 7. I'm not sure if PHP 7 will complain also if it is not called.

ruflin:
Pull request opened: https://github.com/ruflin/Elastica/pull/837

http://php.net/manual/en/reserved.other-reserved-words.php

stof:
@ruflin if it is never used, PHP 7 will not complain (as the file won't be loaded)

ruflin:
@stof That is what I assume. It is time to get a local PHP 7 environment running for testing this stuff. Which docker or vagrant images are you using for php7?

stof:
I don't have a PHP 7 setup yet. I relied on Travis until now.

digitalkaoz:
you could go with rasmus box https://github.com/rlerdorf/php7dev

ruflin:
@digitalkaoz Thx for the link. Will try it out.  There are also some Docker images available: https://registry.hub.docker.com/search?q=php7&searchfield=

im-denisenko:
Fixed via #826, #837, #868

PR #826, opened 2015-04-29
Add testing on PHP 7 on Travis

stof:
This is the first step for #825: running tests on PHP 7 (with failures allowed as the library is not compatible yet).
memcache.so is not available for PHP 7 on Travis currently, so enabling it is skipped (as done for HHVM currently).

I als removed a few useless sudo usages for Composer calls

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,13 @@ php:
   - 5.4
   - 5.5
   - 5.6
+  - 7
   - hhvm
 
 matrix:
     allow_failures:
         - php: 5.3
+        - php: 7
         - php: hhvm
     fast_finish: true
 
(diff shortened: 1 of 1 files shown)


im-denisenko:
Sudo isn't useless. Ansible runs as root and owner for vendor directory is root also. There must be either `chown vendor` or `sudo composer`.

stof:
hmm indeed.

It may be better to limit the usage of sudo on Travis to be able to use their new infrastructure though. I will revert the sudo removal.

ruflin:
@stof @im-denisenko I'm a little bit undecided if I should merge the pull request already as it will still fail for quite some time or if we should wait until there are also the changes for PHP 7 compatibility.

@stof About the new travis infrastructure: I'm still hoping that they will support docker containers which would heavily speed up the builds.

stof:
@ruflin regarding the change for PHP 7, the first question is to decide what the new name of these ``Bool`` classes should be (both the query and the filter). It is easy to keep BC for PHP 5.x users with a class_alias (on PHP 7, the only option is to use the new name as ``Bool`` is forbidden)

im-denisenko:
@ruflin I tried to run tests on php 7 and got some weird errors, like segmentation fault in http transport test, which could be not related to Elastica and not be fixed soon. Hence I guess, it could be merged already.

About broken classes. They were named `Filter\BoolFilter`, `Query\BoolQuery` and `Transport\NullTransport` until #303. We could just rename them back, though it damages naming consistency.

ruflin:
I merged the pull request so we know what the status is. I suggest we move all further discussions to https://github.com/ruflin/Elastica/issues/825

Issue #827, opened 2015-05-01
Scroll does not respect ordered results by default


hasumedic:
According to elasticsearch's documentation:
http://www.elastic.co/guide/en/elasticsearch/reference/1.5/search-request-scroll.html#scroll-scan
> "Normally, you just want to retrieve all results and the order doesn’t matter. Scrolling can be combined with the scan search type to disable any scoring or sorting and to return results in the most efficient way possible. All that is needed is to add search_type=scan to the query string of the initial search request:"

I understand that this means that scoring and sorting are being taken into account by default.

This conflicts with the current implementation of the ScanAndScroll.php class, where the first query to obtain the "scroll_id" is made:

```php
    /**
     * Start the initial scan search
     * @link http://php.net/manual/en/iterator.rewind.php
     * @throws \Elastica\Exception\InvalidException
     * @return void
     */
    public function rewind()
    {
        $this->_search->getQuery()->setSize($this->sizePerShard);

        $this->_search->setOption(Search::OPTION_SEARCH_TYPE, Search::OPTION_SEARCH_TYPE_SCAN);
        $this->_search->setOption(Search::OPTION_SCROLL, $this->expiryTime);

        // initial scan request
        $this->_setScrollId($this->_search->search());

        // trigger first scroll request
        $this->_scroll();
    }
```

I would think that the setOption for the search type should be optional and default to none.

Am I missing something? I'm fairly new to this repository, but if I can get some pointers on how this could be fixed without breaking BC I'm more than happy to do a PR.

hasumedic:
After few stabs at the problem (#828), I think I came to the conclusion that what's actually happening is that the library is missing the implementation of Scroll all together. Scroll and Scan-Scroll are different things, with different behaviours, meanings and results.

webdevsHub:
That is correct: `ScanAndScroll` class is dedicated to http://www.elastic.co/guide/en/elasticsearch/guide/current/scan-scroll.html and should be used to reindex or export whole indexes (see https://github.com/ruflin/Elastica/pull/617#issuecomment-43440568)

I could not see any advantage of using `scroll` over `from`, `size` with normal queries when I wrote the Iterator, but actually it offers some few advantages:

* faster (Quote: "The scroll API keeps track of which results have already been returned and so is able to return sorted results more efficiently than with deep pagination")
* retryable (send an old scroll id to get the old results)
* removable context (useless in my opinion)

[...]

PR #828, opened 2015-05-05
Added option to set behaviour for scroll in regards to sorting results #827

hasumedic:
This PR should allow the library user to choose whether they want to change the default's behaviour of Elastica when dealing with Scrolls in regards to sorting.

The changes should be backwards compatible, although I'd agree that it might look a bit odd since Elastica is defaulting to completely the opposite to what ES is defaulting to.

Any feedback appreciated.

### DIFF
--- a/changes.txt
+++ b/changes.txt
@@ -1,5 +1,8 @@
 CHANGES
 
+2015-05-05
+- Added option to choose whether the ordering of results should be taken into account when Scrolling #827
+
 2015-04-23
 - Allow bool in Query::setSource function #818 http://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-source-filtering.html
 - Fix empty bool query to act as match all query #817
--- a/lib/Elastica/ScanAndScroll.php
+++ b/lib/Elastica/ScanAndScroll.php
@@ -45,17 +45,31 @@ class ScanAndScroll implements \Iterator
     protected $_currentResultSet = null;
 
     /**
+     * By default, ES's scroll-scan respects sorting on the result set although with a performance cost.
+     * Elastica ignores by default sorting on scroll for performance reasons, but this flag allows to do
+     * not ignore it when needed
+     *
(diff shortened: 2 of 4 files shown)


ruflin:
Should we change to the es default and just make it an option? This would break BC but if it makes it more understandable, why not? What would the engineer expect? What do you think?

hasumedic:
I would consider this change a minor BC break. So according to strict semver, this should go into 2.0.x. But some projects (like Doctrine2) also accept these kind of BC breaks into their minor versions, so it could go into 1.5.x.
What's the next tag you're thinking from master? 1.4.4 or 1.5.0? Depending on that I can adapt this PR to be BC (or not).

ruflin:
Next tag planned is 2.0.0. This will also be the move to "proper" semantic versioning. Currently Elastica has 4 digits and is linked to the es versioning.

hasumedic:
Oh good! So let's break things! :)

I can keep this branch to get it merged into maintenance branches (potentially 1.4.4?).

I'll do another one with the non-BC changes for 2.0.0.

hasumedic:
Actually, I misunderstood what the purpose of the class ScanAndScroll is. It implements exactly what it's described in this section of the documentation:
http://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-scroll.html#scroll-scan

What I think the library is missing is the implementation of the class Scroll. This class would work very similarly to how the ScanAndScroll class works, but with a couple of subtle differences:
- It does not specify a type_search to the Search, so it respects sorting and scoring
- The "size" parameter is not based on per a shard basis, but on a result's total basis
- The first request to scroll already contains the first subset of results

I reckon that this PR can be rejected and that I should work on the Scroll implementation instead. Thoughts?

webdevsHub:
> I reckon that this PR can be rejected and that I should work on the Scroll implementation instead.

:thumbsup:

ruflin:
:+1: I now closed this pull request.

PR #829, opened 2015-05-06
add reindex function in Elastica/Util

xwei3752:
add reindex function in Elastica/Util by scan and bulk way
my project always change,so this function handy for me.i think others will need it too.


ruflin:
Thx. Merged.

PR #837, opened 2015-05-11
PHP7 Compatibility: Rename Bool to BoolQuery and BoolFilter

ruflin:


stof:
There is one missing part: the BC layer with the old class names

ruflin:
Jep, working on this one. I want to first check if it builds with renamed classes as my PHP 7 setup is not ready yet.

stof:
@ruflin you should change the Travis config to trigger builds only for the master branch and PRs targetting the master branch. This would avoid having 2 builds being triggered for your own PRs, once for the PR and once for the branch push.
Given that Travis limits the number of concurrent jobs per repo owner (i.e. among all your repos) and that each build requires 12 jobs currently (ES_COMPOSER_NODEV being true or false for 6 different PHP versions), this would make the feedback faster on Travis (you currently have 16 builds in the Travis queue, with only one of them being running because of the restriction)

stof:
btw, you might want to cancel builds for PRs which have been updated after that, keeping only the latest build for each PR to make the queue smaller

ruflin:
@sten Jep. I already canceled most of the builds. It worked in the past as there were less jobs and it seems like travis gets more and more busy for open source projects: https://www.traviscistatus.com/ I'm also working on making the builds less heavy, especially the setup.

stof:
@ruflin is it really needed to run both values of ES_COMPOSER_NODEV for each PHP version ?

ruflin:
@stof There were some differences in the past where this was needed (and we had less PHP versions). But I will also look into this during checking for build process improvements.

ruflin:
Seems like there are some issue with phpunit and PHP 7. Not sure what causes the segmentation fault: https://travis-ci.org/ruflin/Elastica/builds/62187485 To debug it I must setup my local PHP 7 version.

stof:
@ruflin I suggest changing the coveralls config to report the status through the github PR status instead. Getting email notifications each time they build it is annoying (especially when they tell us they don't know what they should say)

ruflin:
@stof I disabled comments for the moment. What I like about the comments is that it shows the coverage and the coverage changes. I'm testing codecov at the moment as there seem to be some issues with coveralls. https://github.com/ruflin/Elastica/pull/850

PR #841, opened 2015-05-14
Curl connect timeout

rmruano:
CURLOPT_CONNECTTIMEOUT added to the HTTP transport curl options to allow connections to unresponsive nodes to fail as soon as possible. The timeout can be customized with the connectTimeout parameter (default of 5 seconds).

### DIFF
--- a/lib/Elastica/Connection.php
+++ b/lib/Elastica/Connection.php
@@ -38,6 +38,12 @@ class Connection extends Param
     const TIMEOUT = 300;
 
     /**
+     * Number of seconds after a connection timeout occurs for every request
+     * Use a small value if you need a fast response in case of dead servers.
+     */
+    const CONNECT_TIMEOUT = 5;
+
+    /**
      * Creates a new connection object. A connection is enabled by default
      *
      * @param array $params OPTIONAL Connection params: host, port, transport, timeout. All are optional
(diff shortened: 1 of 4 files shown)


lib/Elastica/Connection.php
```diff
@@ -38,6 +38,12 @@ class Connection extends Param
     const TIMEOUT = 300;
 
     /**
+     * Number of seconds after a connection timeout occurs for every request
+     * Use a small value if you need a fast response in case of dead servers.
+     */
+    const CONNECT_TIMEOUT = 5;
```
ruflin: This means setting the timeout for all clients that don't implement it to 5 seconds. Isn't the default (without setting it) 300? So it would make sense to keep the behaviour the same as before to set it to 300? Or what is the default currently used?

rmruano:
Indeed, I've been too aggressive with the default. The idea behind of this is to prevent stacking thousands of connections in high demanded sites in case of cluster outage or unresponsivenes, firewall rules or simple issues with the connection. Recently we had a complete outage because of the default behavior which according to my research is 0 (no connection timeout). Please notice that this timeout only applies during the connection phase: If your client is unable to open the socket in 5 seconds, then something very bad is going on.

Anyway, we've got 3 options: 
 - Increasing the default to 300 seconds to match the standard timeout.
 - Increasing the default to a less aggressive value value (PHP default_socket_timeout property is 60 sec)
 - Setting the default value to null and ignore the parameter if it's not been explicitly set. 

What do you think?

ruflin:
@rmruano I think 5 seconds is totally reasonable. But we should not change the default behaviour. So my favorite options is option number 3. Set it to null so in case the default value is changed by curl, it will also change.

rmruano:
Done. I've opted to set it to 0 instead of null since it's [how curl treats the option](http://curl.haxx.se/libcurl/c/CURLOPT_CONNECTTIMEOUT.html)

ruflin:
Merged

Issue #783, opened 2015-02-17
Elastica/Query params converted to array on setters


ewgRa:
I have case, when Query must be created by two classes, for example there is BaseFooClass that set some basic logic like CompanyId match or Enabled match. And next one FooClass that add to this query additional logic, like Category match for example.

Problem is that when BaseFooClass create query and set Filter to Query, this filter cast and stored as array and next FooClass can't just get "getPostFilter" and add there some additional logic.

```php
    public function setPostFilter($filter)
    {
        if ($filter instanceof AbstractFilter) {
            $filter = $filter->toArray();
        } else {
            trigger_error('Deprecated: Elastica\Query::setPostFilter() passing filter as array is deprecated. Pass instance of AbstractFilter instead.', E_USER_DEPRECATED);
        }
```

In this case I set object, but get array back and can't build query step by step in different places.

Can someone describe me a way to achieve desired in my case with current Elastica code? For now I create something like custom decorator that finally build Query with calling method "toQuery" for example, but I want have this functionality in Elastica.

I understand that this is a BC changes, but maybe there is possible to have decorator, or QueryBuilder (not DSL, DSL not working in this case).

I can create pull request for this, but first need feedback to understand is it is a good idea.

ewgRa:
Or the best way implement ArrayAccess interface to Param and another classes? And in this case we haven't BC but can work with queries parts as objects.

ruflin:
@ewgRa I understand the issue. In case we change this, we would have to change this across the hole library I think. Because this "issue" exists for almost any object. Most of the time the objects are directly transformed to arrays to not have to deal with it at a later stage. I like the idea with ArrayAccess, but perhaps it could make even more sense to use ArrayObject as a foundation? 

We should discuss the pros and cons of the different options we have in detail as this would mean quite a radical change to the library.

ewgRa:
ArrayAccess and ArrayObject will be also BC, or than it must be cloned, because can be such code:
$filter = new FIlter()
$query->setPostFilter($filter);
$filter->addSomething...
in current version in $query will be empty filter, but when we remove toArray - there will be object and in query will be not empty object.
I will write test for this later and make a pull request for this case to fix this in tests.

On my mind there is two options:

1. Create new class that will be decorator of Query, and on final call make DecoratorQuery->toQuery, on "search" methods add "if $query instanceof DecoratorQuery then call toQuery".
Props: 
    - no BC
[...]

PR #842, opened 2015-05-14
added Scroll Iterator

webdevsHub:
Implementation of `Scroll` Iterator requested by @hasumedic in https://github.com/ruflin/Elastica/issues/827

The `Scroll` Iterator executes every request fully isolated. That means search options manipulated during iterations are changed and reverted automatically.

```php
$search->setOption(Search::OPTION_SEARCH_TYPE, Search::OPTION_SEARCH_TYPE_COUNT);

$path = $search->getPath();
foreach ($search->scroll() as $scrollId => $resultSet) {
    $search->getPath() === $path; // returns true
}
$search->getPath() === $path; // returns true
```

In addition I changed the `ScanAndScroll` Iterator to a child class of `Scroll` in order to profit from that isolation.

Finally I added a `Search::scroll()` function just like `Search::scanAndScroll()`

### DIFF
--- a/lib/Elastica/ScanAndScroll.php
+++ b/lib/Elastica/ScanAndScroll.php
@@ -3,49 +3,21 @@
 namespace Elastica;
 
 /**
- * scan and scroll object
+ * Scan and Scroll Iterator
  *
- * @category Xodoa
- * @package Elastica
  * @author Manuel Andreo Garcia <andreo.garcia@gmail.com>
+ *
  * @link http://www.elastic.co/guide/en/elasticsearch/guide/current/scan-scroll.html
  */
-class ScanAndScroll implements \Iterator
+class ScanAndScroll extends Scroll
 {
-    /**
-     * time value parameter
-     *
-     * @var string
-     */
-    public $expiryTime;
-
     /**
      * @var int
      */
     public $sizePerShard;
 
     /**
-     * @var Search
-     */
-    protected $_search;
-
-    /**
-     * @var null|string
-     */
-    protected $_nextScrollId = null;
-
-    /**
-     * @var null|string
-     */
-    protected $_lastScrollId = null;
-
-    /**
-     * @var null|ResultSet
-     */
-    protected $_currentResultSet = null;
-
-    /**
-     * Constructs scroll iterator object
+     * Constructor
      *
      * @param Search $search
      * @param string $expiryTime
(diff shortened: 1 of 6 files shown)


lib/Elastica/ScanAndScroll.php
```diff
@@ -53,104 +25,57 @@ class ScanAndScroll implements \Iterator
...
-     * @throws \Elastica\Exception\InvalidException
-     *
-     * @return void
      */
     public function rewind()
     {
-        $this->_search->getQuery()->setSize($this->sizePerShard);
+        // reset state
+        $this->_nextScrollId = null;
+        $this->_options = array(null, null, null, null);
```
ruflin: shouldn't it be only 3 null?

ruflin: Nevermind, I just saw that there is a fourth param size for ScanAndScroll

ruflin:
Does this break an BC or not?

webdevsHub:
> Does this break an BC or not?

no

webdevsHub:
I would like hearing @hasumedic opinion on this before merge

ruflin:
Ok, I will wait with merging until @hasumedic put his feedback in.

hasumedic:
Good stuff @webdevsHub! I'll have a look at it at Today at lunch time.

lib/Elastica/Scroll.php
```diff
@@ -0,0 +1,175 @@
...
+    protected $_currentResultSet = null;
+
+    /**
+     * 0: scroll<br>
+     * 1: scroll id<br>
+     * 2: search type.
+     *
+     * @var array
+     */
+    protected $_options = array(null, null, null);
```
hasumedic: Should this be a VO instead?

webdevsHub: > Should this be a VO instead?

VO?

hasumedic: I was wondering if we should be using Value Objects for these pieces of config. But I haven't seen them implemented anywhere else in the library, so I would assume that arrays are the conventions in this case. Isn't that the case @ruflin?

ruflin: @hasumedic Currently the convention is arrays. This is mainly related to the history of Elastica as it started very simple and grew together with elasticsearch. If we change to ArrayObjects, we should consider a change across the full library. @ewgRa brought up this issue and I just realised I never responded to it :-( https://github.com/ruflin/Elastica/issues/783#issuecomment-75366015 I suggest to continue this discussion in #783 an until then we use arrays.

hasumedic: That being the case, this PR looks good and should go forward! :+1:

webdevsHub:
I deleted `protected $_lastScrollId` because it is useless. This would be an BC break if someone extended `ScanAndScroll` and used that member. Should I mentioned it somewhere?

ruflin:
Mention it in the CHANGELOG.md file

ruflin:
Merged. Thx.

hasumedic:
@webdevsHub just a quickie. 

How would this API work in a cross-request scenario? ie. First request loads the data and returns the scrollId by means of Scroll::key(). Second request sends the scrollId to fetch the next bulk of data. How is this scrollId set into the Scroll object to query ES for the next set of results?

ruflin:
@hasumedic I think @webdevsHub is on vacation and will only respond next week ...

hasumedic:
@ruflin that's OK. We'll figure it out then :) Thanks for letting me know!

webdevsHub:
@hasumedic scroll id is set in https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Scroll.php#L77 and `protected $_nextScrollId` in https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Scroll.php#L142

hasumedic:
Thanks for replying @webdevsHub ! I guess that that's precisely what my question was.
This works fine in an iterable scenario, but it prevents others such as multiple requests to work.

What I was asking in my previous comment was in terms of web requests. So you make a first web request to process a first batch of (lets say) 1000 elements. This generates the scrollId and the first 1000 results. With this you generate a response.
Now you want to manually trigger the process for the next 1000 elements, so you make a new web request. In order to continue processing from the element 1001 to the 2000, you need the scrollId to be passed into the Scroll object and run Scroll:next(), but I don't see how this can work at the moment. 

Does this make any sense?

webdevsHub:
As you can see [here](https://github.com/ruflin/Elastica/blob/master/test/lib/Elastica/Test/ScrollTest.php#L71) it is perfectly fine to use the Iterator without foreach and without setting something manually in the class.

hasumedic:
I'll give it a go. Thanks!

PR #847, opened 2015-05-17
SignificantTerms aggregation

rmruano:
SignificantTerms aggregation is almost identical to Terms Aggregation. Extended it and added a convenience method for setting a background filter. Simple tests added, although they may not be very good examples: SignificantTerms doesn't shine on automatically generated datasets.

### DIFF
--- a/lib/Elastica/Aggregation/SignificantTerms.php
+++ b/lib/Elastica/Aggregation/SignificantTerms.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace Elastica\Aggregation;
+use Elastica\Filter\AbstractFilter;
+
+/**
+ * Class SignificantTerms
+ * @package Elastica\Aggregation
+ * @link http://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-significantterms-aggregation.html
+ */
+class SignificantTerms extends AbstractTermsAggregation
+{
+
+    /**
+     * The default source of statistical information for background term frequencies is the entire index and this scope can
+     * be narrowed through the use of a background_filter to focus in on significant terms within a narrower context
+     * @param AbstractFilter $filter
+     * @return $this
+     * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-significantterms-aggregation.html#_custom_background_context
+     *
+     */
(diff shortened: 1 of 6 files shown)


lib/Elastica/Aggregation/SignificantTerms.php
```diff
@@ -0,0 +1,24 @@
...
+{
+
+    /**
+     * The default source of statistical information for background term frequencies is the entire index and this scope can
+     * be narrowed through the use of a background_filter to focus in on significant terms within a narrower context
+     * @param AbstractQuery $query
+     * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-significantterms-aggregation.html#_custom_background_context
+     */
+    public function setBackgroundFilter(AbstractQuery $query) {
+        $this->setParam("background_filter", $query->toArray());
```
im-denisenko: Missed `return`

im-denisenko:
@rmruano Could you look into my eyes and say that every existing parameter of Terms aggregation can be applied to SignificantTerms? If it's not, class should be extended from AbstractSimpleAggregation.

lib/Elastica/Aggregation/SignificantTerms.php
```diff
@@ -0,0 +1,24 @@
...
+class SignificantTerms extends Terms
+{
+
+    /**
+     * The default source of statistical information for background term frequencies is the entire index and this scope can
+     * be narrowed through the use of a background_filter to focus in on significant terms within a narrower context
+     * @param AbstractQuery $query
+     * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-significantterms-aggregation.html#_custom_background_context
+     */
+    public function setBackgroundFilter(AbstractQuery $query) {
```
im-denisenko: According to parameter name, shouldn't here be `AbstractFilter` indeed?

rmruano:
@im-denisenko Thanks for your feedback. Apparently, the only missing feature is `order`, to overcome this without duplicating code, I've created a common AstractTermsAggregation that both Terms & SignificantTerms. Please take a look and let me know if you're ok with it.

I've also fixed SignificantTerms::setBackgroundFilter() with your suggestions.

im-denisenko:
Merged, thanks!

Issue #832, opened 2015-05-11
Speed up tests


ruflin:
The test runs should be speed up to have every build below 3 minutes if possible. There are several steps that can be taken to improve it lake the setup of the machines or tear down to tests. The setup could be done with Docker images (not supported by Travis yet).

ruflin:
@im-denisenko You were the one that implemented https://github.com/ruflin/Elastica/blob/master/test/lib/Elastica/Test/Base.php#L50 I see the reasoning behind it and it makes sense. There are two potential problems here and I would like to get your opinion on that. The first problem is that I "assume" that cleaning all indexes takes longer then just the ones that were created. On the other side, the same amount of indexes should exist as was created. The second problem is, that some let the tests run on the same cluster as their staging or even production environment (yes, they shouldn't). All these indexes would also be removed. The good thing here is they will learn not to do it. At the same time I think tests should only mess with the data / indexes they created. What are your thoughts here?

im-denisenko:
There are two reasons why I delete all indices instead of just created.

First is that index could be created not only by [this method](https://github.com/ruflin/Elastica/blob/master/test/lib/Elastica/Test/Base.php#L24). If we want to really remove all indices that were created during tests, we should find all existing indices in setUp, find all existing indices in tearDown, diff these lists, and then delete the difference. This is will be even slower than now.

Second is that ES should always have clean state to run tests on it. If someone run Elastica's tests on their cluster where some indices already exists, how you could garantee that tests will not randomly failed or that their indices will be unbroken? I even don't mention that last test in Elastica testsuite shuts down entire cluster.

im-denisenko:
We could split tests by their group and don't perform complete removal on every test.

I see two groups there: *unit* and *functional*. Unit is pure test where ES is not involved. Functional is about ES integration. There are no reason to delete all indices from ES for tests from *unit* group.

This separation could be done either using [phpunit groups](https://phpunit.de/manual/current/en/appendixes.annotations.html#appendixes.annotations.group), or by two test's trees instead of just one.

ruflin:
Good idea. I would probably not change the test trees as it will "duplicate" a lot of classes. About the speed gain / loss I agree. Currently the slowest past is probably the setup. Unfortunately there is no Docker support from Travis yet.

ruflin:
About clean es instance: Agree that we should expect that to have. So lets keep removing all indices and not lets add complexity here.

im-denisenko:
[...]

PR #848, opened 2015-05-17
Split tests into different groups

im-denisenko:
Was discussed in https://github.com/ruflin/Elastica/issues/832#issuecomment-101819800.

This PR adds one of *unit*, *functional* or *shutdown* group to every existing test.
It has lot of changes, because almost every setUp method was rewrited to avoid hit ES in tests from *unit* group.
Also, it has some random cleanups like "replace multiple calls of addDocument to one addDocuments".

Now travis will run *functional* tests first, then *shutdown*, and then *unit*. 
It means that unit test can't make any request, because ES cluster is down at the time they runned.

Need to say, I'm quite disappointed about speed increase. I was hoping it'll be more than one minute.

@ruflin review please?


im-denisenko:
@coveralls you're kidding :)

.travis.yml
```diff
@@ -24,16 +24,17 @@ install:
   - /bin/bash ansible/provision.sh
 
 before_script:
-  - mkdir -p build/logs
+  - mkdir -p build/logs build/coverage
```
ruflin: Based on the new Makefile I suggest we "simplify" the travis.yml file by directly using the make commands (https://github.com/ruflin/Elastica/pull/852). Here for example make prepare

.travis.yml
```diff
@@ -24,16 +24,17 @@ install:
...
   - sh -c 'if [ "$TRAVIS_PHP_VERSION" != "hhvm" ] && [ "$TRAVIS_PHP_VERSION" != "7" ]; then echo "extension=memcache.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini; fi;'
   - if [ $ES_COMPOSER_NODEV == "no" ] && ! $(phpenv version-name | grep -q '5.3'); then sudo composer require "guzzlehttp/guzzle" "4.2.*" --dev --no-ansi --no-progress --no-interaction; fi
 
 script:
-  - phpunit -c test/phpunit-travis.xml
+  - phpunit -c test/ --coverage-clover build/coverage/clover-functional.xml --group functional
```
ruflin: make phpunit

.travis.yml
```diff
@@ -24,16 +24,17 @@ install:
...
+  - phpunit -c test/ --coverage-clover build/coverage/clover-shutdown.xml --group shutdown
+  - phpunit -c test/ --coverage-clover build/coverage/clover-unit.xml --group unit
 
 after_script:
   - cat /var/log/elasticsearch/*.log
   - cat /var/log/nginx/*.log
-  - cat build/logs/phpunit-tap.log
-  - sudo composer require satooshi/php-coveralls --no-ansi --no-progress --no-interaction
-  - php vendor/bin/coveralls -v
+  - sudo rm composer.lock && sudo composer require satooshi/php-coveralls dev-master --no-ansi --no-progress --no-interaction
```
ruflin: * is it necessary to remove the .lock file?
* Does the phpunit output still exist?

test/lib/Elastica/Test/Base.php
```diff
@@ -89,10 +89,42 @@ protected function _waitForAllocation(Index $index)
...
+    {
+        parent::setUp();
+
+        $hasGroup = $this->_isUnitGroup() || $this->_isFunctionalGroup() || $this->_isShutdownGroup();
+        $this->assertTrue($hasGroup, 'Every test must have one of "unit", "functional" or "shutdown" group');
+    }
+
     protected function tearDown()
     {
+        if ($this->_isFunctionalGroup()) {
```
ruflin: That is cool :+1:

ruflin:
Looks good to me. I see you also did some cleaning up. Thx.

See my comments above.

ruflin:
@im-denisenko I suggest to merge this and make improvements in a second step. About the speed up: The nice part is now that the tests can be split up and only parts can be run. For example for each commit unit tests could be run, but only pull requests all tests are run ... Unfortunately I don't know if such configuration is possible with travis.

.travis.yml
im-denisenko: * composer.lock causes composer error when ES_COMPOSER_NODEV is "yes". [Like this](https://travis-ci.org/im-denisenko/Elastica/jobs/62861662#L2177).
* Yep, I remove it. Now phpunit has verbose enough configuration, not like it was in https://github.com/ruflin/Elastica/commit/784049df4dde462ccea45c81f97010ed36191c4a

im-denisenko:
> Unfortunately I don't know if such configuration is possible with travis.

Travis has environment variable TRAVIS_PULL_REQUEST which holds PR id for PR build or false otherwise.

ruflin:
Merged :-)

ruflin:
Interesting. We should then run the unit tests for every commit but functional and shutdown only for pull requests.

PR #851, opened 2015-05-18
Cardinality properties

hasumedic:
This PR introduces the ability to set a couple of properties to the "Cardinality" aggregation as per ES documentation:

https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-cardinality-aggregation.html

The 2 options are:
* precision_threshold (integer)
* rehash (boolean)

Both setters control that the parameter given is of the right type.

### DIFF
--- a/test/lib/Elastica/Test/Aggregation/CardinalityTest.php
+++ b/test/lib/Elastica/Test/Aggregation/CardinalityTest.php
@@ -78,4 +78,47 @@ class CardinalityTest extends BaseAggregationTest
             'more-than-max' => array(40001),
         );
     }
+
+    /**
+     * @dataProvider validRehashProvider
+     * @param bool $rehash
+     */
+    public function testRehash($rehash)
+    {
+        $agg = new Cardinality('rehash');
+        $agg->setRehash($rehash);
+
+        $this->assertNotNull($agg->getParam('rehash'));
+        $this->assertInternalType('boolean', $agg->getParam('rehash'));
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


test/lib/Elastica/Test/Aggregation/CardinalityTest.php
```diff
@@ -33,4 +33,92 @@ public function testCardinalityAggregation()
...
+     * @expectedException \InvalidArgumentException
+     * @param mixed $rehash
+     */
+    public function testInvalidRehash($rehash)
+    {
+        $agg = new Cardinality('rehash');
+        $agg->setRehash($rehash);
+    }
+
+    public function invalidRehashProvider()
```
ruflin: I would suggest to make the 4 "provider" functions protected as they don't need to be public.

ruflin:
Looks good, see comment. Can you update the CHANGELOG.md file?

test/lib/Elastica/Test/Aggregation/CardinalityTest.php
hasumedic: Thanks for the quick review @ruflin! The annotated data providers need to be public so they can be called using reflection by PHPUnit.
https://phpunit.de/manual/4.6/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers

ruflin: Make sense.

ruflin:
Merged. Thx.

Issue #840, opened 2015-05-14
Add helper to populate index with documents


im-denisenko:
Library doesn't provide standartized way to populate some index with set of documents, while this is very common task. As a result, every new user should implement this chunk of code again and again.

Examples are: [one](https://github.com/conjecto/nemrod/blob/master/src/Conjecto/Nemrod/ElasticSearch/Populator.php#L60), [two](https://github.com/darklow/ff-elastica-manager/blob/master/src/FF/ElasticaManager/IndexManager.php#L227), [three](https://github.com/doctrine/search/blob/master/lib/Doctrine/Search/ElasticSearch/Client.php#L69), [four](https://github.com/ecommerce/EcommerceElasticsearchBundle/blob/master/Service/ProductProvider.php#L48), [five](https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/Provider/AbstractProvider.php#L63), [six](https://github.com/mathielen/glueframework/blob/master/src/Infrastructure/Persistence/ElasticSearch/BatchSaver.php#L31), [seven](https://github.com/modulargaming/search/blob/master/classes/Task/Search/Index.php#L38), [eight](https://github.com/silverstripe-australia/silverstripe-elastica/blob/master/src/SilverStripe/Elastica/ElasticaService.php#L140), [nine](https://github.com/VeggieMeat/search_api_elasticsearch/blob/7.x-1.x/modules/elastica/includes/SearchApiElasticsearchElastica.inc#L249), etc.

I think, that #829, #830, and #836 is related to this, because "populate index from another" is just special case of more general task "populate index from somewhere".

If this issue will be resolved, we don't need Util::copy function at all. Implementation draft below explains why.

im-denisenko:
Implementation could be similar to fos elastica one, but not rely on symfony stack.

How it works:
- create instance of Populator
- implement class with interface ProviderInterface
- pass instance of provider into populator

```php
interface ProviderInterface
{
    /**
     * @return int
     */
    public function getDocumentCount();

    /**
     * @return Document
     */
    public function getNextDocument();
}
```

```php
class Populator
{
    public function __construct(Client $client);

    public function setBulkSize($size = 1000);

    public function processProvider(ProviderInterface $provider, callable $onEveryIteration = null);   
}
```

Predefined provider - uses elastica index as data source.
```php
class IndexProvider implements ProviderInterface
{
    protected $scanAndScroll;

    protected $documentCount;

    public function __construct(Index $index)
    {
[...]

PR #836, opened 2015-05-11
Rename reindex function to copy and switch params

ruflin:

Issue #830, opened 2015-05-08
Util method - is this an issue or am i missing something?


virtuman:
Here's a method in Util.php:
```
    public static function reindex(Client $client, $newIndex, $oldIndex, $expiryTime = '1m', $sizePerShard = 1000)
    {
        $search = new Search($client);
        $oldIndex = new Index($client, $oldIndex);
        $newIndex = new Index($client, $newIndex);
        
        $bulk = new Bulk($client);
        $bulk->setIndex($newIndex);
                
        $search->addIndex($oldIndex);
        $search->setOption(Search::OPTION_SEARCH_TYPE, Search::OPTION_SEARCH_TYPE_SCAN);
        $scanAndScroll = new ScanAndScroll($search, $expiryTime, $sizePerShard);
        foreach ($scanAndScroll as $resultSet) {
            $data = $resultSet->getResponse()->getData();
            foreach ($data['hits']['hits'] as $d) {
                $meta = array('_index' => $newIndex->getName() , '_type' => $d['_type'], '_id' => $d['_id']);
                $actions[] = new Action(Action::OP_TYPE_INDEX, $meta ,$d['_source']);
            }
            //$actions = new Bulk\Action();
            $bulk->addActions($actions);
            $bulk->send();
        }
        $newIndex->refresh();
        return "reindex complete";
    }
```

Lines of interest here are:
```
        foreach ($scanAndScroll as $resultSet) {
            $data = $resultSet->getResponse()->getData();
            foreach ($data['hits']['hits'] as $d) {
                $meta = array('_index' => $newIndex->getName() , '_type' => $d['_type'], '_id' => $d['_id']);
                $actions[] = new Action(Action::OP_TYPE_INDEX, $meta ,$d['_source']);
            }
            //$actions = new Bulk\Action();
            $bulk->addActions($actions);
            $bulk->send();
        }
```

$actions container here seems to be constantly appended to during all scanandscroll + hits iteration.
Wouldn't this result in results doubling up on each iteration, ie:
1. scanAndScroll is set to select 10 entries from each of 5 shards
2. during first iteration $actions will become container for 50 docs
3. addActions and send() will push these products to newIndex
4. during second iteration $actions will get another 50 added to it and will now become a container for 100 docs
5. addActions and send() will now push all 100 products
6. ... same repeats until the end resulting in $actions to hold entire dataset that was being selected by scanAndScroll to avoid out of memory issues

[...]

Issue #473, opened 2013-10-08
Add reindex method to index class


Tobion:
There should probably be a method in reindex all documents from one index into a new index. This is needed when one wants to change the mapping with no downtime. See http://www.elasticsearch.org/blog/changing-mapping-with-zero-downtime/

Several other client libraries already include that functionality:

- https://github.com/geronime/es-reindex (ruby)
- http://pythonhosted.org/pyes/references/pyes.es.html#pyes.es.ES.reindex (phyton)

Early draft. For some reason I get a Java exception inside elastic search as soon as there are documents in the old index.

```
[Elastica\Exception\ResponseException]
  MapperParsingException[mapping [product]]; nested: ClassCastException[java.
  util.ArrayList cannot be cast to java.util.Map];
```

The method is missing arguments to optionally overwrite the mapping and settings for the new index. Otherwise it would just be a copy.

```
// Elastica\Index

    /**
     *
     *
     * @return string The new index name
     *
     * @see http://www.elasticsearch.org/blog/changing-mapping-with-zero-downtime/
     */
    public function reindex()
    {
        $scrollResponse = $this->request('_search', Request::GET,
            array('query' => array(
                'match_all' => array()
            )),
            array('search_type' => 'scan', 'scroll' => '10m', 'size' => '100')
        );
        //var_dump($scrollResponse);

        $scrollId = $scrollResponse->getScrollId();
        unset($scrollResponse);

        $mapping = $this->getMapping();
        $settings = $this->getSettings()->get();

        $config = array();
        $config['mappings'] = $mapping[$this->_name];
        $config['settings'] = $settings;

        $this->_name = $this->_name . date('-Y-m-d-Gis');

        $this->create($config);

        unset($mapping, $settings, $config);

        $scrollResponse = $this->getClient()->request('_search/scroll', Request::GET,
            array(),
            array('scroll' => '10m', 'scroll_id' => $scrollId)
        );
        //var_dump($scrollResponse);

        return $this->_name;
    }
```

progrupa:
+1

michellesanver:
We just added reindexing at work using scan and scroll, I'll spend some time to extract it into Elastica\Util and make a pull request :)

ruflin:
@michellesanver Looking forward to it.

webdevsHub:
Added in https://github.com/ruflin/Elastica/issues/853 
Thanks @Tobion

PR #853, opened 2015-05-18
Added Tool\CrossIndex class

webdevsHub:
Implementation of https://github.com/ruflin/Elastica/issues/830, https://github.com/ruflin/Elastica/issues/473, https://github.com/ruflin/Elastica/pull/829, https://github.com/ruflin/Elastica/pull/836.

In unreleased https://github.com/ruflin/Elastica/blob/reindex-refactoring-to-copy/lib/Elastica/Util.php#L199 branch `reindex`/`copy` was added to `Elastica\Util` class. It does not belong there in my opinion if you look at the other functions in that class. So I created a new `Tool` namespace.

That namespace can be used by https://github.com/ruflin/Elastica/issues/840 as well (or any other future tools).

### DIFF
--- a/lib/Elastica/Tool/CrossIndex.php
+++ b/lib/Elastica/Tool/CrossIndex.php
@@ -1,0 +1,161 @@
+<?php
+
+namespace Elastica\Tool;
+
+use Elastica\Bulk;
+use Elastica\Index;
+use Elastica\Query\MatchAll;
+use Elastica\ScanAndScroll;
+use Elastica\Search;
+use Elastica\Type;
+
+/**
+ * Functions to move documents and types between indices
+ *
+ * @author Manuel Andreo Garcia <andreo.garcia@gmail.com>
+ */
+class CrossIndex
+{
+    /**
+     * Type option
+     *
+     * type: string | string[] | \Elastica\Type | \Elastica\Type[] | null
+     * default: null (means all types)
+     */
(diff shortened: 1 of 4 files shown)


ruflin:
@webdevsHub Can you briefly elaborate on what belongs into Util and what into the Tool namespace? We should document this somewhere so we can use it for future features to decide where they belong. The part I don't like too much is that we have a class with only static functions. It feels a little bit "wrong" :-)

webdevsHub:
Naming is inspired by Doctrine. `Utility` is a collection of things you need for internal Elastica coding and `Tool` for complex functionality for lib users (in Doctrine it is Pagination and the console commands). Do you have any other ideas for a good naming?

I used static like in `Util` because there is no state and therefore no need to create an Instance

```php
$crossIndex = new CrossIndex();
$crossIndex->reindex(/*...*/);
```

=> more code overhead as

```php
CrossIndex::reindex(/*...*/);
```

ruflin:
I added a note here and merged your pull request. https://github.com/ruflin/Elastica/commit/288a1ecea41cb35a133173d972e2d8637bc06fe6

webdevsHub:
stooop :) changelog missing

webdevsHub:
I am going to add it immediately

ruflin:
Missed that one. Sorry.

PR #858, opened 2015-05-21
Build improvements

ruflin:
Improve the Travis builds


.travis.yml
```diff
@@ -29,9 +29,9 @@ before_script:
   - if [ $ES_COMPOSER_NODEV == "no" ] && ! $(phpenv version-name | grep -q '5.3'); then sudo composer require "guzzlehttp/guzzle" "4.2.*" --dev --no-ansi --no-progress --no-interaction; fi
 
 script:
+  - phpunit -c test/ --coverage-clover build/coverage/clover-unit.xml --group unit
```
im-denisenko: Just FYI, if `unit` tests will be placed before `shutdown`, they potentially can hit ES server, that obviously not thing `unit` test is supposed to do.

ruflin: I see, haven’t thought of that. The best thing would be to only install and boot up elasticsearch after the unit tests were run. Perhaps we can set environment variables for each build and for the unit tests set the ES_HOST to null.

ruflin:
I finally solved this issue by just ignoring dev for PHP 5.3. The part I don't get is:

```
composer remove "guzzlehttp/guzzel"
```

This works as expected and removes the package after I reply yes.

If I use the -n option and set "discard-changes" config to true or false (I tried both ...). Nothing happens. Why is the correct parameter that it automatically answers yes during the removal without my interaction?

@stof I'm sure there is a flag or config I just missed.

.travis.yml
```diff
@@ -17,21 +17,16 @@ matrix:
 
 env:
   matrix:
```
stof: you should remove this key entirely as it is empty now

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ All notable changes to this project will be documented in this file based on the
 - Introduction of Changelog standard based on http://keepachangelog.com/. changes.txt moved to CHANGELOG.md [#844](https://github.com/ruflin/Elastica/issues/844/)
 - Make host for all tests dynamic to prepare it for a more dynamic test environment #846
 - Node information is retrieved based on id instead of name as multiple nodes can have the same name.
+- Remote NO_DEV builds from travis build matrix to speed up building. All builds include no dev packages.
```
stof: typo: ``Remove``

ruflin:
@stof Thx for the review. Will fix it. Any idea about the "remove" issue I mentioned above?

stof:
I don't know. But the first question would be to know why there are changes in the package (if it asks to discard changes, it means you did some changes first, or that the package has some weird stuff triggering changes in it on install, for instance a file listed in ``bin`` without beign executable)

ruflin:
The problem is PHP 5.3. I still want to keep the builds alive but Guzzle is not compatible with PHP 5.3. Unfortunately only skipping the tests doesn't work as it already checks the libraries in the beginning and an exception is thrown. So my idea is that instead of making special cases for all others (means installing an extra package) I want to remove Guzzle for php including the libs in the vendor directory.

im-denisenko:
@ruflin I would suggest to remove guzzle completely and lets `ivory/http-adapter` take care about alternative transports.

As option, may be even purge entire `Transport` namespace and do all requests through http adapter.

What do you think?

im-denisenko:
Though, nevermind. It's not causes so much troubles to have yet another bc.

ruflin:
@im-denisenko In the long term I agree that we should think about that, but currently it is not a priority. I think dropping of memcache and especially Thrift will cause some uproar.

Issue #299, opened 2013-01-13
Should we remove @package and @category annotations ?


lavoiesl:
Branched discussion started in #223

Are the documentation annotations `@package` and `@category` still relevant ?
Should they be removed and just mentioned in a README ?
Does phpDocumentator still requires a package when namespaces are present ?

im-denisenko:
Both `@package` and `@category` is removed since #864.

PR #864, opened 2015-05-29
Linting

ruflin:


im-denisenko:
I would suggest :
* Add `phpdoc_order` fixer to config file.
* Manually remove `@category` and close #299. `@package` already removed thanks to `phpdoc_no_package` fixer.
* Pin exact commit hash of php-cs-fixer in `require-dev` so this tool will always generate the same result as before (of course, while someone will not change pinned version)

ruflin:
Sounds good to. Are you able to "push" into my pull requests?

im-denisenko:
Yep, I can. I'll add these changes at evening.

im-denisenko:
Done.

ruflin:
Merged.

Issue #861, opened 2015-05-26
PHP 7 issue with phpunit and DEBUG constant


ruflin:
PHP 7 in combination with phpunit and a constant named DEBUG leads currently to the following output:

```
make: *** [phpunit] Segmentation fault
```

This is cause by the constant DEBUG which is used for the test environment: https://github.com/ruflin/Elastica/blob/master/test/phpunit.xml.dist#L27-L29

If the variable is renamed, the problem disappears.

Currently I'm not sure if this issue is cause by phpunit or PHP 7. I assume there is some name conflict. In all versions before PHP 7 it works without any issues. One solution is to rename the constant and make it ELASTICA_DEBUG to adjust it to make it namespace specific.

ruflin:
@webdevsHub @im-denisenko I need some fresh eyes here. Do you know about any "know" bug or issue why PHP 7 / phpunit and a constant "DEBUG" crashes everything?

webdevsHub:
I have no experience in PHP 7, maybe @sebastianbergmann can help? Would be nice!

sebastianbergmann:
`DEBUG` is not used by PHPUnit (or any of its dependencies). I'm afraid there's not much debugging I (or a PHP core dev) can do without a minimal, self-contained, reproducing script.

ruflin:
I will try to get a minimal script together to reproduce this.

im-denisenko:
Well. DEBUG constant it's not a real problem. It seems like curl is broken in certain cases.

It's why segfault happens at low level:
* When DEBUG is setted, Elastica\Util::debugEnabled will return true.

* Elastica\Transport\Http will set additional option to curl handler [here](https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Transport/Http.php#L134). 

* After that, curl_getinfo [here](https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Transport/Http.php#L155) will return key "request_header". 

* This key will be added to result array [here](https://github.com/php/php-src/blame/master/ext/curl/interface.c#L3011).

* ch->header.str is a pointer to zend_string, as defined [here.](https://github.com/php/php-src/blob/master/ext/curl/php_curl.h#L167)

* Segmentation fault means that this pointer is null at some time when it must not.

* As defined [below](https://github.com/php/php-src/blob/master/ext/curl/interface.c#L3017), when curl_getinfo called with second argument, it returns copy of pointer, not pointer itself.

[...]

PR #868, opened 2015-06-02
Fixed php7 build using wrapper for curl_getinfo

im-denisenko:
Was discussed here: https://github.com/ruflin/Elastica/issues/861#issuecomment-107118984

### DIFF
--- a/lib/Elastica/Util.php
+++ b/lib/Elastica/Util.php
@@ -191,12 +191,4 @@ class Util
 
         return $message;
     }
-
-    /**
-     * @return bool Returns true if debugging is enabled, otherwise false
-     */
-    public static function debugEnabled()
-    {
-        return defined('ELASTICA_DEBUG') && ELASTICA_DEBUG;
-    }
 }
--- a/lib/Elastica/Util.php
+++ b/lib/Elastica/Util.php
@@ -197,6 +197,6 @@ class Util
      */
     public static function debugEnabled()
     {
-        return defined('DEBUG') && DEBUG;
+        return defined('ELASTICA_DEBUG') && ELASTICA_DEBUG;
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@ All notable changes to this project will be documented in this file based on the
 
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/2.1.0...HEAD)
 
+### Backward Compatibility Breaks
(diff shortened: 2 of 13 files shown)


im-denisenko:
All is green! :)

lib/Elastica/Util.php
```diff
@@ -197,6 +197,6 @@ public static function convertRequestToCurlCommand(Request $request)
      */
     public static function debugEnabled()
     {
-        return defined('DEBUG') && DEBUG;
+        return defined('ELASTICA_DEBUG') && ELASTICA_DEBUG;
```
ruflin: I didn't rename the constant on purpose as this will break BC. If we change it, it must be mentioned as BC break in the CHANGELOG.

ruflin:
:+1: Can you update the CHANGELOG?

lib/Elastica/Util.php
im-denisenko: Hmm, haven't thought if this is BC. In that case, I would suggest to remove this `ELASTICA_DEBUG` completely, because `debug` should be an parameter of connection constructor and not some magic constant.

ruflin: Agree. I'm not even sure if "debug" is the correct name. Perhaps it is more "verbose" or something like that.

im-denisenko: Though, only difference I see between "debug mode" and not "debug mode" is just two things: 
* Would `$response->setQueryTime` be invoked after request.
* Would curl option `CURLINFO_HEADER_OUT` be applied or not.

Both of them has no impact on performance, is this really worth to keep `debug` option?

ruflin: I remember it was more a memory issue then a performance issue. CURLINFO_HEADER_OUT returns the full string as part of the response object. If you use bulk mode this can add up to quite a large object. CURLINFO_HEADER_OUT could also be set through the curl options.

The query time part should not have an affect and I think this is what is mostly needed.

Best would be to active the benchmark tests to see if there is a difference. They should work now again but are not enabled for travis.

im-denisenko: > They should work

I tried, but they don't.

Then I would like to remove this constant. One who want to get headers in response could pass `CURLINFO_HEADER_OUT` via curl options.

im-denisenko:
@ruflin Changelog updated, debug removed, bc break mentioned.

ruflin:
Thx. Nice one as complexity reduced. We could mention in the CHANGELOG more specific that the curl params now must be used.

PR #869, opened 2015-06-02
Remove ansible folder from release

pborreli:


ruflin:
@im-denisenko Probably it is best you have a look here as you know best on how you are using this. From my point of view it makes sense to remove it.

im-denisenko:
Heh, I already wanted to exclude ansible folder (but forgot). 
It was a reason for my question in https://github.com/ruflin/Elastica/pull/745#issuecomment-87463054.

@pborreli Could you please do one more commit and add to export-ignore *all* files except these?

```
lib/
CHANGELOG.md
composer.json
LICENSE.txt
README.md
```

ruflin:
@im-denisenko If we exclude all files, I'm not even sure if we need all the .md and .txt files. My assumption is that if a release package is created, only the files are need which will be accessed by a computer.

im-denisenko:
@ruflin In my opinion, release package (if we support this `export-ignore` feature) should include all files except those that useful only for library maintainers. From this point of view, we should include only files I mentioned above.

I think it's quite subjective. @webdevsHub what do you think? Let's have quorum! :smile:

ruflin:
@pborreli Your take on this?

pborreli:
@ruflin personally, I like when i can find tests and docs in vendor lib, it helps me when I'm working with no internet connection. I already saw that tests are excluded in elastica, and docs are in separate repository. when I was looking for example of code I saw that ansible folder was here and probably not used useless so I made this PR.

I don't have a global vision on this, but if you think we should exclude this `ansible` folder, let's merge this PR

ruflin:
I checked a few other php libraries and actually none of the libs are using it. I think we should go for the one of the two extremes: Either include everything (removing .gitattributes) or removing everything which is not needed to use it as a depndency, which is mainly the lib folder.

@pborreli For development you can use --prefer-source and it will fetch the full library. I think the problem here is that most engineers don't know about this command, which would also explain why most projects don't use .gitattributes. As the size of the library with and without does not differ too much I tend to go in the direction of always including everything as I can't find an example at the moment where removing it would be a big advantage.

im-denisenko:
@pborreli Merged, thanks!

PR #872, opened 2015-06-02
Pass term arguments to Query\Term constructor

iUtvikler:
The arguments where missing from the QueryBuilder


im-denisenko:
Could you please mention new parameter in docblock of this method?

im-denisenko:
Merged, thanks.

Issue #551, opened 2014-02-13
Add integration test to Elastica\Query\TermTest


ruflin:
At the moment the integration tests are missing

PR #873, opened 2015-06-03
Add functional has-parent test

ruflin:


test/lib/Elastica/Test/Query/HasParentTest.php
```diff
@@ -52,4 +56,53 @@ public function testSetScope()
...
+
+        $shopType = $index->getType('shop');
+        $productType = $index->getType('product');
+        $mapping = new Mapping();
+        $mapping->setParent('shop');
+        $productType->setMapping($mapping);
+
+
+        $shopType->addDocument(new Document('zurich', array('brand' => 'google')));
+        $shopType->addDocument(new Document('london', array('brand' => 'apple')));
```
im-denisenko: Use `addDocuments` to avoid unnecessary requests during tests.

test/lib/Elastica/Test/Query/HasParentTest.php
```diff
@@ -52,4 +56,53 @@ public function testSetScope()
...
+
+        $doc1 = new Document(1, array('device' => 'chromebook'));
+        $doc1->setParent('zurich');
+
+        $doc2 = new Document(2, array('device' => 'macmini'));
+        $doc2->setParent('london');
+
+
+        $productType->addDocument($doc1);
+        $productType->addDocument($doc2);
```
im-denisenko: Same thing.

ruflin: I actually used here addDocument on purpose in case some parent mapping goes wrong or is incorrect, it is easier to find where the exception was thrown.

test/lib/Elastica/Test/Query/HasParentTest.php
ruflin: Will change it, even though the difference is quite minimal.

im-denisenko:
lgtm, merged.

There are yet another issue about missing functional tests: #551, if you want one :)

PR #876, opened 2015-06-08
Remove gitattributes

im-denisenko:
As far as I see, existence of `export-ignore` feature in context of php packages distribution it's just a reason for lots of debates.

It was discussed in #869 and #745, and also in many places across internet, but for now it's not widely used, hence developers most likely will not know answer for "why part of the library is missing in my installation".

Also, most of packages I've seen while thinking about this issue follows the same scenario: "add export-ignore" -> "issues arrived" -> "remove export-ignore". 

I suggest to remove this file completely.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,0 @@
-/ansible export-ignore
-/test export-ignore
-.gitattributes export-ignore
-.gitignore export-ignore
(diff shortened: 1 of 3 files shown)


ruflin:
Agree. Merged.

PR #877, opened 2015-06-14
StrategyFactory should check predefined strategies first.

### LINKED ISSUE #721: class_exists() call in StrategyFactory
Class: ``Elastica\Connection\Strategy\StrategyFactory``
Line: 26

This call bubbles up to our auto-loader which throws an exception if class was not found (it has a full index of defined classes pre-loaded, so if it does not find something, it throws an exception). That's the first issue.

Second issue is that there is great chance that there is a class called ``Simple`` in the global scope, so this check is not solid. You can capture a class that is not for Elastica to be found, but used for something else.

I've set our auto-loader to ignore checks for ``Simple`` class, but that's super-dirty workaround and should not happen in the first place. Not sure which strategy discovery approach you should use, just pointing out that current one has some problems.

Thanks for an awesome library. It saves us a lot of time.

im-denisenko:
Closes #721, #730.

* Changed order of the `class_exists` methods.
* Changed exception message.
* Updated `StrategyFactory::create` to be more readable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to this project will be documented in this file based on the
 
 ### Improvements
 - `CallbackStrategy` now will accept any `callable` as callback, not only instance of `Closure`. [#871](https://github.com/ruflin/Elastica/pull/871)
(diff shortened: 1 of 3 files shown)


ruflin:
I had to rerun the build as PHP 7 and HHVM failed. But now everything is green.

PR #881, opened 2015-06-18
Update elasticsearch dependency to elasticsearch 1.6.0

ruflin:


ruflin:
To fix the snapshot problems, the paths need to be added to the config:
```
path.repo: ["/tmp/test_register", "/tmp/test_repository"]
```

In addition, the file systems above must be shared across the nodes: https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-snapshots.html

ruflin:
Bulk requests now throw exceptions for unrecognised parameters https://github.com/elastic/elasticsearch/pull/11331. _percolate seems to be unrecognised / unsupported. This param must be changed in the BulkTest::testSend

ruflin:
@im-denisenko FYI, in case you have some time to have a look at it.

im-denisenko:
Exceptions in Query\ImageTest seems to be related to https://github.com/elastic/elasticsearch/pull/10157/.
At least, [here](https://github.com/SibaTokyo/elasticsearch-image/blob/master/src/main/java/org/elasticsearch/index/mapper/image/ImageMapper.java#L258) second parameter or BytesStreamInput should be removed.
@Jmoati Perhaps you could be interested in that?

im-denisenko:
> _percolate seems to be unrecognised / unsupported

Also, methods setPercolate, getPercolate, hasPercolate in AbstractUpdateAction strongly outdated and should be also removed. Percolator doesn't work this way since ES 1.0. [Related commit] (https://github.com/elastic/elasticsearch/commit/b7c4adeea3ce0b8341a5ac66fd83cd676eaade64).

ruflin:
@im-denisenko Thx for looking into it. If the image plugin doesn't work the moment, I suggest to disable the tests and check if the plugin will be compatible again soonish. For the percolate things: I didn't check it in detail, but if these are not supported anymore, we should remove it.

ruflin:
@im-denisenko For removing the percolate functions I opened a separate issue: https://github.com/ruflin/Elastica/issues/883

ruflin:
@kzwang Do you plan any updates to the elasticsearch image plugin to make it compatible with ES 1.6.0?

ruflin:
@im-denisenko I was thinking about a new category for "plugin tests". Like this we know when something breaks if it is internal or plugin related. What do you think? The main problem is how "not" to mix again functional tests and unit tests. I would be mainly for the functional plugin tests ...

ruflin:
Seems like the newest version of the attachment mapper (or es) introduces a flaky test in our test suite. So far I'm not sure what causes it. It happens after indexing a docx file and then searching it not always works. Perhaps @rmuir can help?

ruflin:
I merged this pull request for the next release. Issues were opened for the existing issues (see above).

im-denisenko:
> I was thinking about a new category for "plugin tests". Like this we know when something breaks if it is internal or plugin related. What do you think?

@ruflin Add new category it's not a problem, but If we have at least one functional test per plugin, isn't this a test category you're talking about?

ruflin:
There could be also unit tests for the plugin. So what you say is that these would stay in the unit test category and only the functional would be added to the new category. Sounds good to me.

PR #891, opened 2015-07-08
Update .travis.yml

stevepeak:
Please do not merge. Just checking the output files. Thank you :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,11 +29,6 @@ before_script:
 script:
   - make phpunit
 
-after_script:
-  - cat /var/log/elasticsearch/*.log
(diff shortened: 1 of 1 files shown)


stevepeak:
Looks like the setting to build pr's is disabled. I'll activate it on my side.

ruflin:
It is quite strange that it didn't build, because the settings is actually enabled?!?

To create the coverage files locally is actually quite easy if you have boot2docker installed. Just run the command:

```
make run RUN="make phpunit"
```

The first time will take some time to fetch the docker images. This will produce all three coverage files in `build/coverage/`. I will try to get you some more data tomorrow.

PR #893, opened 2015-07-17
Elasticsearch 1.7.0

ruflin:


im-denisenko:
As I can see from changelogs, in both 1.6 and 1.7 releases public api doesn't changed.

Question is, should we add new class ([even empty](https://github.com/ruflin/Elastica/blob/master/lib/Elastica/QueryBuilder/Version/Version150.php)) to Elastica\QueryBuilder\Version namespace every ES release, or this is useless consistency?

// cc @webdevsHub as QueryBuilder author

webdevsHub:
I assume there will be further changes in future releases. So a `Latest` class could fill that gap until there is an API change.

TODO now:
* create `class Latest extends Version140`
* set default version of QueryBuilder to `Latest`
* deprecate `Version150`

TODO on next release with API change:
* create new version class
* modify super class of `Latest`

ruflin:
@webdevsHub @im-denisenko I like the suggestion with the Latest.

PR #897, opened 2015-07-27
Latest version class

webdevsHub:
description of this PR: https://github.com/ruflin/Elastica/pull/893

another option would be to just use `Version140` as default `QueryBuilder` version without `Latest` class but it could be misunderstood as "Elastica is outdated".

I will add changelog entry after discussion.

### DIFF
--- a/lib/Elastica/QueryBuilder/Version/Latest.php
+++ b/lib/Elastica/QueryBuilder/Version/Latest.php
@@ -1,0 +1,14 @@
+<?php
+namespace Elastica\QueryBuilder\Version;
+
+/**
+ * Latest elasticsearch DSL.
+ *
+ * @link https://www.elastic.co/guide/en/elasticsearch/reference/current/index.html
+ *
+ * @author Manuel Andreo Garcia <andreo.garcia@gmail.com>
+ */
+class Latest extends Version140
+{
+    // this class always points to the latest valid DSL version
+}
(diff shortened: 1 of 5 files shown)


ruflin:
I like the Latest as it always refers to the dependency in the README file. Hopefully the ES API does not change every time. This could even be mentioned in the doc block of Latest.

ruflin:
I would merged this pull request, if there are no objections.

webdevsHub:
What exactly do you want to mention in the comments?

ruflin:
That Latest references to the version mentioned in the README.md file. Like this it is always clear which one is meant as the README is also versioned and it is not confused with the latest version of elasticsearch.

webdevsHub:
I had a small log in the comment, but deleted it because it would require changing this file on every release. I could add it again

```php
/*
 * no changes since 1.4.0:
 * 	- 1.5.0 (March 23, 2015)
 *  - 1.6.0 (June 09, 2015)
 *  - 1.7.0 (July 16, 2015)
 */
```

lib/Elastica/QueryBuilder/Version/Latest.php
```diff
@@ -0,0 +1,14 @@
+<?php
+namespace Elastica\QueryBuilder\Version;
+
+/**
+ * Latest elasticsearch DSL.
+ *
+ * @link https://www.elastic.co/guide/en/elasticsearch/reference/current/index.html
```
ruflin: @webdevsHub Here I would add a comment: Latest always refer to the version mentioned in README.md

Like this it never has to be updated here :-)

ruflin:
See my comment in the code. I would leave out the comment that has to be changed every time.

webdevsHub:
comment added

ruflin:
Merged. Thx.

PR #898, opened 2015-07-27
php-cs-fixer PreIncrementFixer

webdevsHub:
some housekeeping:

php-cs-fixer recently added PreIncrementFixer in [1.9](https://github.com/FriendsOfPHP/PHP-CS-Fixer/releases/tag/v1.9) to `symfony` level. See also the [discussion](https://github.com/symfony/symfony/pull/14121).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file based on the
 
 ### Improvements
 - Update Elasticsearch dependency to 1.7.0 and update plugin dependencies
+- Update php-cs-fixer to 1.10 [#898](https://github.com/ruflin/Elastica/pull/898)
 
 ### Deprecated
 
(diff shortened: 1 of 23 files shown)


ruflin:
@webdevsHub If it is in 1.9 also the php-cs-fixer dependencies should be updated: https://github.com/ruflin/Elastica/blob/master/Dockerfile#L40

webdevsHub:
Okay, I am going to do this before merge

webdevsHub:
I modified `Dockerfile` to `RUN composer global require "fabpot/php-cs-fixer=1.10.*"`
but `make run RUN="make lint"` do not touch the files.

I downloaded it manually and `php php-cs-fixer.phar fix --config-file=./.php_cs` (with 1.10) changes one file `lib/Elastica/Aggregation/ReverseNested.php`. I am not into docker, but seems like I have to rebuild it or something?

ruflin:
Yes, the image has to be rebuilt. The problem is, that by default it takes the remote image. As soon as we merge the pull request, it will build the new image. In the docker compose file a line is commented out. If you switch the comment, it will build it locally.

I suggest you push it here and we merge it. There is nothing major that can break ;-)

webdevsHub:
that worked thx! Should be ready to merge

ruflin:
Merged. Thx.

ruflin:
@webdevsHub Here you can see that it just created the newest build for the latest push: https://registry.hub.docker.com/u/ruflin/elasticsearch-elastica/builds_history/205891/

PR #901, opened 2015-07-28
Makefile update

ruflin:
Use Docker as default in Makefile to simplify setup further. To not use Docker, the param DOCKER="" must be passed.


ruflin:
@webdevsHub @im-denisenko I suggest to have the docker environment as default for all make commands. This will simplify the setup even further and engineers will automatically use the docker setup. What do you think?

im-denisenko:
@ruflin I'm ok with it. Should we drop vagrant then to not confuse which one tool is actually used?

ruflin:
I currently still use vagrant to test if the ansible scripts work as expected as these are still used for travis. What I would like to do for travis is to switch to the container based architecture, which means removing sudo. Our main usage of sudo is here: https://github.com/ruflin/Elastica/blob/master/ansible/provision.sh#L5 So the question is if this could be replaced with the way travis installs the dependencies: http://docs.travis-ci.com/user/apt/ If we get this working, we don't the vagrant part anymore.

The best case would be if we could directly use our docker-compose setup. I still hope one of the CI/CD platforms is going to support this soon (@travis-ci @codeship ... ;-) ).

Makefile
```diff
@@ -1,103 +1,113 @@
...
+	phploc --log-csv ./build/logs/phploc.csv ${SOURCE}
+
+# Handling virtual environment
+
+build:
+	docker-compose build
+
+setup: build
+	docker-compose scale elasticsearch=3
+	docker-compose run elasticsearch chmod -R 777 /mount/
```
ruflin: @im-denisenko I'm really unhappy about this line. Not the 777 but that I have to change the access rights after the images are running. What I need is a shared data container where all elasticsearch instances have write access. I tried to do it directly in the data image https://github.com/ruflin/Elastica/pull/901/files#diff-6ddc070b659c249d339814ba724740edR10 or also in the elasticsearch https://github.com/ruflin/Elastica/pull/901/files#diff-66fcadff5e0bc20f3becadf59df62f5bR28 But only the setup part works. I'm quite sure there is just a small command that can be used to get it working. Do you have any experience here?

ruflin:
@im-denisenko I further worked on this. It is finally working also on remote machines with the Snapshots. I tried it on Google Cloud Engine. It still needs some cleanup and documentation needs to be added. If you have some time, have a first look at it to have a second thought on it.

ruflin:
@im-denisenko I merged it myself as this does not change any of the library functionality itself. I will add more documentation later.

Makefile
```diff
@@ -1,107 +1,113 @@
 #/bin/bash
 
-BASEDIR = $(shell pwd)
-SOURCE = "${BASEDIR}/lib"
-IMAGE = "elastica"
+.PHONY: update clean build setup start stop destroy run checkstyle checkstyle-ci code-browser cpd messdetector messdetector-ci dependencies phpunit test tests doc lint syntax-check loc phploc gource
```
im-denisenko: A lot of targets, some of them just copypasted from ant config file that was deleted far ago :) 
Which of them is *actually* used?

im-denisenko:
@ruflin Sorry I didn't respond in time. I left question in the code.

Makefile
ruflin: I actually planned to clean up the .PHONY part by putting it on top of every command. Like this no "old" commands should appear here. In general I tried to reduce the number of targets we have in the makefile, but there is still quite a list :-)

im-denisenko:
> Our main usage of sudo is [here](https://github.com/ruflin/Elastica/blob/master/ansible/provision.sh#L5)

Nope, it's actually [here](https://github.com/ruflin/Elastica/blob/master/ansible/provision.sh#L30). Ansible escalates it's privileges to root when runned, we don't see it because of passwordless sudo in  travis.

I recently did a lot of ansible-related work on my job, now I know this tool even better, and I'm think I know that must be changed in order to use containers. Give me a couple of days...

Makefile
ruflin: @im-denisenko Here is the pull request for the Makefile. Have a look and merge it if you think that helps: https://github.com/ruflin/Elastica/pull/917

ruflin:
@im-denisenko About the sudo and containers: Let me know when you have an update. I know quite a bit about the containers but I'm not at all an expert with ansible :-)

PR #905, opened 2015-08-05
Implemented index template support

torinaki:
Initial implemtation of elastica index templates support:
https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-templates.html

Required for FriendsOfSymfony/FOSElasticaBundle#916

### DIFF
--- a/lib/Elastica/IndexTemplate.php
+++ b/lib/Elastica/IndexTemplate.php
@@ -1,0 +1,118 @@
+<?php
+namespace Elastica;
+
+use Elastica\Exception\InvalidException;
+
+/**
+ * Elastica index template object.
+ *
+ * @author Dmitry Balabka <dmitry.balabka@gmail.com>
+ */
+class IndexTemplate
+{
+    /**
+     * Index template name.
+     *
+     * @var string Index pattern
+     */
+    protected $name = '';
+
+    /**
+     * Client object.
+     *
+     * @var \Elastica\Client Client object
+     */
+    protected $client = null;
+
+    /**
+     * Creates a new index template object.
+     *
+     * @param \Elastica\Client $client Client object
+     * @param string           $name   Index template name
+     *
+     * @throws \Elastica\Exception\InvalidException
+     */
+    public function __construct(Client $client, $name)
+    {
+        $this->client = $client;
+
+        if (!is_scalar($name)) {
+            throw new InvalidException('Index template should be a scalar type');
+        }
+        $this->name = (string) $name;
+    }
+
+    /**
+     * Deletes the index template.
+     *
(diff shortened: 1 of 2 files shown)


lib/Elastica/IndexTemplate.php
```diff
@@ -0,0 +1,118 @@
...
+ * @author Dmitry Balabka <dmitry.balabka@gmail.com>
+ */
+class IndexTemplate
+{
+    /**
+     * Index template name.
+     *
+     * @var string Index pattern
+     */
+    protected $name = '';
```
ruflin: I know that symfony has the naming standard without the underline, but as the Elastica project everywhere use $_* for protected variables, please adjust it.

lib/Elastica/IndexTemplate.php
```diff
@@ -0,0 +1,118 @@
+<?php
+namespace Elastica;
+
+use Elastica\Exception\InvalidException;
+
+/**
+ * Elastica index template object.
+ *
```
ruflin: I suggest to add here the link to the es documentation: https://www.elastic.co/guide/en/elasticsearch/reference/1.6/indices-templates.html

ruflin:
In general this looks good to me. It would be good to have 1-2 more functional tests. This would also discover in case the API changes in the future.

ruflin:
@torinaki I'm happy to merge the pull request if you adjust the coding styles above so https://github.com/FriendsOfSymfony/FOSElasticaBundle/issues/916 can be pushed forward.

mablae:
Another awsome feature! I will use it. Thanks.

ruflin:
@torinaki I made the adjustments here and merged it: https://github.com/ruflin/Elastica/commit/f64e8ae45de2467d99b223f165e6780d0cd8bb36 It would be nice if you could add some more integration tests later.

lib/Elastica/IndexTemplate.php
torinaki: I just made it like that because in contributing notes is mentioned that code must be in PSR-2 code style, but exactly PSR-2 says that:
> Property names SHOULD NOT be prefixed with a single underscore to indicate protected or private visibility.
https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#42-properties

torinaki:
@ruflin thank you for quick response. I'll prepare more functional tests with another PR soon

lib/Elastica/IndexTemplate.php
ruflin: Yep, we take the SHOULD part here ;-) I fixed it and pushed it to master. Thanks for the contribution.

ruflin:
:+1:

PR #910, opened 2015-08-07
typofix - https://github.com/vlajos/misspell_fixer

vlajos:


ruflin:
Merged, thx. We should get this probably as one of our build commands as typos are very common.

PR #916, opened 2015-08-13
lazy toArray

ewgRa:
This is PR for #783, convert object toArray when it is really needed. Comments, tests, review - welcome.

BC: 
1. Now toArray is lazy. This mean that if you set object to Query, Aggregation and so on, and later will change it - changes will be affected at toArray also.
For example such test:
```php
        $query = new Query();
        $aggregation = new \Elastica\Aggregation\Terms('text');

        $query->addAggregation($aggregation);

        $aggregation->setName('another text');

        $anotherQuery = new Query();
        $anotherQuery->addAggregation($aggregation);

        $this->assertEquals($query->toArray(), $anotherQuery->toArray());
```
with previous code will be not passed. With this PR - it is passed.

2. if your code work with params somehow (direct access to _params, getParams() methods and so on) - it can be broken.
For example, before:
```php
$script = Elastica\Script::create($script);
return $this->setParams($script->toArray());
```
Now:
```php
return $this->setParam('script', Elastica\Script::create($script));
```
Same as some named object:
before:
```php
$this->_aggs[$aggregation->getName()] = $aggregation->toArray();
```
after:
```php
$this->_aggs[] = $aggregation;
```

As you can see, aggregation name in toArray also become lazy.

### DIFF
--- a/lib/Elastica/Aggregation/Filter.php
+++ b/lib/Elastica/Aggregation/Filter.php
@@ -32,7 +32,7 @@ class Filter extends AbstractAggregation
      */
     public function setFilter(AbstractFilter $filter)
     {
-        return $this->setParam('filter', $filter->toArray());
+        return $this->setParam('filter', $filter);
     }
 
     /**
@@ -41,11 +41,11 @@ class Filter extends AbstractAggregation
     public function toArray()
     {
         $array = array(
-            'filter' => $this->getParam('filter'),
+            'filter' => $this->getParam('filter')->toArray(),
         );
 
         if ($this->_aggs) {
-            $array['aggs'] = $this->_aggs;
+            $array['aggs'] = $this->_convertArrayable($this->_aggs);
         }
 
         return $array;
--- a/lib/Elastica/Aggregation/Filter.php
+++ b/lib/Elastica/Aggregation/Filter.php
@@ -1,6 +1,7 @@
 <?php
 namespace Elastica\Aggregation;
 
+use Elastica\Exception\InvalidException;
 use Elastica\Filter\AbstractFilter;
 
 /**
@@ -36,10 +37,16 @@ class Filter extends AbstractAggregation
     }
 
     /**
+     * @throws \Elastica\Exception\InvalidException If filter is not set
+     *
      * @return array
      */
     public function toArray()
     {
+        if (!$this->hasParam('filter')) {
+            throw new InvalidException('Filter is required');
+        }
+
         $array = array(
             'filter' => $this->getParam('filter')->toArray(),
         );
--- a/lib/Elastica/Aggregation/Filters.php
+++ b/lib/Elastica/Aggregation/Filters.php
@@ -23,9 +23,9 @@ class Filters extends AbstractAggregation
     public function addFilter(AbstractFilter $filter, $name = '')
     {
         if (empty($name)) {
-            $filterArray[] = $filter->toArray();
+            $filterArray[] = $filter;
         } else {
-            $filterArray[$name] = $filter->toArray();
+            $filterArray[$name] = $filter;
         }
 
         return $this->addParam('filters', $filterArray);
@@ -44,14 +44,14 @@ class Filters extends AbstractAggregation
             $key = key($filter);
 
             if (is_string($key)) {
-                $array['filters']['filters'][$key] = current($filter);
+                $array['filters']['filters'][$key] = current($filter)->toArray();
             } else {
-                $array['filters']['filters'][] = current($filter);
+                $array['filters']['filters'][] = current($filter)->toArray();
             }
         }
 
         if ($this->_aggs) {
-            $array['aggs'] = $this->_aggs;
+            $array['aggs'] = $this->_convertArrayable($this->_aggs);
         }
 
         return $array;
--- a/lib/Elastica/Aggregation/Filters.php
+++ b/lib/Elastica/Aggregation/Filters.php
@@ -22,10 +22,12 @@ class Filters extends AbstractAggregation
      */
     public function addFilter(AbstractFilter $filter, $name = '')
     {
-        if (empty($name)) {
-            $filterArray[] = $filter;
-        } else {
+        $filterArray = array();
+
+        if (is_string($name)) {
             $filterArray[$name] = $filter;
+        } else {
+            $filterArray[] = $filter;
         }
 
         return $this->addParam('filters', $filterArray);
--- a/lib/Elastica/Arrayable.php
+++ b/lib/Elastica/Arrayable.php
@@ -1,0 +1,18 @@
+<?php
+namespace Elastica;
+
+/**
+ * Interface for params.
+ *
+ *
+ * @author Evgeniy Sokolov <ewgraf@gmail.com>
+ */
+interface Arrayable
+{
+    /**
+     * Converts the object to an array.
+     *
+     * @return array Object as array
+     */
+    public function toArray();
+}
--- a/lib/Elastica/Nameable.php
+++ b/lib/Elastica/Nameable.php
@@ -1,0 +1,13 @@
+<?php
+namespace Elastica;
+
+/**
+ * Interface for params.
+ *
+ *
+ * @author Evgeniy Sokolov <ewgraf@gmail.com>
+ */
+interface Nameable
+{
+    public function getName();
+}
--- a/lib/Elastica/Suggest/AbstractSuggest.php
+++ b/lib/Elastica/Suggest/AbstractSuggest.php
@@ -1,13 +1,14 @@
 <?php
 namespace Elastica\Suggest;
 
-use Elastica\Nameable;
+use Elastica\Exception\InvalidException;
+use Elastica\NameableInterface;
 use Elastica\Param;
 
 /**
  * Class AbstractSuggestion.
  */
-abstract class AbstractSuggest extends Param implements Nameable
+abstract class AbstractSuggest extends Param implements NameableInterface
 {
     /**
      * @var string the name of this suggestion
@@ -25,7 +26,7 @@ abstract class AbstractSuggest extends Param implements Nameable
      */
     public function __construct($name, $field)
     {
-        $this->_name = $name;
+        $this->setName($name);
         $this->setField($field);
     }
 
@@ -71,6 +72,26 @@ abstract class AbstractSuggest extends Param implements Nameable
     public function setShardSize($size)
     {
         return $this->setParam('shard_size', $size);
+    }
+
+    /**
+     * Sets the name of the suggest. It is automatically set by
+     * the constructor.
+     *
+     * @param string $name The name of the facet.
+     *
+     * @throws \Elastica\Exception\InvalidException If name is empty
+     *
+     * @return $this
+     */
+    public function setName($name)
+    {
+        if (empty($name)) {
+            throw new InvalidException('Suggest name has to be set');
+        }
+        $this->_name = $name;
+
+        return $this;
     }
 
     /**
--- a/lib/Elastica/Suggest/AbstractSuggest.php
+++ b/lib/Elastica/Suggest/AbstractSuggest.php
@@ -1,12 +1,13 @@
 <?php
 namespace Elastica\Suggest;
 
+use Elastica\Nameable;
 use Elastica\Param;
 
 /**
  * Class AbstractSuggestion.
  */
-abstract class AbstractSuggest extends Param
+abstract class AbstractSuggest extends Param implements Nameable
 {
     /**
      * @var string the name of this suggestion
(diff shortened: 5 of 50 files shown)


lib/Elastica/Nameable.php
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * Interface for params.
+ *
+ *
+ * @author Evgeniy Sokolov <ewgraf@gmail.com>
+ */
+interface Nameable
+{
+    public function getName();
+}
```
im-denisenko: I think that for consistency, interface should include `setName` as well.

lib/Elastica/Suggest/AbstractSuggest.php
```diff
@@ -1,12 +1,13 @@
...
 namespace Elastica\Suggest;
 
+use Elastica\Nameable;
 use Elastica\Param;
 
 /**
  * Class AbstractSuggestion.
  */
-abstract class AbstractSuggest extends Param
+abstract class AbstractSuggest extends Param implements Nameable
```
im-denisenko: My commet above + `AbstractSuggest` should have setter for name regardless of new interface.

lib/Elastica/Aggregation/Filter.php
```diff
@@ -41,11 +41,11 @@ public function setFilter(AbstractFilter $filter)
     public function toArray()
     {
         $array = array(
-            'filter' => $this->getParam('filter'),
+            'filter' => $this->getParam('filter')->toArray(),
```
im-denisenko: `$filter` is optional in constructor, will be fatal error if it is not defined

im-denisenko: I'm wrong, `getParam` will throw `InvalidException`. Still, it's better to check that was returned.

lib/Elastica/Aggregation/Filters.php
```diff
@@ -23,9 +23,9 @@ class Filters extends AbstractAggregation
     public function addFilter(AbstractFilter $filter, $name = '')
     {
         if (empty($name)) {
-            $filterArray[] = $filter->toArray();
+            $filterArray[] = $filter;
```
im-denisenko: `$filterArray` was not initialized.

lib/Elastica/Aggregation/Filters.php
```diff
@@ -23,9 +23,9 @@ class Filters extends AbstractAggregation
     public function addFilter(AbstractFilter $filter, $name = '')
     {
         if (empty($name)) {
```
im-denisenko: Could be filter named as zero?
`empty('0') === true`

lib/Elastica/Arrayable.php
```diff
@@ -0,0 +1,18 @@
+<?php
+namespace Elastica;
+
+/**
+ * Interface for params.
+ *
+ *
+ * @author Evgeniy Sokolov <ewgraf@gmail.com>
+ */
+interface Arrayable
```
im-denisenko: Offtop: still don't sure about this name, in russian "аррябл" sounds quite silly :)

lib/Elastica/Nameable.php
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Interface for params.
+ *
+ *
+ * @author Evgeniy Sokolov <ewgraf@gmail.com>
+ */
+interface Nameable
+{
+    public function getName();
```
im-denisenko: Missing docblock.

im-denisenko:
Awesome!

I suspect something is broken, but all tests seems to be green, hence lgtm.

I left comments in the code, also could you please:
1) update changelog
2) add docblocks to all new methods
3) run php-cs-fixer
4) save result of _getBaseName into variable in toArray methods. It's sounds like microoptimization, but [here](https://github.com/ruflin/Elastica/pull/916/files#diff-235a2e26d1dfd979335082c5d219d001R164) this method was called 6 times, it's too much

lib/Elastica/Arrayable.php
ewgRa: I disagree. "Arrayable" is more used, for example https://github.com/search?l=php&p=1&q=Arrayable&ref=searchresults&type=Code&utf8=%E2%9C%93 and https://github.com/search?l=php&p=1&q=ToArrayInterface&ref=searchresults&type=Code&utf8=%E2%9C%93

Also Arrayable used in some frameworks, like Yii and seems something have Laravel API.

It is common practice to call such objects like "*able", that show to which one thing it can be transformed. Also there is already SearchableInterface. 

At least I will rename Nameable to NamebleInterface and Arrayable to ArrayableInterface for match naming convention.

For name it ToArrayInterface lets hear another opinions, because your one looks like a little bit personal.

im-denisenko: Ok, sorry, your arguments looks more reasonable :)

You right, it's subjective. I didn't seen much `*able` interfaces before and I'm just used to `Interface` suffix at the end of the name.

lib/Elastica/Aggregation/Filters.php
ewgRa: I sync it with toArray logic, replace empty to is_string

lib/Elastica/Nameable.php
ewgRa: Done

lib/Elastica/Suggest/AbstractSuggest.php
ewgRa: Done

lib/Elastica/Aggregation/Filter.php
ewgRa: Check added

lib/Elastica/Aggregation/Filters.php
ewgRa: done

ewgRa:
All done, if squash needed or something else, let me know.

Also does I need add some tests for increase code-coverage?

ruflin:
@im-denisenko @ewgRa Looks good to me.
@ewgRa Adding tests is almost always a good thing :-)

im-denisenko:
@ewgRa Thanks, merged. If you still want add more test for coverage, feel free to open another PR.

ruflin:
Looks like the next Elastica version will be 2.3.0 because of the BC changes. We should get this out of the door as soon as possible and mention the changes. @ewgRa did a good job in the pull request explaining it.

ewgRa:
Ok, thanks!

I will add some tests later.

ruflin:
@ewgRa @im-denisenko There as an issue with the anonymous filters: https://github.com/ruflin/Elastica/commit/5a08b5f66b76f2f91ecba6213a7a6b9c9225326d I think it is fixed now. The main issue is actually that even though some tests fail on travis, it still reports green. I have to look into this.

ewgRa:
https://github.com/ruflin/Elastica/commit/5a08b5f66b76f2f91ecba6213a7a6b9c9225326d must be not affect of or affected due to this PR, as I understand.

Green travis with fail tests that strange, yes. When you fix it and some test will be broken because of this PR - write it here. As I remember when I work on this PR all tests was passed on my local computer. I will check again when I will be at home.

ruflin:
@ewgRa It looks good now. I just checked the build and this pull request was the first time it was broken. But with the commit above it is fixed.

lib/Elastica/Aggregation/Filters.php
```diff
@@ -22,10 +22,12 @@ class Filters extends AbstractAggregation
      */
     public function addFilter(AbstractFilter $filter, $name = '')
     {
-        if (empty($name)) {
-            $filterArray[] = $filter->toArray();
+        $filterArray = array();
+
+        if (is_string($name)) {
```
stof: this looks broken as the default value of the empty string is still a string

ruflin: @stof Agree. There is something wrong here. Did you spot this because of the discussion or did you run already in some issues with it?

stof: I looked at this PR from the release changelog

ruflin: @stof Here is the fix that I applied for the current release: https://github.com/ruflin/Elastica/commit/5a08b5f66b76f2f91ecba6213a7a6b9c9225326d

stof:
There is another thing which might not possible anymore: cloning a Query object (or any other objects impacted here): given you keep objects internally, most changes on 1 query could affect the other one too if you have any method mutating an inner object (I don't remember the API of all these classes enough to know whether this happens anywhere), while it would have worked before as arrays where used.
Would it make sense to implement ``__clone`` in these classes to clone the inner objects too ?

lib/Elastica/Aggregation/Filters.php
ewgRa: @ruflin change that @stof notice I made to sync this check between toArray and addFilter, because they must be same. I think additionally to https://github.com/ruflin/Elastica/commit/5a08b5f66b76f2f91ecba6213a7a6b9c9225326d there is must be changed addFilter method in same way. What do you think? Or I was wrong when say that check must be same between this two methods.

This change came from https://github.com/ruflin/Elastica/pull/916#discussion-diff-37099377

ewgRa:
Nice notice. Looks like another projects clone all inner objects and this is a good practice and this make sense. Also this will be similar to how clone work before.

But, if it will be implemented via __clone, there is no possibility to clone just only object, without clone inner objects. And there is not a small chance that somebody need this feature.
And clone object with clone inner objects can be implemented via deepClone method, or similar.

I look to symfony code for example, and there is in many places __clone cause clone of inner objects. And how developer can just clone only object if needed?
Is there any good discussion about this?
 
If implement this, can it be considered as bugfix to 2.3.0? Because it is kind of broke BC :(

ewgRa:
@ruflin @stof  I think that object that can be set via setters can't be considered as "inner" objects and must be not cloned. Good example is symfony Request::duplicate and Request::__clone.
What is @stof meant - this is duplicate on my mind.

If for example Param::$_params will be an SplObjectStorage instance - than we must clone it, because for example $cloned->removeParam call must haven't affect on $original.

But because it is array now, we don't need to clone objects inside it.

So. I +1 for duplicate() when it will be requested by someone, but -1 for __clone() as duplicate().

lib/Elastica/Aggregation/Filters.php
ruflin: @ewgRa Not sure what you meant with your second last sentence. Please open a pull request with some tests if possible for the suggested changes, so we can directly discuss it based on code.

ruflin:
@stof Thanks for bringing up this discussion.

A few notes/thoughts here:

* The good news is that we marked this change as BC break. The bad news is that there are very few people like @stof that read code changes and on top of it even realise the consequences. The part I'm most worried about about is if it will have any consequences to the usage of the FOSElasticaBundle. @stof @XWB can probably help here.
* We should try to actively inform people about this change. I suggest to create more "explicit" release notes explaining the consequences. @ewgRa Can you help here?
* For an implementation of __clone I suggest we wait until a specific problem with a use case pops up. As always, software is used in many ways and best is to implement by solving a problem. Then we can also check if we reached the goal.
* Changes on one query should not change an other "independent" query as we don't work with singletons. Of course as soon the same object is passed to multiple queries, changes will have affect on all queries, which I think is also partially intented. The question is if this the affect and consequences are understood. Lets see if the topic "cloning" will pop up here.

lib/Elastica/Aggregation/Filters.php
ewgRa: Ok, I will write tests on this case and we will discuss.

ruflin: :+1:

ewgRa:
> Can you help here?
Ok, I will try to write something more verbose with examples and explanations.
But from 23 Sep I'm on vacation and will return 5 Oct. Don't know if I can finish it before vacation. If not - then only after 5 Oct. If it is urgent - better to ask someone else with help. Also I'm not a native English writer :)

stof:
> The good news is that we marked this change as BC break

@ruflin respecting semver, this means the release should have been numbered 3.0 btw.

ruflin:
@stof Yep. If we are strict with semver ofr every BC break we must increase the full version. I normally sneak smaller BC breaks also into the MINOR releases. I think we should keep 3.0 for the changes to elasticsearch 2.0.

ruflin:
@ewgRa Don't worry, most of us are not native english writers. Let me know in case you have a draft and I can take a look at it. Otherwise just drop me a line and will check if I can do it.

ewgRa:
@ruflin 
@im-denisenko 

Article about broken BC: https://gist.github.com/ewgRa/217bb9226943309b1974

Please review, how it is looks like, is this exactly what needed, or not, and so on. Feel free to change, add you as co-author, publish and so on. I have vacation from 23 Sep and there is not so much time left to complete future requested changes in article. Do it yourself since my vacation will be started, or wait till I return from vacation.

ruflin:
@ewgRa That is exactly what we needed. That is where we can point people to in case questions pop up. I revised the article and published it here on Elastica.io: http://elastica.io/2015/09/22/elastica-2-dot-3-0-broken-backward-compatibility-explanation/

It is linked in the release notes so it is easy to find. @ewgRa Thanks a lot and enjoy your vacations.

PR #917, opened 2015-08-16
Improve Makefile

ruflin:
.PHONY as part of every command. This keeps the build file cleaner


im-denisenko:
Merged.

Looks quite complicated, but I guess gnu make lack better syntax for pure targets.

ruflin:
The "shorter" version would be to leave out .PHONY for most commands where no file / folder exists, which is the case for about 95%. This is more a failure safe option.

PR #920, opened 2015-08-23
add Script Id feature

nicolassing:
I've also refactorize all Scripting type


lib/Elastica/AbstractScript.php
```diff
@@ -1,21 +1,36 @@
...
+    protected $script;
+
+    protected $lang;
+
     /**
      * @param array|null $params
      * @param string     $id
      */
-    public function __construct(array $params = null, $id = null)
+    public function __construct($script, array $params = null, $lang = null, $id = null)
```
ruflin: If I understand this right, this is only a BC break for people that extended from AbstractScript. For all the inherited classes there is no difference that is not part of the library?

lib/Elastica/Client.php
```diff
@@ -716,4 +716,27 @@ public function setLogger(LoggerInterface $logger)
...
+    /**
+     * @param Script $script
+     * @param $lang
+     * @param $id
+     *
+     * @return Response
+     */
+    public function addIndexedScript(Script $script, $lang, $id)
+    {
+        return $this->request(sprintf('_scripts/%s/%s', $lang, $id), Request::POST, $script->toArray());
```
ruflin: Can you use string concatenation here? `'_script/' . $lang . '/' . $id)` ? The main reason is that it is used like this across the library.

lib/Elastica/Client.php
```diff
@@ -716,4 +716,27 @@ public function setLogger(LoggerInterface $logger)
...
+
+    /**
+     * @param $lang
+     * @param $id
+     *
+     * @return Response
+     */
+    public function getIndexedScript($lang, $id)
+    {
+        return $this->request(sprintf('_scripts/%s/%s', $lang, $id), Request::GET);
```
ruflin: See above.

test/lib/Elastica/Test/ScriptFileTest.php
```diff
@@ -53,7 +53,7 @@ public function testConstructor()
         $expected = array(
             'script_file' => $value,
         );
-        $this->assertEquals($value, $scriptFile->getScriptFile());
```
ruflin: As getScriptFile is not released yet I suggest we don't call this change a BC change. See https://github.com/ruflin/Elastica/blob/master/CHANGELOG.md Or should we create a proxy method getScriptFile in the scriptFile object?

ruflin:
The changes look good. See my comments above. Can you also update the CHANGELOG.md file? 

I like that it seems to simplify the script part once again :+1:

ruflin:
@nicolassing Any updates here? Would love to get this into master soon.

Tobion:
Solved by #1242

PR #924, opened 2015-09-02
Allow PHP 7.0 failures

XWB:
* We should aim at 100% PHP 7 compatibility
* PHP 5.3 works anyway

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,13 +5,12 @@ php:
   - 5.5
   - 5.4
   - 5.3
-  - 7
+  - 7.0
   - hhvm
 
 matrix:
   allow_failures:
     - php: 5.3
-    - php: 7
     - php: hhvm
   fast_finish: true
 


XWB:
Travis passes, not sure why `codecov/project` fails.

ruflin:
@XWB Codecov fails because it is below 60%. They have a bug in the PHP coverage reporting, so it is not consistent. 

The reason I would like to keep PHP 5.3 in allow failures is to make it obvious, that yes it still works but we could get rid of it at any time. There are already some hacks for PHP 5.3 that we need to cleanup earlier or later.

XWB:
@ruflin :+1: I restored the PHP 5.3 entry.

ruflin:
The title is a little bit wrong as it is more about "Not allow" PHP 7 failures ;-)

ruflin:
Merged anyways :-)

Issue #515, opened 2013-12-10
http compression


priyolahiri:
how do I make elastica use http compression after switching on http compression in elasticsearch?

PR #927, opened 2015-09-10
Enable the support of "http.compression: true" in elasticsearch

yvann:
Hi !

After some tests and the reading of https://github.com/ruflin/Elastica/issues/515
I figured out that compress the request and the response can be a big performance improvement !
I don't know exactly where you want to put the configuration to enable this compression but here are the lines that works.

Hope it will help you !

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ All notable changes to this project will be documented in this file based on the
 - Add Script File feature #902 #914
 
 ### Improvements
+- Support the http.compression in the Http transport adapter #515
 
 ### Deprecated
 
--- a/lib/Elastica/Client.php
+++ b/lib/Elastica/Client.php
@@ -35,6 +35,7 @@ class Client
         'roundRobin' => false,
         'log' => false,
         'retryOnConflict' => 0,
+        'compression' => false,
     );
 
     /**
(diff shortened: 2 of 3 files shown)


killerwolf:
+1

ruflin:
@yvann Would you suggest to have it enabled as default? Best would to have config option for the connection which is probably called "compression". Check this line here: https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Client.php#L34

ruflin:
BTW: Can you also update the CHANGELOG.md?

yvann:
Let me do that, I come back in a minute :)

ruflin:
Thx, merged.

yvann:
Do you consider a version 2.2.2 with this merge ? Thanks

ruflin:
Why not :-) But it would be 2.3.0 because of https://github.com/ruflin/Elastica/pull/916

lenybernard:
Thanks, it looks good !
However, this is breaking things when a connection is given, it doesn't call the getConfig and then the compression param is not set correctly.
For the symfony guys who are using https://github.com/FriendsOfSymfony/FOSElasticaBundle, I made the PR https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/932 to set the compression param by default and avoid to crash.

2 notices, It should be interesting to up the version and it should be cool to have a readme to have more details about the request compression effects. 

Thank you !

ruflin:
@lenybernard Thanks for bringing this up.
@yvann Can you have a look at this. I also just realised that the $compress config is at the wrong place. It should be noted as part of the $connections array here: https://github.com/yvann/Elastica/blob/11ec208cec201c8297570c2cdfe83608f0c713fb/lib/Elastica/Client.php#L34 If you want to set it as a global option, it must be fetched from the client. I suggest to add some tests to verify this.

PR #934, opened 2015-09-16
test that clone doesn't clone inner objects

ewgRa:
Test, that cover current work when object cloned.

From this comment: https://github.com/ruflin/Elastica/pull/916#issuecomment-140550935

### DIFF
--- a/test/lib/Elastica/Test/QueryTest.php
+++ b/test/lib/Elastica/Test/QueryTest.php
@@ -272,6 +272,23 @@ class QueryTest extends BaseTest
     /**
      * @group unit
      */
+    public function testNotCloneInnerObjects()
+    {
+        $query = new Query();
+        $termQuery = new Term();
+        $termQuery->setTerm('text', 'value');
+        $query->setQuery($termQuery);
+
+        $anotherQuery = clone $query;
+
+        $termQuery->setTerm('text', 'another value');
+
+        $this->assertEquals($query->toArray(), $anotherQuery->toArray());
+    }
+
+    /**
+     * @group unit
+     */
     public function testSetQueryToArrayChangeQuery()
     {
         $query = new Query();


test/lib/Elastica/Test/QueryTest.php
```diff
@@ -272,6 +272,23 @@ public function testSetQueryToArrayCast()
...
+        $query = new Query();
+        $termQuery = new Term();
+        $termQuery->setTerm('text', 'value');
+        $query->setQuery($termQuery);
+
+        $anotherQuery = clone $query;
+
+        $termQuery->setTerm('text', 'another value');
+
+        $this->assertEquals($query->toArray(), $anotherQuery->toArray());
```
ruflin: @ewgRa The question is if that is now the expected result or not :-)

ewgRa: On my mind it is expected now. I don't see any strong or even weak arguments against.

When somebody will need full clone with inner objects, better to make "duplicate" method.

ruflin:
@ewgRa I got a notification for a comment but can't see a comment here?!?

ewgRa:
@ruflin hmmm, maybe I accidentally delete it when try to edit... sorry.

I think it is expected. Nobody give any arguments against. At least we need this test to fixate current one state.

ruflin:
@ewgRa I merged it as you are right that it only tests the current state. 

When I see the code my initial gut feeling would tell me, that it would not be equal. That is why I think some people will stumble over this in the future. But lets wait until we have the first complain.

ewgRa:
@ruflin I have two cons on clone inner objects:

1. We lose ability to clone only object, without clone inner objects.
2. Users can extends from Elastica classes (Filter for example) and deny clone in customized objects. In this case they will have exceptions when they try to clone Query for example, when we try to clone inner objects. We don't know does inner object can be cloned, or not.

Current state haven't such cons, but we still have ability to implement clone with inner objects by introduce "duplicate" method.

ruflin:
@ewgRa My question for the first one: I'm not even sure what the use case of cloning would be if all contained objects are the same anyways?

ewgRa:
@ruflin For example clone and change only limit, offset, add new additional term, there is can be such cases, I'm sure.

Whatever can happen will happen (c) Murphy

ruflin:
:+1:

Issue #933, opened 2015-09-16
Term filter and diacritics


giorrrgio:
I have a field with multiple mapping: 
```php
'model' => array('type' => 'multi_field', 'fields' => array(
    'model' => array('type' => 'string', 'index' => 'analyzed'),
    'original' => array('type' => 'string', 'index' => 'not_analyzed')
    )
)
```
and I need to filter on its original value with an exact match. Everything is okay until I have a value with diacritics:
```json
{"model": "4er Gran Coupé"}
```
This is my query:
```php
$elasticaQuery = new \Elastica\Query\Filtered();
$elasticaFilterBool = new \Elastica\Filter\Bool();
$term = new \Elastica\Filter\Term();
$term->setTerm('model.original', '4er Gran Coupé');
$elasticaFilterBool->addMust($term);
$elasticaQuery->setFilter($elasticaFilterBool);
```
This is the result query:
```json
{
    "term": {
        "model.model.original": "4er Gran Coup\u00e9"
    }
}
```
As you can see the diacritics are escaped, but this way the query won't match anything in the database. If I run the same query via CURL without the escaped diacritics I get my results.
Any clue?

ewgRa:
I think problem not in Elastica.

How you convert array to json?

try echo(json_encode($elasticaQuery->toArray(), JSON_UNESCAPED_UNICODE));

ewgRa:
Check JSON.php and #559

im-denisenko:
@giorrrgio Please specify php version and used transport.

giorrrgio:
@ewgRa I'm not doing array to json conversions, at least not directly. The query I reported was obtained from the Symfony profiler with FOSElasticaBundle.

@im-denisenko I'm using Elastica Http transport and PHP 5.6.4

If you need other details or a test case to reproduce the issue let me know!

ewgRa:
It is not about Elastica. it is about how FOSElasticaBundle show such queries:

https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/Resources/views/Collector/elastica.html.twig#L54

```
<code>{{ query.data|json_encode }}</code>
```

Better I think to create issue in FOSElasticaBundle

giorrrgio:
@ewgRa the point is: the query is not giving me expected results, and it's not related to the way the query itself is encoded and displayed by the FOSEB collector or the Symfony profiler.

ewgRa:
It is related.

[...]

PR #937, opened 2015-09-17
Diacritics Term Filter integration test added, see #933

giorrrgio:


ruflin:
@giorrrgio I would recommend to do an integration/functional test as the unit test will only check if Elastic itself works as expected but not if the communication with elasticsearch works. Here you can find some example on how to do the mapping: https://github.com/ruflin/Elastica/blob/master/test/lib/Elastica/Test/Type/MappingTest.php

PR #938, opened 2015-09-17
avoid environment dependecies in tests

ewgRa:
On my local computer not all plugins installed, proxy not set and etc. and tests failed, when it must be skipped.

This PR add skip test and so on where it is needed.

### DIFF
--- a/test/lib/Elastica/Test/Aggregation/DateHistogramTest.php
+++ b/test/lib/Elastica/Test/Aggregation/DateHistogramTest.php
@@ -69,6 +69,8 @@ class DateHistogramTest extends BaseAggregationTest
      */
     public function testSetOffsetWorks()
     {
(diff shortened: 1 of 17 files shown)


ruflin:
@im-denisenko I think this could also going to be quite useful for ES 2.0

ruflin:
@ewgRa The other option to run all tests locally without needing the setup is using the docker-compose setup. The only thing you need to do is run `make tests` and have docker installed.

ewgRa:
I will try later, but usually developers just go and run "phpunit". Better to have there passed or skipped tests for everyone, in this case more easy to ask another developers run tests.

ruflin:
@ewgRa The problem with just running phpunit is that a lot of things can be different on each machine and some issues can be hard to reproduce. If everyone is running make tests we all have the exact same environment.

PR #940, opened 2015-09-20
Update Multi Match to support AUTO fuzziness

JustinHook:


test/lib/Elastica/Test/Query/MultiMatchTest.php
```diff
@@ -83,6 +83,14 @@ public function testFuzzy()
         $resultSet = $this->_getResults($multiMatch);
 
         $this->assertEquals(0, $resultSet->count());
+
+        $multiMatch = new MultiMatch();
+        $multiMatch->setQuery('Tritsan'); // Misspell on purpose
+        $multiMatch->setFields(array('full_name', 'name'));
+        $multiMatch->setFuzziness('AUTO');
```
ruflin: @JustinHook I would suggest we make a "constant" out of this which can be used. What do you think?

ruflin:
I like the addition, see my input above.

lib/Elastica/Query/MultiMatch.php
```diff
@@ -144,13 +144,13 @@ public function setType($type)
     /**
      * Set fuzziness.
      *
-     * @param float $fuzziness
+     * @param mixed $fuzziness
```
ewgRa: int|string instead of mixed? Or instead of mix argument add method like autoFuzziness and left setFuzziness as is?

ruflin: I like the int|string (or is it float|string)? Can $fuzzniess be a float? I think setFuzzniess must allow all possibilites. If we have an additional autoFuzziness method, why not. As a side mark here: auto is the default, so I assume this call will be rarely used.

ruflin:
@JustinHook Any updates here?

JustinHook:
@ruflin I've added the `const` and updated the `@param`.

ruflin:
@JustinHook Looks good. Can you also update the CHANGELOG.md? Afterwards I think we can merge it.

JustinHook:
@ruflin done

ruflin:
Thanks. Waiting for the build to get green :-)

ruflin:
Merged. Thx.

PR #941, opened 2015-09-22
Add JSON_BIGINT_AS_STRING solution for #717

### LINKED ISSUE #717: I'm facing problem with json_decode()
Hi,

I'm facing problem with Notice: `json_decode(): integer overflow detected`.

The weird is that in my dev enviroment this work fine but when I put this in the stage enviroment with another hardware and strucuture (AWS) I'm get this error.

The full log is: 

```json
"query": {
                  "aggs": {
                     "max_price": {
                        "nested": {
                           "path": "avail.rooms"
                        },
                        "aggs": {
                           "max_price": {
                              "max": {
                                 "field": "avail.rooms.rate.priceavg"
                              }
                           }
                        }
                     },
                     "min_price": {
                        "nested": {
                           "path": "avail.rooms"
                        },
                        "aggs": {
                           "min_price": {
                              "min": {
                                 "field": "avail.rooms.rate.priceavg"
                              }
                           }
                        }
                     },
                     "suppliers_map": {
                        "terms": {
                           "field": "supplier_id",
                           "size": 1000
                        }
                     }
                  },
                  "query": {
                     "filtered": {
                        "filter": {
                           "bool": {
                              "must": [
                                 {
                                    "and": {
                                       "filters": [
                                          {
                                             "term": {
                                                "city_id": 4202
                                             }
                                          },
                                          {
                                             "term": {
                                                "avail.checkin": "2014-12-15"
                                             }
                                          },
                                          {
                                             "term": {
                                                "avail.los": 2
                                             }
                                          },
                                          {
                                             "term": {
                                                "active": true
                                             }
                                          }
                                       ]
                                    }
                                 },
                                 {
                                    "nested": {
                                       "path": "avail.rooms",
                                       "query": {
                                          "filtered": {
                                             "filter": {
                                                "and": {
                                                   "filters": [
                                                      {
                                                         "range": {
                                                            "avail.rooms.capacity": {
                                                               "gte": 1
                                                            }
                                                         }
                                                      },
                                                      {
                                                         "range": {
                                                            "avail.rooms.availmin": {
                                                               "gte": 1
                                                            }
                                                         }
                                                      },
                                                      {
                                                         "range": {
                                                            "avail.rooms.rate.priceavg": {
                                                               "from": "105",
                                                               "to": "2991"
                                                            }
                                                         }
                                                      }
                                                   ]
                                                }
                                             }
                                          }
                                       }
                                    }
                                 }
                              ]
                           }
                        }
                     }
                  },
                  "sort": {
                     "category_order": {
                        "order": "desc"
                     }
                  }
               },
               "error_message": "Notice: json_decode(): integer overflow detected in /home/ota/html/vendor/ruflin/elastica/lib/Elastica/JSON.php line 33",
               "error_code": "Notice: json_decode(): integer overflow detected in /home/ota/html/vendor/ruflin/elastica/lib/Elastica/JSON.php line 33"
            }
```json

You know what I solve this? 

Thank you!

### PR DESCRIPTION
**Pull request by oldskool**
Fixes #717

### DIFF
--- a/lib/Elastica/Response.php
+++ b/lib/Elastica/Response.php
@@ -56,6 +56,13 @@ class Response
     protected $_status = null;
 
     /**
+     * Whether or not to convert bigint results to string (see issue #717)
+     *
+     * @var bool
+     */
+    protected $_jsonBigintConversion = false;
+
+    /**
      * Construct.
      *
      * @param string|array $responseString Response string (json)
@@ -183,7 +190,11 @@ class Response
                 $this->_error = true;
             } else {
                 try {
-                    $response = JSON::parse($response);
+                    if ($this->getJsonBigintConversion()) {
+                        $response = JSON::parse($response, false, 512, JSON_BIGINT_AS_STRING);
+                    } else {
+                        $response = JSON::parse($response);
+                    }
                 } catch (JSONParseException $e) {
                     // leave response as is if parse fails
                 }
@@ -304,5 +315,25 @@ class Response
         }
 
         return $data['_scroll_id'];
+    }
+
+    /**
+     * Sets whether or not to apply bigint conversion on the JSON result.
+     *
+     * @param bool $jsonBigintConversion
+     */
+    public function setJsonBigintConversion($jsonBigintConversion)
+    {
+        $this->_jsonBigintConversion = $jsonBigintConversion;
+    }
+
+    /**
+     * Gets whether or not to apply bigint conversion on the JSON result.
+     *
+     * @return boolean
+     */
+    public function getJsonBigintConversion()
+    {
+        return $this->_jsonBigintConversion;
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ All notable changes to this project will be documented in this file based on the
 ### Bugfixes
 
 ### Added
+- Connection option to convert JSON bigint results to strings can now be set [#717](https://github.com/ruflin/Elastica/issues/717)
 
 ### Improvements
 
--- a/lib/Elastica/Client.php
+++ b/lib/Elastica/Client.php
@@ -19,6 +19,7 @@ class Client
      *
      * log: Set to true, to enable logging, set a string to log to a specific file
      * retryOnConflict: Use in \Elastica\Client::updateDocument
+     * bigintConversion: Set to true to enable the JSON bigint to string conversion option (see issue #717)
      *
      * @var array
      */
@@ -35,6 +36,7 @@ class Client
         'roundRobin' => false,
         'log' => false,
         'retryOnConflict' => 0,
+        'bigintConversion' => false,
     );
 
     /**
--- a/lib/Elastica/Transport/Guzzle.php
+++ b/lib/Elastica/Transport/Guzzle.php
@@ -103,6 +103,9 @@ class Guzzle extends AbstractTransport
 
         $response = new Response((string) $res->getBody(), $res->getStatusCode());
         $response->setQueryTime($end - $start);
+        if ($connection->hasConfig('bigintConversion')) {
+            $response->setJsonBigintConversion($connection->getConfig('bigintConversion'));
+        }
 
         $response->setTransferInfo(
             array(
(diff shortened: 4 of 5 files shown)

### CODE REVIEW

File: lib/Elastica/JSON.php [hunk 1]
```diff
@@ -28,6 +28,17 @@ public static function parse(/* inherit from json_decode */)
             $args[] = true;
         }
 
+        // add code to set use JSON_BIGINT_AS_STRING by default
+        if (!isset($args[2])) {
```
ruflin: We should mention here why this code is here by linking to the issue: https://github.com/ruflin/Elastica/issues/717

ruflin:
@oldskool See my comment in the issue concerning config. I would suggest instead of making it available by default we have a config variable. Can you also add a small test to confirm this works as expected?

ewgRa:
@oldskool Cover by some tests also would be nice

oldskool:
Working on that :) @ruflin What would be the best way to pass the config to the JSON class, everything is static/stand-alone so the JSON class has no knowledge of the client config. Any hints on how to pass that properly?

ruflin:
Good question. It will be used in Result object, but I don't want to have there all the Config options. A more nasty but simpler way could be to set in the client and the assign it to some global variables which are read in JSON.

Any better ideas?

ewgRa:
It is definitely must be not a default thing.
On my mind response can have info about parsingOptions and use it in JSON::parse call.
Transport, when create Response can take this parsingOptions from $params in exec method, and to params such options must come from connection options and/or can be directly set/unset for specific request.

oldskool:
@ruflin @ewgRa I updated the PR with my new idea, how about something like this?
I'm not sure how to properly test this though, probably in the ResponseTest I think?

ewgRa:
It is close to what I mention in my comment, but broke backward compatibility.

I think this can be done without broke BC.

My suggestion:

1. Do not add jsonBigintConversion to Response constructor and instead use setter.

2. Is JSON class changes needed? If you have jsonBigintConversion, you can just call JSON::parse(content, true, flag). There is no reason to move it inside JSON class.

oldskool:
@ewgRa Where did BC break? I set defaults for everything trying to prevent just that.

oldskool:
@ewgRa Also applied your feedback.

lib/Elastica/Response.php
```diff
@@ -183,7 +190,8 @@ public function getData()
                 $this->_error = true;
             } else {
                 try {
-                    $response = JSON::parse($response);
+                    $flag = $this->getJsonBigintConversion() ? JSON_BIGINT_AS_STRING : 0;
+                    $response = JSON::parse($response, false, 512, $flag);
```
ruflin: I would suggest the following change here:
```
if (bigint) {
   $response = JSON::parse($response, false, 512, JSON_...)
} else {
   $response = JSON::parse($response);
}
```
The main reason for this are the defaults. They are correct at the moment, but in case a PHP version changes these defaults, we would stay with the old ones. I think if we apply special flags, this is ok. But for the default we should hardcoded fewer flags.

ewgRa:
Thanks! Much simpler and much better on my mind.

Broken BC because JSON::parse become non-static.
And also for future I there is can be several Response realization, that can implement interface. Json specific argument in Response constructor hard to maintenance.

Check elasticsearch-php for example. They have several serializers, also as there is no guarantee that later elasticsearch will be not support xml, or yml, or another output format.

ruflin:
LGTM. See my comment above. Please also update the CHANGELOG.md file.

oldskool:
@ruflin Done. I'm still not sure where/how to test this change, any suggestions on that?

ruflin:
@oldskool You would have to trigger the exception case somehow. I will merge it anyways for the moment as it will not break anything as long as it is not enabled :-)

oldskool:
@ruflin Cool! Could you tag this fix in a new release so we can pull it into our projects with Composer? Thanks! :smile:

ruflin:
@oldskool As was actually planning that the next release is 3.0 with elasticsearch 2.0 support, but will check what I can do as quite a few other things were sneaking in in the meantime.

ruflin:
@oldskool I released now 3.0.0-beta1 with this change. Is it for you already possible to upgrade to the newest version?

oldskool:
@ruflin Unfortunately not yet. We make extensive use of the dropped `NestedFilter` in our project. So we'll need to be looking at changing that.

ruflin:
@oldskool :-( The problem for me is that it is quite an effort to keep multiple release branches and make sure all work as expected. That is why I took it directly into 3.0

PR #947, opened 2015-10-07
Added setMinDocCount and setExtendedBounds options

markeilander:
Changed the correct file this time ;-)

### DIFF
--- a/lib/Elastica/Aggregation/DateHistogram.php
+++ b/lib/Elastica/Aggregation/DateHistogram.php
@@ -127,4 +127,39 @@ class DateHistogram extends Histogram
     {
         return $this->setParam('format', $format);
     }
+    
+    /**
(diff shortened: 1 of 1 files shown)


ruflin:
Merged, thx.

CedCannes:
@markeilander thank you for your work , I just need to use " extended_bound " :+1: 
By cons it seems that the function setMinDocCount already existed in Histogram

https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Aggregation/Histogram.php#L55

There are two functions to do the same...

ruflin:
@CedCannes Good point. I forgot about this one. @markeilander Can you remove the duplicate or use it as proxy? Because I like the "short" version naming.

markeilander:
Made a new pull request: https://github.com/ruflin/Elastica/pull/950

Issue #639, opened 2014-06-20
field_value_factor support


dwenaus:
will field_value_factor be supported?

yehosef:
I think this should be reopened.  Other options of the function_score are supported natively: weight, script, random, decay.  It's not clear why the field_value_factor should not be supported natively.

I think it's fine to leave open for someone to make the changes and pull request - but in my mind this is still an unresolved issue.

ruflin:
@yehosef Agree. I reopened it.

ruflin:
Implemented through https://github.com/ruflin/Elastica/pull/953/files

dwenaus:
:D

PR #953, opened 2015-10-14
#639 +field_value_factor

ruFog:
The field_value_factor support is added: https://www.elastic.co/guide/en/elasticsearch/reference/1.6/query-dsl-function-score-query.html#function-field-value-factor


ruflin:
Merged, thx.

PR #954, opened 2015-10-15
Fix docblocks

mortenhauberg:


ruflin:
Very cool, thanks. Did you go through all of it manually or did you have a tool that helped?

mortenhauberg:
Both, actually :)
Went through it manually, used PHPStorm to check it in each file and did the changes manually.

PR #955, opened 2015-10-16
Change check to OR instead of AND

davidyell:
Fixes the check for when `$args` is an array. So that the check doesn't fail, preventing the use of array arguments.

```php
$args = [];
        var_dump((!is_array($args) || !($args instanceof AbstractFilter))); // true
        var_dump((!is_array($args) && !($args instanceof AbstractFilter))); // false
```


davidyell:
I think this is a failure of understanding on my part.

Issue #964, opened 2015-10-20
_timestamp enable support in mapping


savdeep:
I was trying to find out  use timestamp with intial index mapping which generally is done by put request to elasticsearch as "_timestamp" : { "enabled" : true , "store" : "yes"} 

Is there any way to do this in elastica?

ruflin:
For the Mapping, there is currently not proxy function for _timestamp. But using Mapping->setParam you can set all params by just passing an array to it: https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Type/Mapping.php

For a later reference I put here also the es docs link: https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-timestamp-field.html

It would be nice to have a proxy function for this in Elastica. So feel free to also open a pull request with it.

In case the above doesn't work as expected, let me know.

ruflin:
Closing as https://github.com/ruflin/Elastica/pull/965 was merged.

PR #965, opened 2015-10-20
 support of adding _timestamp in during mapping

savdeep:
Added proxy methods for _timestamp

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ All notable changes to this project will be documented in this file based on the
 - Support for field_value_factor #953
 - Added setMinDocCount and setExtendedBounds options #947
 - Avoid environment dependecies in tests #938
+- Added support for intialise mapping with _timestamp
 
 ### Improvements
 - Update elasticsearch dependency to elasticsearch 1.7.3 #957
--- a/lib/Elastica/Type.php
+++ b/lib/Elastica/Type.php
@@ -299,6 +299,22 @@ class Type implements SearchableInterface
     }
 
     /**
+     * Sets value type mapping for this type with timestamp .
+     *
+     * @param \Elastica\Type\Mapping|array $mapping Elastica\Type\MappingType object or property array with all mappings
+     *
+     * @return \Elastica\Response
+     */
+    public function setMappingWithTimeStamp($mapping,$timestamp)
+    {
+        $mapping = Mapping::create($mapping);
+        $mapping->setType($this);
+        if(isset($timestamp) && is_array($timestamp))
+            $mapping->setTimestamp($timestamp);
+        return $mapping->send();
+    }
+
+    /**
      * Returns current mapping for the given type.
      *
      * @return array Current mapping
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ All notable changes to this project will be documented in this file based on the
 - Support for field_value_factor #953
 - Added setMinDocCount and setExtendedBounds options #947
 - Avoid environment dependecies in tests #938
+- Added support for intialise mapping with _timestamp
```
ewgRa: can you add reference to #964?

lib/Elastica/Type.php
```diff
@@ -299,6 +299,22 @@ public function setMapping($mapping)
     }
 
     /**
+     * Sets value type mapping for this type with timestamp .
+     *
+     * @param \Elastica\Type\Mapping|array $mapping Elastica\Type\MappingType object or property array with all mappings
+     *
+     * @return \Elastica\Response
+     */
+    public function setMappingWithTimeStamp($mapping,$timestamp)
```
ewgRa: I'm not sure is this method needed? Mapping know about timestamp, Type know only about mapping, but not mapping details.

lib/Elastica/Type/Mapping.php
```diff
@@ -294,4 +294,40 @@ public static function create($mapping)
...
+    }
+
+    /**
+     * Enables _timestamp for all documents in this type.
+     *
+     * @param bool $enabled OPTIONAL (default = true)
+     *
+     * @return $this
+     */
+    public function enableTimestamp($enabled = true)
```
ewgRa: not sure is this needed?

lib/Elastica/Type/Mapping.php
```diff
@@ -294,4 +294,40 @@ public static function create($mapping)
...
+    }
+
+    /**
+     * Enables and stores _timestamp for all documents in this type.
+     *
+     * @param string $store OPTIONAL (default = "yes")
+     *
+     * @return $this
+     */
+    public function storeTimestamp($store = "yes")
```
ewgRa: not sure is this needed?

ewgRa:
Would be nice if you add some tests

CHANGELOG.md
savdeep: Of course

lib/Elastica/Type.php
savdeep: Yes, type should know about timestamp. But in type class there isn't any way of referencing mapping directly (on members). In setMapping method send is called immediately after making mapping, so i needed to set timestamp param before that. There is no room to update type w.r.t  timestamp.

lib/Elastica/Type/Mapping.php
savdeep: setTimestamp is needed, but rest i added for other persons perusal

savdeep:
Tests, absolutely but after verifying above changes.

lib/Elastica/Type.php
ruflin: Instead of introducing a new function, I would suggest to add a param to setMapping

ewgRa: Disagree. _timestamp it is a mapping this. There is no reason to introduce it in Type. And what about another magic mapping fields like _size, _index... etc.? Also add as param?

savdeep:
I will do some changes and get back to you

PR #966, opened 2015-10-20
Added support for intialise mapping with _timestamp, _ttl and other such metas

savdeep:


ruflin:
@savdeep Thanks. Looks good to me. Do you plan to add some tests?

I think a valid point you brought up in the other pull request is that it is a little bit unexpected that setMapping directly sends the mapping. But lets keep that for the moment.

ruflin:
For reference, here is the discussion in the old PR: https://github.com/ruflin/Elastica/pull/965
@ewgRa FYI

savdeep:
Yup, i'll be adding them. @ewgRa whats your take on this pull request?

ewgRa:
I still can't understand general idea why it is needed and who deny developer to configure mapping before call Type::setMapping. I don't see any case (maybe you give example?) and any idea why one thing must be done with two ways, and second way (already) have magic inside and your PR add more magic there.

But if @ruflin ok with this, than ok.

I just have small cosmetic notices, will point it in lines review.

lib/Elastica/Type.php
```diff
@@ -287,17 +287,22 @@ public function getName()
      * Sets value type mapping for this type.
      *
      * @param \Elastica\Type\Mapping|array $mapping Elastica\Type\MappingType object or property array with all mappings
+     * @param argArray = array('_timestamp' => array('enabled' => true, 'store' => 'yes'), '_ttl' => array('enabled' => true));
      *
      * @return \Elastica\Response
      */
-    public function setMapping($mapping)
+    public function setMapping($mapping,$argArray = NULL)
```
ewgRa: can you add space after comma?

$mapping, $argArray = NULL

lib/Elastica/Type/Mapping.php
```diff
@@ -294,4 +294,5 @@ public static function create($mapping)
 
         return $mappingObject;
     }
+
```
ewgRa: can you revert such changes that haven't any affects? also in previous file.

to make diff more clear

ruflin:
After having a second look, I think I now understand the concern of @ewgRa . The function belongs in Mapping and not type so we always work with objects. @savdeep I'm sorry as you were right with your first implementation. I think the main concern there was if we need the enabled and store methods (@ewgRa)? 

@savdeep Can you update your first PR again, remove the code in Type and the two methods above? Sorry again for the extra rounds.

ewgRa:
> I'm sorry as you were right with your first implementation

First implementation, if we say about #965 also not right, on my mind this implementation better than that.
_timestamp  - it is special mapping field, like _size and so on. Why Type must know anything about how to configure mapping?

> I think the main concern there was if we need the enabled and store methods

Main concern is about why Type work with mapping configuration?

My use case (and for example how FOSElasticaBundle use it):
```
$mapping = new Mapping...
$mapping->set... / configure mapping
$type = new Type...
$type->setMapping($mapping);
```
And use case of this PR:
```
$mapping = new Mapping...
$mapping->set... / configure mapping, optional lines
$type = new Type...
$type->setMapping($mapping, {mappingConfigurationHere});
```

What is point to give Type extra functionality to work with Mapping configuration?


On my mind Type must be left as is there is no changes needed that related to _timestamp.

For mapping there is can be added: Mapping::setTimestamp($options)
Options is an array like (enable=> true, ...).

And a best way:
Create TimestampMappingField class that implement methods like setEnable, setStore, setParam and so on. With toArray method.

Add Mapping::setTimestamp(TimestampMappingField $field) where $field->toArray will be called.

And use case will be looks very nice and readable:

```
$timestamp = new TimestampMappingField();
$timestamp->setEnable(true);

$mapping = new Mapping...
$mapping->setTimestamp($timestamp);
$type = new Type...
$type->setMapping($mapping);
```

ruflin:
@ewgRa We are on the same page here. That is why I mentioned above ` remove the code in Type`. I would probably not directly got with the TimestampMappingField object as I think it is a little bit overkill, having ->setTimestamp should be enough.

@ewgRa Thanks for insisting on this one as otherwise I would not have spotted this.

ewgRa:
I would happy to have TimestampMappingField.

What happens when I want setTimestamp without such class? Right, I need open elasticsearch documentation, because I don't remember all configuration options. With TimestampMappingField I will have it in my IDE code completion.

There is some props with better refactoring and find places where I use specific functionality, also as more flexible work with deprecations and so on.

But of course introduce such classes will need more work with support it, but it is easy and I think Elastica community will handle it.

ruflin:
@ewgRa Yep, the main thing I worry about is maintenance. I'm currently facing the challenge with the changes to es 2.0 :-)

savdeep:
Thanks @ewgRa and @ruflin, creating a mapping objecct before and initiating it with setParam is the right option. Yes we can have that proxy functions but lets do that in 2.0. Cheers.

Issue #946, opened 2015-10-06
Migrate to Elasticsearch 2.0


ruflin:
Elasticsearch 2.0 bring several BC breaks. Adaptions must be made. Here is a checklist of open issues:

* [x] Status API was removed:
  * https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-status.html
  * https://www.elastic.co/guide/en/elasticsearch/reference/2.0/indices-status.html
* [x] Thrift support was removed
* [x] Facets were replaced by aggregations and removed -> Facets are removed from the code.
* [x] Memcached support was removed
* [x] Bulk UDP API has been removed
* [x] Delete by query was moved to a plugin?
* [x] Error handling got more extensive. Example:

Before:
```
{
  "error" : "IndexMissingException[[lkwer] missing]",
  "status" : 404
}
```
After
```
{
  "error" : {
    "root_cause" : [ {
      "type" : "index_not_found_exception",
      "reason" : "no such index",
      "resource.type" : "index_or_alias",
      "resource.id" : "334",
      "index" : "334"
    } ],
    "type" : "index_not_found_exception",
    "reason" : "no such index",
    "resource.type" : "index_or_alias",
    "resource.id" : "334",
    "index" : "334"
  },
  "status" : 404
}
```

yellow1912:
thank you. I'm closely following your progress here as well.

Lumbendil:
Going to follow this issue aswell :) Btw, you have _Thrift support was removed_ dupplicated.

ruflin:
@Lumbendil Thanks. That is what you get using the Union merge :-)

Lumbendil:
Yay, this got updated.

Is there any simple issue which could be tackled by someone who hasn't delved to much on the library, to contribute a bit?

ruflin:
@Lumbendil Good question. Probably the easiest to start is get a clone from the elasticsearch-2.0 branch locally, install docker and docker-compose and run `make tests`. This will output quite a few failing tests. Pick one that you think is easy to fix. 

Take for example the Fuzzy queries which are failing: https://travis-ci.org/ruflin/Elastica/jobs/87636268 Haven't looked into it by I assume the syntax changed somehow.

Lumbendil:
I already have docker-compose (Thanks for providing a test env with the lib!) I'll try to tackle some of them if I can get time for it :D

ruflin:
[...]

PR #968, opened 2015-10-27
Elasticsearch 2.0

ruflin:
https://github.com/ruflin/Elastica/issues/946


ruflin:
@ewgRa I now added the plugin in the commits above. Clean up your environment by running `docker-compose stop` and `docker-compose rm`. Now if you run again `make tests` you should have the new setup.

Some of the deleteByQuery tests now seem to work again, others return wrong results: https://travis-ci.org/ruflin/Elastica/jobs/90520618#L1660

Have a look and let me know.

ewgRa:
Thanks. Yes, I have a plugin installed. But seems it is not working well, maybe bugs, maybe I use it not in a right way.

```
curl -XGET 'http://172.17.0.14:9200/twitter/tweet/_search'
{"took":1,"timed_out":false,"_shards":{"total":2,"successful":2,"failed":0},"hits":{"total":1,"max_score":1.0,"hits":[{"_index":"twitter","_type":"tweet","_id":"1","_score":1.0,"_source":{
    "tweet" : {
        "user" : "kimchy",
        "post_date" : "2009-11-15T14:12:12",
        "message" : "trying out Elasticsearch"
    }
```

but 
```
curl -XDELETE 'http://172.17.0.14:9200/twitter/tweet/_query?q=user:kimchy'
{"took":0,"timed_out":false,"_indices":{"_all":{"found":0,"deleted":0,"missing":0,"failed":0}},"failures":[]}
```

Will continue on it.

What I have now. If plugin not installed, answer is
```
{"found":false,"_index":"twitter","_type":"tweet","_id":"_query","_version":1,"_shards":{"total":1,"successful":1,"failed":0}}
```
We can check if '_id' in answer is equals to '_query', than we can say that plugin not installed.

When plugin installed we can have such response if index is not exists
```
{"error":{"root_cause":[{"type":"index_not_found_exception","reason":"no such index","resource.type":"index_or_alias","resource.id":"twitter","index":"twitter"}],"type":"index_not_found_exception","reason":"no such index","resource.type":"index_or_alias","resource.id":"twitter","index":"twitter"},"status":404}
```

and

```
{"took":0,"timed_out":false,"_indices":{"_all":{"found":0,"deleted":0,"missing":0,"failed":0}},"failures":[]}
```
if index exists.

On my mind if ES deprecate delete by query, than it is must be exception or something like this. Now _query considered as _id and it is not so right. I will create issue in elasticsearch, maybe they fix it.

ruflin:
@ewgRa I would go step by step. Lets see if we can get the existing queries and tests running and then check if we can throw proper exceptions. I didn't check the delete-by-query plugin in detail, but I would expect your second line to work.

About the tests I would suggest to add the same flag as we do for other plugins. So tests are only run, when plugin is installed.

ewgRa:
About failed test with delete type. In ES2.0 there is no possible to delete Type anymore: https://www.elastic.co/guide/en/elasticsearch/reference/2.0/indices-delete-mapping.html

Quite strange for me... now maintenance of type becomes much more complex :(

On Elastica side I prefer wait when #985 will be closed and make a PR that throw DeprecatedException on this method.

ruflin:
@ewgRa #985 is now merged.

PR #972, opened 2015-11-08
Fixed GeoShapeProvided relation parameter position

legendik:


ruflin:
Can you update the Changelog?

legendik:
Sure, updated.

ruflin:
Thanks. Somehow the downloads for the build are currently broken / down. Will wait until it builds again. Is this a recent change or something that was always broken?

legendik:
ok. I think it was always broken according to the first commit of this file. It is broken only if you need to change the shape relation. It works fine with default values (relation type `intersects `)

ruflin:
Good to know. Merged. Thx.

legendik:
Thanks!

rarothedude:
Hi all, i think this should be reversed. I tested this and with the relation variable back to line 68 (same level as the shape variable) and the relation works

ruflin:
@rarothedude @legendik Could these difference be related to elasticsearch versions? Which versions are you using?

rarothedude:
{
  "status" : 200,
  "name" : "Brainchild",
  "cluster_name" : "elasticsearch",
  "version" : {
    "number" : "1.7.4",
    "build_hash" : "0d3159b9fc8bc8e367c5c40c09c2a57c0032b32e",
    "build_timestamp" : "2015-12-15T11:25:18Z",
    "build_snapshot" : false,
    "lucene_version" : "4.10.4"
  },
  "tagline" : "You Know, for Search"


Hmm this sounds quiet old, isn;t it ? :P

rarothedude:
According to these elastic docs it should be at the same level as shape:
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-geo-shape-query.html

ruflin:
@rarothedude Even though your elasticsearch version is pretty old and is not supported by Elastica 5.0, it seems you are right. The part that bugs me is that our tests do not seem to catch the error.

Could you open a PR with the change and also add a test that currently fails and afterwards works? That would be great.

PR #978, opened 2015-11-15
typo in getErrorMessage

ewgRa:
In original version there is kind of typo, always return empty. This PR fix it and also add some tests.


ruflin:
I'm actually thinking about introducing a more complex object for the Error as with 2.0 we can now get much more information out of it. It is good to have now the current version tested so any changes will be better documented.

PR #981, opened 2015-11-18
ES2.0 Fix MoreLikeThis

ewgRa:
Some notes:

1. MoreLikeThis::setLikeText is deprecated. I create setLike, but remove trim magic there. Any information why it is needed in setLikeText? In ES documentation there is no any notices about must be "like" trimmed, or not. So, at least if it is really needed, I prefer to move it at toArray to follow "what we get is what we set"

2. MoreLikeThis API was removed from ES 2.0 at all.

3. I increase setMaxQueryTerms in one test case. I don't see any another reason why this test case fail.

### DIFF
--- a/test/lib/Elastica/Test/Query/MoreLikeThisTest.php
+++ b/test/lib/Elastica/Test/Query/MoreLikeThisTest.php
@@ -2,7 +2,7 @@
 namespace Elastica\Test\Query;
 
 use Elastica\Document;
-use Elastica\Filter\Bool;
+use Elastica\Filter\BoolFilter;
 use Elastica\Filter\Term;
 use Elastica\Index;
 use Elastica\Query;
@@ -104,7 +104,7 @@ class MoreLikeThisTest extends BaseTest
 
         $query = new Query\Filtered($mltQuery);
         // Return just the visible similar
-        $filter = new Bool();
+        $filter = new BoolFilter();
         $filterTerm = new Term();
         $filterTerm->setTerm('visible', true);
         $filter->addMust($filterTerm);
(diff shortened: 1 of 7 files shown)


ewgRa:
Test fail seems not related.

test/lib/Elastica/Test/Query/MoreLikeThisTest.php
```diff
@@ -2,6 +2,8 @@
 namespace Elastica\Test\Query;
 
 use Elastica\Document;
+use Elastica\Filter\Bool;
```
ruflin: You have to use BoolFilter here to make it compatible with PHP 7: https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Filter/BoolFilter.php

test/lib/Elastica/Test/Query/MoreLikeThisTest.php
```diff
@@ -41,23 +43,78 @@ public function testSearch()
...
+        $mltQuery = new MoreLikeThis();
+
+        $mltQuery->setMinTermFrequency(1);
+        $mltQuery->setMinDocFrequency(1);
+
+        $mltQuery->setLike($doc);
+
+        $query = new Query\Filtered($mltQuery);
+        // Return just the visible similar
+        $filter = new Bool();
```
ruflin: https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Filter/BoolFilter.php as above.

ruflin:
@ewgRa The test failure is related to the Bool usage. I just realised we should probably mark https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Filter/Bool.php deprecated (and the other which are not compatible with PHP 7)

ewgRa:
Fixed. Totally forget about reserved words in PHP7.

One test seems will never end, but output show that it exit with code 0.

ruflin:
LGTM now. Yes, the build process checking is somehow broken. I still have to fix that.

Somehow the attachment mapper tests seem to fail now. Not sure why ...

PR #985, opened 2015-11-21
Introduce DeprecatedException and trigger_error in deprecated places

ewgRa:
This PR introduce DeprecatedException that prevent hidden unexpected results when used functionality, that not supported by ES and haven't BC layer.

Also this PR add trigger_error to several deprecated methods and announce removing deprecated code at Elastica 4.0

### DIFF
--- a/lib/Elastica/Aggregation/DateHistogram.php
+++ b/lib/Elastica/Aggregation/DateHistogram.php
@@ -1,5 +1,6 @@
 <?php
 namespace Elastica\Aggregation;
+use Elastica\Exception\DeprecatedException;
 
 /**
  * Class DateHistogram.
@@ -11,7 +12,7 @@ class DateHistogram extends Histogram
     /**
      * Set pre-rounding based on interval.
      *
-     * @deprecated Option "pre_zone" is deprecated as of ES 1.5. Use "time_zone" instead
+     * @deprecated Option "pre_zone" is deprecated as of ES 1.5 and will be removed in Elastica 4.0. Use "time_zone" instead
      *
      * @param string $preZone
      *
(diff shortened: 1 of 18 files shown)


lib/Elastica/Aggregation/DateHistogram.php
```diff
@@ -11,7 +12,7 @@ class DateHistogram extends Histogram
     /**
      * Set pre-rounding based on interval.
      *
-     * @deprecated Option "pre_zone" is deprecated as of ES 1.5. Use "time_zone" instead
+     * @deprecated Option "pre_zone" is deprecated as of ES 1.5 and will be removed in Elastica 4.0. Use "time_zone" instead
```
ruflin: I would probably not name a "specific" version of Elastica, as I'm also willing to remove deprecated functions in minor version.

lib/Elastica/Filter/Bool.php
```diff
@@ -1,6 +1,8 @@
 <?php
 namespace Elastica\Filter;
 
+trigger_error('Use BoolFilter instead. From PHP7 bool is reserved word and this class will be removed in Elastica 4.0', E_USER_DEPRECATED);
```
ruflin: Nice one. The only concern I have is that I don't know what people are using to catch errors. But Lets keep it in.

ruflin:
@ewgRa I merged this one as I want to move forward here as fast as possible. We can still take the inputs in in a next PR.

ruflin:
@ewgRa BTW: Nice work. Thanks.

lib/Elastica/Filter/Bool.php
ewgRa: There is already some trigger_errors in another methods. Usually this is not a problem for production, just more log entries.

lib/Elastica/Aggregation/DateHistogram.php
ewgRa: Usually deprecated stuff removed at major versions, also it is a good to give developers some directions for future. But it is up to you. I will change it to "and will be removed in further Elastica releases" and make PR for this.

ruflin: My main problem is that I don't want let us "nail down" on a specific version number.

Issue #944, opened 2015-09-30
Implement elasticsearch-php as foundation of Elastica


ruflin:
Elastica is a high level elasticsearch client, elasticsearch-php is a low level elasticsearch client. What both have in common are the low level implementations of connector, transport etc. Instead of trying both clients up to date with the newest improvements on these levels I suggest to take the elasticsearch-php client as the foundation of Elastica for the connection part to elasticsearch. This will bring together efforts to implement things like Guzzle and connection strategies.

As an approach for implementation I suggest to extend \Elastica\Client by \Elasticsearch\Client and make the \Elastic\Client use connection and transport form \Elasticsearch\Client. If possible backward compatibility for all request should not break.

The above will in addition allow every user of Elastica to use all low level functionality of the extended client in addition.

xkidro:
+1 for this, I currently use the elasticsearch-php sdk but found that I need to implement high level functions myself :(

ruflin:
@xkidro Good to hear. First I was thinking about introducing this with the es 2.0, but there are still quite a few open changes, so perhaps I will even prioritise this one higher. If you have some to help with this migration, that would be cool.

CrAzE124:
+1 definitely, I haven't been on this repo for a while, and was just wondering whether this was possible. How can I help on this?

ruflin:
@CrAzE124 My plan to implement this was as following, but because of 2.0 I didn't have the time yet to try it out.

* Elastica\Client should extend php-elasticsearch::Client
* At first, the only method that uses php-elasticsearch::Client is Elastica\Client::request(...). 

The above would mean, php-elasticsearch has to be added to the composer file and the request method has to be adjusted so that it calls the php-elasticsearch client. At first, it doesn't even have to extend Elastica\Client.

The best help would be to try this out an open a pull request with these changes so we can discuss it on actual code.

merk:
@ruflin I've been hitting issues with FOSEB which aren't worth fixing and prompt me to start over. Where are you at with this change? Is it still on the radar?

I dont really want to start from scratch on a dependency of Elastica if Elastica is going to move to become a value add on top of the official client, I'd prefer to support both from the get go.

ruflin:
@merk This is definitively still on the radar. Unfortunately it seems like @CrAzE124 and @ewgRa both didn't find the time yet to progress on this one. Let me see if I find a way to push this change in faster.

[...]

PR #989, opened 2015-11-23
Review for integration of official PHP library 

CrAzE124:
Added "elasticsearch/elasticsearch" to the composer.json for integration with the official PHP client;
Updated the current Elastica\Client::request to use the transport from the Elasticsearch\Client member variable;
Added a new member to the Elastica\Client class to house an instance of Elasticsearch\Client;
TODO - Do we want to stick with Aggregation over Inheritance?
Added a factory method to get the appropriate type of ConnectionException - WIP;

This is all a WIP PR. 

References #944

### DIFF
--- a/lib/Elastica/Exception/Factory/ConnectionExceptionFactory.php
+++ b/lib/Elastica/Exception/Factory/ConnectionExceptionFactory.php
@@ -1,0 +1,36 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: thomas
+ * Date: 2015/11/23
+ * Time: 10:53 PM
+ */
+
+namespace Elastica\Exception\Factory;
+
+
+use Elastica\Exception\Connection\GuzzleException;
+use Elastica\Exception\ConnectionException;
+use Elastica\Transport\AbstractTransport;
+use GuzzleHttp\Exception\TransferException;
+
+/**
+ * TODO - Determine if this is actually going to work (and if it's even proper OOP design???)
+ * The idea here is to pass in the Transport type,
+ * and from that we should generate the appropriate
+ * ConnectionException
+ *
+ * Class ConnectionExceptionFactory
+ * @package Elastica\Exception\Factory
+ */
+class ConnectionExceptionFactory
+{
+    /**
+     * @param AbstractTransport $transport
+     * @return ConnectionException
+     */
+    public static function getConcreteConnectionException(AbstractTransport $transport)
+    {
+        return new GuzzleException(new TransferException());
+    }
+}
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,8 @@
     ],
     "require": {
         "php": ">=5.3.3",
-        "psr/log": "~1.0"
+        "psr/log": "~1.0",
+        "elasticsearch/elasticsearch": "~2.0"
     },
     "require-dev": {
         "guzzlehttp/guzzle": "5.3.*"
--- a/lib/Elastica/Adapter/Elasticsearch/TransportAdapter.php
+++ b/lib/Elastica/Adapter/Elasticsearch/TransportAdapter.php
@@ -1,0 +1,73 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: thomas
+ * Date: 2015/11/23
+ * Time: 10:24 PM
+ */
+
+namespace Elastica\Adapter\Elasticsearch;
+
+
+use Elastica\Request;
+use Elasticsearch\Transport;
+
+/**
+ * Simple adapter so we can performRequest on a Request object,
+ * instead of a set of strings and arrays
+ *
+ * Class TransportAdapter
+ * @package Elastica\Adapter\Elasticsearch
+ */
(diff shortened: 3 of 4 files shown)


lib/Elastica/Exception/Factory/ConnectionExceptionFactory.php
```diff
@@ -0,0 +1,36 @@
...
+ */
+class ConnectionExceptionFactory
+{
+    /**
+     * @param AbstractTransport $transport
+     * @return ConnectionException
+     */
+    public static function getConcreteConnectionException(AbstractTransport $transport)
+    {
+        return new GuzzleException(new TransferException());
```
CrAzE124: This is obviously not correct - just a PoC.

ruflin:
@CrAzE124 Very interesting. I wasn't even thinking of what you just did here but this is a very interesting approach. It makes it possible to first test if the integration works as expected and move in a second step.

In the long term my idea is that elasticsearch-php replaces to full transport and connection layer of Elastica. This is the part where the two libraries have a lot of code in common and improvements always have to be made on both sides. So elasticsearch-php is like the transport and connection library of Elastica :-)

CrAzE124:
@ruflin yes I thought that was your plan! If you're happy with the general idea (and with the use of a couple more adapter classes) I can move forward on this sometime in the week? 

As far as I understand the requirement, we need to refactor the codebase to implement the transport and connection facilities as provided by php-elasticsearch, whilst maintaining backwards compatibility. Correct?

ruflin:
The question is how much more work would it be to go the full way and extend Elastica\Client by elasticsearch-php client and remove all Transport and connections on the Elastica side? With moving to elasticsearch 2.0 we will have lots of BC breaks, so it would also be a good timing to do the above. The reason I like to go the full way directly is because it will also remove a large portion of code on our side. The main method that would have to be changed is Elastica\Client::request(...): https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Client.php#L615 Inside it should do what you did in the change above. If this is the case, we could probably even change it without major BC breaks if someone doesn't have very special transport and connection configs.

Currently it sounds quite simple but I'm sure I forgot a lot of things. I'm happy to also jump on a call to have direct conversation which could be worth for the above case.

CrAzE124:
@ruflin yes I see what you're saying with regards to Transport and Connections. I think it can be done without much BC breakage. 

I think we should schedule a Skype/Hangouts session - pop me a mail (I believe my email is public) and we can set something up!

ruflin:
@CrAzE124 Any updates here. One thing that came to my mind which probably is going to be a challenge: Map the configuration options from Elastica\Client to elasticsearch-php transport.

CrAzE124:
@ruflin sorry, been extremely busy with work! Will be looking at this again tonight.

ruflin:
:+1:

CrAzE124:
@ruflin 2 questions,

1 - Are we trying to maintain BC with this? I'm thinking I can just about strip out the entire `Connection` and `Transport` namespaces if we don't need to maintain BC. If we do, perhaps I'll just replace the guts of the classes in these namespaces with some nice little wrappers.  
2 - What version of Elastica should I be building against?

ruflin:
For BC: If possible, but not required. I'm happy to get rid of the Connection and Transport layer as this would only be "half" a BC break, as I assume most of the people are not accessing it directly. The question here is more about the configuration options if we keep it or not.

For Elastica version: 3.0.*. Based on how fast we get this in we could even get it in for version 3 as we are still in beta, otherwise I would also be willing to go to 3.1 or even 4.0 quite soon.

ruflin:
@CrAzE124 Did you make any progress here?

ewgRa:
ping @CrAzE124 ? Do you need any help here?

CrAzE124:
@ewgRa sorry, yes, I think I might be in over my head here.

ruflin:
@ewgRa @CrAzE124 Reviewing old PR's I was stumbling over this one. By now we have the official elasticsearch client for queries. I still really like this idea to also have it as an adapter. As soon as we have it as an adapter and it works well, it could also be enabled as default adapter in the long run. Anyone interested to push this one forward?

ewgRa:
@ruflin I'm still interested, but want return to it later, have another several things in priority and not so much free time.

But from my point of view, make it as adapter - it is partial solution, no? And it is not enough. The best solution - all transport layer, as low-level layer, must be based on official client. So, all adapters, any cases of connections (AWS, etc.) must be supported by official client and must be out of scope for Elastica.
Of course, this is a broke backward compatibility, but it is forward step, not a side step as adapter.

ruflin:
@ewgRa Fully agree. The end result will be what you described above. The part I'm hoping with first going with the adapter is that it allows us to learn on how it works and it splits up the work in smaller and more digestable pieces. If you want to go all in, also good for me :-D

ruflin:
Closing this PR as by now we have the client in and this one is pretty old. But there is still work do be done ;-) @CrAzE124 Thanks for initially kicking this off.

PR #990, opened 2015-11-23
fix bulk conflict between document and metadata

ewgRa:
This is a draft version of PR that will fix problems with CrossIndex failed tests.

Actually problem in Bulk actions.

Elastica try set _id in metadata and also in document, now Elasticsearch consider it as error.

Idea is to protect Bulk actions from misuse, when source is setted with metadata fields.


ruflin:
@ewgRa Let me know when this is ready to be merged.

ewgRa:
I saw you fix it already, this PR also deprecate _percolate param and add more asserts at CrossIndexTest

ruflin:
Thanks. Sorry for fixing some parts twice.

PR #995, opened 2015-11-26
fix NestedAggregationTest

ewgRa:
Actually it is strange how it work before. Since NestedAggregation introduced in documentation there is field named as in fix.

I found a reason why test fail, but not sure that this PR just not to hide problem.

### DIFF
--- a/test/lib/Elastica/Test/Aggregation/NestedTest.php
+++ b/test/lib/Elastica/Test/Aggregation/NestedTest.php
@@ -51,14 +51,13 @@ class NestedTest extends BaseAggregationTest
     {
         $agg = new Nested('resellers', 'resellers');
         $min = new Min('min_price');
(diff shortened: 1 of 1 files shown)


ruflin:
@ewgRa Lol, I just made the same fix a few minutes ago. I think the reason that it worked before is that ES < 2.0 made some attempts to find fields when they were not defined correctly which is not the case anymore in 2.0

ewgRa:
My was first :) Seems I need install gitter at home :)

And seems there is no possibility to notice user about this change, that actually kind of BC.

ruflin:
As it didn't conflict with my changes, I still merged it.

ruflin:
@ewgRa No change on our side needed. Lots of other BC changes exist that are related to ES 2.0 and users need to know about it.

PR #998, opened 2015-12-03
Add context suggestor and return only suggest result

fubuki:


ruflin:
@fubuki Thanks for the contribution. Could you add 2 things?
* Update the CHANGELOG.md file
* Add some tests? Best some integration tests if possible to make sure it works as expected.

lib/Elastica/Search.php
```diff
@@ -418,6 +425,10 @@ public function getPath()
             $path .= '/'.implode(',', $types);
         }
 
+        if (true == $this->_onlySuggest) {
```
ewgRa: Maybe better to be more strict "true ===" ?

lib/Elastica/Search.php
```diff
@@ -449,6 +460,10 @@ public function search($query = '', $options = null)
             $data = $query->toArray();
         }
 
+        if (true == $this->_onlySuggest) {
```
ewgRa: Maybe better to be more strict "true ===" ?

lib/Elastica/Search.php
```diff
@@ -523,6 +538,15 @@ public function setSuggest(Suggest $suggest)
     }
 
     /**
+     * @param boolean $onlySuggest
+     */
+    public function onlySuggest($onlySuggest = false)
```
ewgRa: Maybe better to have by default $onlySuggest = true?

When I will try to call for example $search->onlySuggest() - I expected that onlySuggest will be enabled, but not disabled. It is how code will be readed.

ewgRa:
@ruflin @fubuki What do you think about implement it like SuggestSearch?

I'm not sure, but seems "context" is not related to _search for example, it is _suggest thing. And Search class start mixing different things when suggest functionality moved there.

ruflin:
@ewgRa Thanks for bringing this up. I'm a little bit unsure where to put it, as it seems like a filter on the search result. At the same time, it calls a different API on line 429 which is a little bit unexpected.

fubuki:
@ruflin @ewgRa  Thanks for your reply. I will update the change log file and test.

Why I use the different api for suggest?   I  need develop the  autocomplete function in the production environment so I used elasticsearch and context suggester.

But there are some problem when I trace the search result. If you use the _search api for suggest,  elasticsearch will return the hit and suggest result. Then it will take more time than using _suggest api.

use _search api
```  
{
  "took": 228,
  "timed_out": false,
  "_shards": {
    "total": 5,
    "successful": 5,
    "failed": 0
  },
  "hits": {
``` 
use _suggest api
```
{
  "_shards": {
    "total": 5,
    "successful": 5,
    "failed": 0
  },
  "song_suggest": [
    {
```

ewgRa:
@fubuki Maybe you a re right, it depends what is Search implement, API, or not. I think @ruflin mean "endpoint", not an API.


My 2 cents, that Elastica must show how to use API clearly and prevent misuse.

For example if we mix two endpoints in one Search class, we can do something like this:
```
$search = new Search();
$search->onlySuggest();
$search->setQuery(...);
```
As far as I understand, set query for _suggest make no sense, and would be nice if Elastica prevent to misuse of Search API and when it is onlySuggest - at setQuery throw exception. But catch all this situations and have a lot of checks - make maintenance of code harder, and make harder work with Elastica, because developer maybe don't know that when it is onlySuggest - he can't set query.

Another side when we use Search and Suggest for example.
When I work with Suggest object, my IDE will show me what I have and what I can set.

I think it is big question. If Search is implementation of Search API, - than good point to have it in one class and concern about  misuse in one class. If not, than it can be done like: add Api directory, there will be Api/Search/Search and Api/Search/Suggest for example. Deprecate Elastica/Search for long term and add backward compatibility layer.


I'm sorry that I make things harder and more complex :)
What do you think on all of this? On my mind there is only advantages and more friendly for developers. Disadvantage only that it is broke BC, but we can provide backward compatibility for a long term.

ruflin:
I think what is clearly missing at the moment is the API call for _suggest. Could we extend the Suggest object to also make API calls? https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Suggest.php

@ewgRa About your suggestion: I have to think this through in more details also related to all the features and API's that exist now. Search for me is like a "general" object to do search queries. Is _suggest also a search? I would argue yes, but I agree it should always us _search and not _suggest.

ruflin:
Closing this PR as no changes in a few years. Please reopen if needed.

Issue #1003, opened 2015-12-08
json_decode() and integer overflow strike back


theDisco:
I have just installed version 3.0.0-beta1 after reading #717 and #941 in hope to get the integer overflow problem in json_decode() sorted out. After some debugging I have found following problems:

1) In some of the older releases of php5-json Debian package `JSON_BIGINT_AS_STRING` is not implemented throwing a warning `[Warning] json_decode(): option JSON_BIGINT_AS_STRING not implemented`. Below my version table:

```
vagrant@dev-box:/srv/app$ apt-cache policy php5-json
php5-json:
  Installed: 1.3.2-2build1
  Candidate: 1.3.9-1+deb.sury.org~trusty+2
  Version table:
     1.3.9-1+deb.sury.org~trusty+2 0
        500 http://ppa.launchpad.net/ondrej/php5/ubuntu/ trusty/main amd64 Packages
 *** 1.3.2-2build1 0
        500 http://archive.ubuntu.com/ubuntu/ trusty/main amd64 Packages
        100 /var/lib/dpkg/status
```

After upgrading to 1.3.9 json_decode() does not complain about `JSON_BIGINT_AS_STRING` anymore. Maybe you should look at how [firebase/php-jwt](https://github.com/firebase/php-jwt/blob/master/src/JWT.php#L247) (found [here](http://stackoverflow.com/questions/19520487/json-bigint-as-string-removed-in-php-5-5)) has solved the problem and implement it similarly?

2) After sorting out the problem with json_decode() and `JSON_BIGINT_AS_STRING` I get following error now `Fatal error: Cannot use object of type stdClass as array in /srv/app/vendor/ruflin/elastica/lib/Elastica/Response.php on line 121` which can be traced to [this line](https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Response.php#L208). Fix is very simple, just replace false with true, but what is more important, I couldn't find a test for it anywhere in the test directory. I am more than happy to provide a PR for that.

3) A smaller thing, but it would save me a lot of time in debugging. The default definition of `bigintConversion` config variable from [here](https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Client.php#L39) should be actually listed under `config` key, since it is obtained by the connection from the config in [here](https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Transport/Http.php#L157). I can provide tests for that too.

Can you please confirm my findings?

ruflin:
[...]

PR #1005, opened 2015-12-09
Fix json_decode problem

theDisco:
As discussed in #1003 here is the PR that should at least make elastica to work properly when bigintConversion setting is set to true.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ All notable changes to this project will be documented in this file based on the
 
 ### Bugfixes
 - Function score query: corrected the `score_method` `average` to `avg` #975 
+- Set `json_decode()` assoc parameter to true in `Elastica\Response` #1005
+- Add `bigintConversion` to keys passed to connection config in `Elastica\Client` #1005
 
 ### Added
 - Elastica\Query\MultiMatch::setFuzziness now supports being set to `AUTO` with the const `MultiMatch::FUZZINESS_AUTO`
--- a/test/lib/Elastica/Test/ResponseTest.php
+++ b/test/lib/Elastica/Test/ResponseTest.php
@@ -201,6 +201,23 @@ class ResponseTest extends BaseTest
     }
 
     /**
+     * @group unit
+     */
+    public function testDecodeResponseWithBigIntSetToTrue()
+    {
+        $response = new Response(json_encode(array(
+            'took' => 213,
+            'items' => array(
+                array('index' => array('_index' => 'rohlik', '_type' => 'grocery', '_id' => '707891', '_version' => 4, 'status' => 200)),
+                array('index' => array('_index' => 'rohlik', '_type' => 'grocery', '_id' => '707893', '_version' => 4, 'status' => 200)),
+            ),
+        )));
+        $response->setJsonBigintConversion(true);
+
+        $this->assertTrue(is_array($response->getData()));
+    }
+
+    /**
      * @group functional
      */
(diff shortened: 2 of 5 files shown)


test/lib/Elastica/Test/ResponseTest.php
```diff
@@ -201,6 +201,23 @@ public function testIsNotOkBulkItemsWithStatusField()
...
+        $response = new Response(json_encode(array(
+            'took' => 213,
+            'items' => array(
+                array('index' => array('_index' => 'rohlik', '_type' => 'grocery', '_id' => '707891', '_version' => 4, 'status' => 200)),
+                array('index' => array('_index' => 'rohlik', '_type' => 'grocery', '_id' => '707893', '_version' => 4, 'status' => 200)),
+            ),
+        )));
+        $response->setJsonBigintConversion(true);
+
+        $this->assertTrue(is_array($response->getData()));
```
ruflin: This test would also pass without the setJonBigintVersion(true), right?

ruflin:
@theDisco Thanks. Can you also update the CHANGELOG.md?

test/lib/Elastica/Test/ResponseTest.php
theDisco: Yes, this test would also pass without L215, but L215 makes the difference :)

theDisco:
@ruflin Sure, done.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ All notable changes to this project will be documented in this file based on the
 
 ### Bugfixes
 - Function score query: corrected the `score_method` `average` to `avg` #975 
+- Set `json_decode()` assoc parameter to true in `Elastica\Response` #1005
```
ruflin: Sorry to bring this up again. But isn't the affect here, that if BIG_INT is enabled or not could be that the response object looks different?

test/lib/Elastica/Test/ResponseTest.php
ruflin: I'm fine with the test. But I assume this test would also have passed without the code changes you made? It is more for me to understand if we test if the bug was actually fixed or if we test that we can use the function without any side affects?

theDisco: No, it wouldn't since the return value of https://github.com/ruflin/Elastica/pull/1005/files#diff-b364251d02e70310f3c6e443372397d3L208 would be `\stdClass` and not an array and there is not conversion or casting happening till the end of method.

CHANGELOG.md
theDisco: BIG_INT_* setting converts integers bigger than PHP_INT_MAX to a string, so yes, the response will be different, but it is the only way to deal with this without overflowing the integer.

ruflin:
@theDisco Thanks for the contribution. We should definitively discuss if we should use `JSON::parse($response, true)` for the non JSON_BIG_INT conversion case. But I'm not sure if that will have any other side affects.

theDisco:
@ruflin if you look at the implementation of `JSON::parse()` you actually do not have to pass true as a second parameter:

```php
        // extract arguments
        $args = func_get_args();
        // default to decoding into an assoc array
        if (sizeof($args) === 1) {
            $args[] = true;
        }
        // run decode
        $array = call_user_func_array('json_decode', $args);
```

And no, passing true would not have any side effects.

ruflin:
@theDisco Ugh :-( Surprising that nobody of us spotted that in the past. But that is great news.

PR #1006, opened 2015-12-11
Elastica\Query\Filtered is deprecated in favour of Elastica\Query\BoolQuery::addFilter()

kukulich:


lib/Elastica/Query/Filtered.php
```diff
@@ -21,6 +23,8 @@ class Filtered extends AbstractQuery
      */
     public function __construct(AbstractQuery $query = null, AbstractFilter $filter = null)
     {
+        trigger_error('Use BoolQuery instead. Filtered query is deprecated since ES 2.0.0-beta1 and this class will be removed in further Elastica releases.', E_USER_DEPRECATED);
```
ewgRa: 1. Better to have trigger error even if developer just include file. See for example https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Query/Bool.php

lib/Elastica/QueryBuilder/DSL/Query.php
```diff
@@ -196,7 +195,7 @@ public function field()
      */
     public function filtered(AbstractQuery $query = null, AbstractFilter $filter = null)
     {
-        return new Filtered($query, $filter);
+        throw new DeprecatedException('Use BoolQuery instead. Filtered query is deprecated since ES 2.0.0-beta1 and this class will be removed in further Elastica releases.');
```
ewgRa: I'm not sure, but if Filtered query just deprecated now, but still can be used, than better here just trigger deprecated error, instead of throw Exception. Also I think here nice to add deprecated info to phpdoc.

ewgRa: And if Filtered deprecated and can't be used now, than constructor must throw DeprecatedException. Main idea to prevent developer do something that make no sense now.

ewgRa:
@kukulich I have question about this PR. Can you clarify, Filtered query deprecated, but not removed? So, in ES 2.0 it still can be used?

If it is deprecated and removed - than exeption must be thrown. If not - than trigger errors only. Now you have in one place - trigger error, in another - throw exception. I think it must be synced. And tests also depends on what happens with Filtered - it is removed, or not.

lib/Elastica/Query/Filtered.php
kukulich: I have tried it but it's not possible to make `FilteredTest` to work then. This is the only way how I was able to make tests work.

kukulich:
@ewgRa You're right. I will remove `DepracatedException` and just trigger an error because filter query still works in ES 2.0.

kukulich:
@ewgRa Done

ruflin:
@ewgRa Please directly merge it as soon as you feel comfortable with the change. I'm mostly on my mobile this week ...

ewgRa:
@kukulich Thanks for PR, it is mostly ready to be merged, just want ask you about some things.

Can you add mark deprecated and trigger error at Elastica\QueryBuilder\DSL\Query::filtered method? And in Elastica\Test\QueryBuilder\DSL\QueryTest check that deprecate error triggered. Also would be nice to call Query::filtered method twice, or load Filtered class before, to be sure that error triggered in method itself and not in loading Filtered class.


I want suggest move trigger deprecated error from constructor to file itself, like in https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Query/Bool.php.

You said that you have in this case error with FilteredTest. So, I try and see what are you talking about. And I want to suggest complete it in this way:

1. Move trigger error from constructor to whole file. My concern here that we need kind of standard general way to deprecate classes, because classes can have static methods, trigger error in constructor not cover call static methods in deprecated classes, also as there can be unserialize things. So, on my mind classes must be deprecated just as whole file.

2. In FilteredTest in setUp supress E_USER_DEPRECATED, and in tearDown - return it. Like: 
setUp - error_reporting(error_reporting() & ~E_USER_DEPRECATED);
tearDown - error_reporting(error_reporting() | E_USER_DEPRECATED);
After this there is no needed to change all test cases for catch this deprecated error.

3. Add test that Filtered class raise deprecated error when class is loaded first time. It can be just first test case at FilteredTest. Yes, it can be broken at some cases related to in which order test runs, but I think it is never happen and sooner or later Filtered will be removed.


So, what do you think? Do you have time for this?

 I can prepare PR to your repository with this changes if you haven't time on this.

VasekPurchart:
I don't think that triggering an error at the file level is a good idea, because there are quite a lot tools/libraries, which scan entire folders and include/require files to find classes etc. This would then trigger the error even when noone is trying to use the deprecated files themselves and can be quite confusing.

Which is also the problem with the 3rd point you wrote, I think that when there are a lot more of the deprecated classes, it would be quite tedious (and maybe in some combinations even impossible) to have tests for these.

This all nails down to a reasonable convention - that files, which define new symbols should have no side effects (execute something).

ewgRa:
Can you name tools/libraries that have problems with that?

Symfony for example have same deprecation convention - they deprecate whole class to mention that whole class is deprecated and as far as I know - there is no problems with another libraries/tools. One thing why they don't want use constructor - a lot of similar deprecation messages when object created.
And it is already in Elastica deprecation for whole classes - in Bool for example.

3rd problem is not a big problem, because usually tests grouped by classes.

VasekPurchart:
For example Doctrine [AnnotationDriver](https://github.com/doctrine/common/blob/v2.6.0/lib/Doctrine/Common/Persistence/Mapping/Driver/AnnotationDriver.php#L236), I know this is something you should not let search your vendor complete vendor dir, but I have seen such cases (and maybe other tools, where it makes more sense, practically every tool that goes through files without tokenizing the content).

But yes, I looked at Symfony and they are using this if a whole class is deprecated, so I guess the real life impact is ok.

ewgRa:
@kukulich ?

kukulich:
@ewgRa I will try to do it.

kukulich:
@ewgRa Force pushed. I really don't know how to do it any better.

ewgRa:
@kukulich Good job! I think most important that now whole class deprecated. How it looks like in tests - it is kind of subjective. If you want to see how I saw it - feel free to message me (Github, Gitter channel).

On quick view it is ready to be merged. I will check carefully on this weekend.

Just one small thing left, can you add to changelog that lib/Elastica/QueryBuilder/DSL/Query::filtered method is deprecated.

Thanks again!

kukulich:
@ewgRa Done.

ewgRa:
Thanks @kukulich !

ruflin:
Thanks @kukulich and @ewgRa for the great work.

PR #1009, opened 2015-12-13
Fix inconsistently failing IndexExists test

JustinHook:
Elasticsearch HTTP calls are asynchronous.

Previously the IndexExists test was calling
```
$index->create();
$status->refresh();
$this->assertTrue($status->indexExists($indexName));
```
Due to the index being created asynchronously it was possible for the index not to have been created when the `refresh()` was called. Causing the test to sometimes fail.
This PR adds a small 10ms sleep to allow elasticsearch to create the index before calling `refresh()`.


ruflin:
Thx. Probably not 100% reliable but better then before

Issue #1010, opened 2015-12-15
Use POST to send bulk requests


hakman:
A RESTful design uses PUT for updates and POST for create. Elastica uses the bulk API to create new documents:
https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Bulk.php#L356

Would it be OK to change the way Elastica sends bulk requests from PUT to POST?
Elasticsearch docs also mention POST as the method for sending bulk requests, though the spec allows both:
http://www.elastic.co/guide/en/elasticsearch/reference/master/docs-bulk.html

damienalexandre:
Good point. No reason to use PUT here, but elasticsearch handle them [exactly the same](https://github.com/elastic/elasticsearch/blob/master/core/src/main/java/org/elasticsearch/rest/action/bulk/RestBulkAction.java).

The official PHP client is also [using POST](https://github.com/elastic/elasticsearch-php/blob/master/src/Elasticsearch/Endpoints/Bulk.php#L80).

ewgRa:
I think you right and it must be synced and POST more right in this case, so +1 for POST.

@hakman Did you face with real problem when PUT used? Or there is no something behind?

hakman:
We have a product that works as a proxy for Elasticsearch to accept and parse logs.
I am writing some examples for that using various languages and libs. One of them is Monolog and had trouble shipping logs with it, because of this issue.

We are also modifying our proxy, but thought this might help others in the future.

otisg:
+1 for supporting POST

ruflin:
I remember there was a reason we used PUT, but I think it was related to one of the deprecated transports. So +1 on this change.

ruflin:
Closing as https://github.com/ruflin/Elastica/pull/1011 was merged.

PR #1011, opened 2015-12-16
Send bulk requests using POST instead of PUT

hakman:
Fixes issue #1010.


ewgRa:
I wondering, is this the backward compatibility broke or not?

Theoretically this change can broke something, because you have real case why this change needed and change must be added to changelog as BC broke.

What do you think?

hakman:
ES treats both PUT and POST bulk the same way, so I don't see any backwards compatibility issue. This would just make the Elastica behave like most other ES libs.
https://github.com/elastic/elasticsearch/blob/2.1/rest-api-spec/src/main/resources/rest-api-spec/api/bulk.json

ruflin:
@hakman LGTM. Can you update the CHANGELOG.md?

hakman:
Done, thank you!

ruflin:
@hakman Thanks for the contribution.

PR #1013, opened 2015-12-28
Update composer.json

rdohms:
Changed PHP version restriction to NOT allow PHP 7, since this library is not PHP 7 ready (due to the Bool class)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         }
     ],
     "require": {
-        "php": ">=5.4.0",
+        "php": "^5.4.0",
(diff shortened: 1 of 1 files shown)


ruflin:
@rdohms It is PHP 7 ready. The Bool classes are deprecated and BoolQuery should be used instead. See also here: https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Query/Bool.php#L4 As long as the Bool class is not used, there shouldn't be any problems with PHP 7.

PR #1016, opened 2015-12-31
Refactor response exception and make it more BC compatible

ruflin:


ruflin:
@nickshanks I already merge this to prepare for the 3.0 release. But please still have a look and comment on it.

JarJak:
+1

ruflin:
@nickshanks I pushed and updated and left some comments. Is there a reason you comment on commits and not the PR changes? Just curious.

JarJak:
Could this be also merged to ~2.3? FOSElasticaBundle requires it and the problem still exists in there.

ruflin:
@JarJak Isn't it only an issue if elasticsearch 2.0 is used?

JarJak:
Could be, however FOSElasticaBundle works fine with elasticsearch 2.0 if not this BC break.

ruflin:
@JarJak Normally I didn't provide fixes for past releases as maintaining it gets much harder. What we could do in case this is a bigger issue is that I create a 2.3 branch out of the most recent 2.3 tag. Then you could open a PR with the fix against this branch based on which I can create a tag and add a release. What do you think?

nickshanks:
"Is there a reason you comment on commits and not the PR changes." None other than that's the first place where I saw the code.

ruflin:
@nickshanks Good reason :-)

nickshanks:
Wow, you're fast. It takes me 8 days to catch up with those emails.

ruflin:
@nickshanks Morning is my "e-mail catch-up time", so the timing was right :-)

JarJak:
@ruflin what is exactly the root of "bigger issue"? Why it couldn't be just merged to 2.3 branch and be tagged as 2.3.2?

ruflin:
@JarJak With "bigger issue" I was referring to you side. Means, in case it is a blocker. The release management of Elastica is very simple and always based on master. So there is currently no 2.3 branch. Like described before, I could create the branch if necessary.

JarJak:
I would be nice to create 2.3.x branch with this fix.

ruflin:
I created the branch `release-2.3` from the tag 2.3.1 here: https://github.com/ruflin/elastica/tree/release-2.3 Please open a PR with your requested changes agains this branch so I can merge it and afterwards tag 2.3.2

JarJak:
2.x needs exactly this PR

ruflin:
@JarJak This PR is already merged and pointed against master so a new one must be created (with the same changes). Can you please do that?

Issue #811, opened 2015-04-06
Object transformations


merk:
A feature that I'm looking to push upstream to Elastica is the handling of transformation between an object and its 'indexable representation'. FOSElasticaBundle provides mapping capabilities (see issue #810) which allows the bundle to accept application objects and convert them to a representation that can be sent to ElasticSearch which may not be the same.

We provide events that let the developer listen for these transformations and make changes to the final Elastica document before it is sent.

Our primary transformer is defined here: https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/Transformer/ModelToElasticaAutoTransformer.php#L60-121 but we support custom approaches in case transformations have unique requirements. Additionally, we support libraries like symfony/serializer and jms/serializer for transformation to JSON.

im-denisenko:
This actually mean implement analog of [doctrine/mongodb-odm](https://github.com/doctrine/mongodb-odm) for elasticsearch, right?

merk:
No. What I propose is transformation of an ODM or ORM object into an Elastica document. FOSElasticaBundle also provides a way to convert an Elastica Result object into its original ORM/ODM object but I'm not sure how useful that is for general Elastica users.

im-denisenko:
> transformation of an --- object into an --- document
> convert an --- result --- into --- object

But isn't it what ODM mean?


> how useful that is for general Elastica users

As general Elastica user, I bet it will be super awesome :)

merk:
I guess it depends if you're planning on using ElasticSearch as a primary document store or not. FOSElasticaBundle doesnt make this assumption. It provides functionality that lets you populate the search indexes based on your primary document stores.

The point of this feature is basically to allow the conversion of your App's BlogPost document to a representation that you can index, performing transformations on data as required by the mappings known. In a lot of cases the fields will be the same, but that isnt a safe assumption to make.

Yes, this could be talk about some kind of mapper for ES, but this conversation is specifically about a service that handles a conversion, lets create a separate issue to talk about some kind of mapper if there is interest in that.

im-denisenko:
I got it. Previously I thought that FosElastica can automagically convert Elastica document into BlogPost object, bypassing ORM mechanism. Now I have seen transformers source code.

But still, mapping config, transformers and potential ODM are tight coupled. I'll try have a look how implement all this in generic way without half of Symfony's stack, if no objections will come from @ruflin or someone will take care of this.

merk:
[...]

PR #1017, opened 2016-01-02
Result Transformations

merk:
Comes from #811 

This is an initial concept for implementing a service that adds a transformed object into a Result.

The first commit moves construction of Result and ResultSet objects into its own service, and is a bit clumsy in the Elastica\Client instance so that appropriate places can get access to it without modifying the current API. It might be worth having a conversation about closing down the Elastica API surface to a single prescribed way of doing things to make adding features like this simpler.

The second commit adds support for a TransformerInterface to be implemented to handle transformation of ResultSet's Result objects based on object identifiers.

The third commit adds ORM and ODM support which fundamentally comes from https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/Doctrine/ORM/ElasticaToModelTransformer.php and https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/Doctrine/MongoDB/ElasticaToModelTransformer.php

There is a lot more to do before its ready, but it shows the direction I've been thinking about. I'd like to see better support for more than just object identifiers, implementing half a dozen transformers like Symfony\Serializer, Doctrine ORM and MongoDB.

The other half of the transformations will require more thought (mixed objects to Elastica Documents) because it needs to have knowledge about ES mappings or be able to introspect the objects passed to do the conversion magically.

### DIFF
--- a/lib/Elastica/ResultSet.php
+++ b/lib/Elastica/ResultSet.php
@@ -15,117 +15,45 @@ use Elastica\Exception\InvalidException;
 class ResultSet implements \Iterator, \Countable, \ArrayAccess
 {
     /**
-     * Class for the static create method to use.
-     *
-     * @var string
-     */
-    protected static $_class = 'Elastica\\ResultSet';
-
-    /**
      * Results.
      *
-     * @var array Results
+     * @var Result[] Results
      */
-    protected $_results = array();
+    private $_results = array();
 
     /**
      * Current position.
      *
      * @var int Current position
      */
-    protected $_position = 0;
+    private $_position = 0;
 
     /**
      * Response.
      *
      * @var \Elastica\Response Response object
      */
-    protected $_response = null;
+    private $_response;
 
     /**
      * Query.
      *
      * @var \Elastica\Query Query object
      */
-    protected $_query;
-
-    /**
-     * @var int
-     */
-    protected $_took = 0;
-
-    /**
-     * @var bool
-     */
-    protected $_timedOut = false;
-
-    /**
-     * @var int
-     */
-    protected $_totalHits = 0;
-
-    /**
-     * @var float
-     */
-    protected $_maxScore = 0;
+    private $_query;
 
     /**
      * Constructs ResultSet object.
      *
-     * @param \Elastica\Response $response Response object
-     * @param \Elastica\Query    $query    Query object
+     * @param Response $response Response object
+     * @param Query $query Query object
+     * @param Result[] $results
      */
-    public function __construct(Response $response, Query $query)
+    public function __construct(Response $response, Query $query, $results)
     {
-        $this->rewind();
-        $this->_init($response);
         $this->_query = $query;
-    }
-
-    /**
-     * Creates a new ResultSet object. Can be configured to return a different
-     * implementation of the ResultSet class.
-     *
-     * @param Response $response
-     * @param Query    $query
-     *
-     * @return ResultSet
-     */
-    public static function create(Response $response, Query $query)
-    {
-        $class = static::$_class;
-
-        return new $class($response, $query);
-    }
-
-    /**
-     * Sets the class to be used for the static create method.
-     *
-     * @param string $class
-     */
-    public static function setClass($class)
-    {
-        static::$_class = $class;
-    }
-
-    /**
-     * Loads all data into the results object (initialisation).
-     *
-     * @param \Elastica\Response $response Response object
-     */
-    protected function _init(Response $response)
-    {
         $this->_response = $response;
-        $result = $response->getData();
-        $this->_totalHits = isset($result['hits']['total']) ? $result['hits']['total'] : 0;
-        $this->_maxScore = isset($result['hits']['max_score']) ? $result['hits']['max_score'] : 0;
-        $this->_took = isset($result['took']) ? $result['took'] : 0;
-        $this->_timedOut = !empty($result['timed_out']);
-        if (isset($result['hits']['hits'])) {
-            foreach ($result['hits']['hits'] as $hit) {
-                $this->_results[] = new Result($hit);
-            }
-        }
+        $this->_results = $results;
     }
 
     /**
@@ -227,7 +155,7 @@ class ResultSet implements \Iterator, \Countable, \ArrayAccess
      */
     public function getTotalHits()
     {
-        return (int) $this->_totalHits;
+        return isset($result['hits']['total']) ? (int) $result['hits']['total'] : 0;
     }
 
     /**
@@ -237,7 +165,7 @@ class ResultSet implements \Iterator, \Countable, \ArrayAccess
      */
     public function getMaxScore()
     {
-        return (float) $this->_maxScore;
+        return isset($result['hits']['max_score']) ? (float) $result['hits']['max_score'] : 0;
     }
 
     /**
@@ -247,17 +175,17 @@ class ResultSet implements \Iterator, \Countable, \ArrayAccess
      */
     public function getTotalTime()
     {
-        return (int) $this->_took;
+        return isset($result['took']) ? $result['took'] : 0;
     }
 
     /**
-     * Returns true iff the query has timed out.
+     * Returns true if the query has timed out.
      *
      * @return bool Timed out
      */
     public function hasTimedOut()
     {
-        return (bool) $this->_timedOut;
+        return !empty($result['timed_out']);
     }
 
     /**
--- a/lib/Elastica/ResultSet/Builder.php
+++ b/lib/Elastica/ResultSet/Builder.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace Elastica\ResultSet;
+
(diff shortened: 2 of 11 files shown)


merk:
I've only been able to run unit tests because of my remote location, so there could be major functional failures which I'll look at when the approach is :+1: to continue.

lib/Elastica/ResultSet.php
```diff
@@ -14,117 +14,45 @@
...
-    protected static $_class = 'Elastica\\ResultSet';
-
-    /**
      * Results.
      *
-     * @var array Results
+     * @var Result[] Results
      */
-    protected $_results = array();
+    private $_results = array();
```
ruflin: Is there a specific reasons these were made private?

lib/Elastica/ResultSet.php
```diff
@@ -211,7 +139,7 @@ public function getAggregation($name)
      */
     public function getTotalHits()
     {
-        return (int) $this->_totalHits;
+        return isset($result['hits']['total']) ? (int) $result['hits']['total'] : 0;
```
ruflin: +1 on moving this into the getters to somplify the init.

lib/Elastica/ResultSet/Builder.php
```diff
@@ -0,0 +1,91 @@
+<?php
+
+namespace Elastica\ResultSet;
```
ruflin: I really like the idea of a result builder

lib/Elastica/Transformer/AbstractTransformer.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace Elastica\Transformer;
```
ruflin: +1 on the transformer idea.

ruflin:
@merk Thanks a lot for pushing this forward. I quite like the idea of the builder and transformer. So the first two commits look quite good to me (didn't check them in detail).

For the transformers itself I'm not 100% sure if they should be part of the Elastica repository or for example in ElasticaTransformers. Like this Elastica would provide interface and abstraction but not be cluttered with all the specific implementations. This especially relates to the dependencies which I try to keep as light as possible. Adding doctrine for everyone that uses Elastica is quite heavy. 

Do you think it would be possible to extract the transformers?

About the event stuff I think I have to read up on the symfony part. Is this like PubSub?

lib/Elastica/ResultSet.php
merk: Habit mostly, I dont like protected unless there is a very specific reason for it to be extendable. It makes code more maintainable.

merk:
I think putting transformer implementations in Elastica isn't really a big deal because they're optional - they don't actually have to bring in hard dependences, that only has to be done if a specific transformer will be used.

The Doctrine dependency is optional, and only specified in dev-dependencies for testing purposes, as is the Symfony dispatcher (though, personally I'd really like to make it mandatory). More detail about the Symfony dispatcher (which, in the documentation says it implements the Mediator pattern): http://symfony.com/doc/current/components/event_dispatcher/introduction.html

For the dispatcher, I actually want to implement it elsewhere in Elastica to enable easier modifications to the way Elastica behaves without needing to override code. This will enable us to close the API surface significantly, making maintenance and feature implementation easier.

lib/Elastica/ResultSet.php
merk: The main reason for the change here is that the right place to manipulate the results is in the ResultSetBuilder, not here - private forces people to use the defined extension point.

ruflin:
@merk Sorry, took me some time to answer.

As far as I understand, Client requires Builder and Builder requires dispatcher, which indirectly means the dispatcher is required not only for dev? Not that this will be a blocker, just for understanding.

The thing I worry about the transformers is not only the dependencies but also the maintenance. As soon as a transformer is part of Elastica, people expect it to be maintained by Elastica. So the question comes up, which transformers are accepted and which ones are not. If there are only a handful of transformers to be expected, it should probably not be too big of a deal, but if there are lots of transformers coming up, it will become an issue. I could think of a ElasticaTransformers Github repo that has all the Transformers inside. Like this also Elastica could make changes to the transformer interface without directly having to update all transformers, as ElasticaTranformers depends on a predefined version.

I think in general we are on the right track here. What do you suggest as next steps?

merk:
The Builder doesn't require the dispatcher, if it isn't supplied (and it isn't by default), the Builder will not dispatch events.

The maintenance burden is a problem, but it remains a problem no matter where the code actually resides. Right now, it resides in FOSElasticaBundle, making the functionality entirely impossible to use unless you're using Symfony. I'd like to fix that, learning from the original implementation, to provide a better interface for all Elastica users for a pretty handy feature.

In my next iteration of this PR I've managed to simplify the Doctrine transformer to only need a single one for any that implement the standard Doctrine\Common\Persistence\ObjectManager interface https://gist.github.com/merk/ba0b0c20a0e4aff1e625

I dont personally see a problem accepting well built and tested transformers into Elastica, as long as we define strong interfaces and extension points.

merk:
As to the question about where to go from here, I've come to the conclusion that the API surface of Elastica doesnt allow for extensions like this, since the Client should not have a hard dependency on the ResultSetBuilder - that belongs to the services that gather the response and return a ResultSet. I'm a big subscriber of the SOLID principles, and I'd like to see the code move in that direction. https://en.wikipedia.org/wiki/SOLID_(object-oriented_design)

I've spent some time trying to extract some of the functionality in the \Elastica\Client class and I hope to have another PR that explains a bit more about the direction I'd like to propose before we make any decisions or commitments about supporting new features like in this PR. I'd like to see Elastica grow into a more opinionated library locks down its internals and only exposes extension points where necessary, but it may end up being such a massive change that it'd be more appropriate to fork or start a new library depending on how big the diff becomes.

ewgRa:
my 2cents

I like whole idea in general.

Every code that require Elastica - solve similar problems with transformations from ES to Doctrine/etc., so, have this at Elastica - give standard way to do this.
Keep transformers in Elastica - right way and right place for this.

I understand @ruflin when he said that maintain bigger code base is harder and there is no resources for this, but if Elastica will limit itself by this doubts - there will be no move forward.

Maybe developers that interesting in transformers (like @merk) will be more involved in supporting this functionality, when they will not have it in their projects.

I also don't see any big advantages from move it to ElasticaTransformers. Keep it in Elastica will give bigger community and not split it. If Elastica can't maintain this code - there is no difference where this code placed. People will always think that Transformers - part of Elastica. And real question is - will it be in Elastica, or not.

I prefer to have it in Elastica, even if maintain this code will be not in priority for people who maintain main code base.

I believe in opensource and if opensource project will be limit herself functionality based on how many resources it have - there will be no progress :)

ewgRa:
At least it can be part of Elastica at beginning and if it will be really huge - moved to another repository and required in composer.

Isinlor:
When it comes to spiting Elastica and Transformers:
http://php-and-symfony.matthiasnoback.nl/2014/04/theres-no-such-thing-as-an-optional-dependency/
http://paul-m-jones.com/archives/4211

I think it could be a good read before deciding one way or another ;) .

ewgRa:
@Isinlor interesting, maybe split it is a good idea.

ruflin:
Perhaps some more background on why I would suggest to put it in a separate Github Repo. Having not mandatory dependencies is one of my concerns, but not the main concern (@Isinlor Thanks for sharing). If I want to release and tests are broken in one of the transformers and Transformers are part of Elastica, I would not release. If they are in a separate, I would release but would keep the dependency of the transformers to the older version of Elastica. This actually makes it possible to move forward Elastica with being blocked by for example the Transformers. In the past these issues existed with the Transport classes and to be honest, I'm quite happy we removed most of them.

@ewgRa Unfortunately I disagree about the limited resources. Limited resources drives you to make very concious decision and often leads to simpler solutions. Three examples of Elastica here are fully automated builds, large test set and no release branches. I could not have maintained Elastica without all the build and test automation over time. I'm also aware that a lot of people would like to have release branches to backport fixes, but it would remove the focus from moving forward.

@merk About moving forward: It would be great if we could find a way to move in the direction you mentioned above with small baby steps but having the overall goal in mind. So it is really great to have these large PR's to understand and discuss the general concept but then split it up into multiple smaller PR's (if possible). Not directly related to this change but probably good to know about is this change I want to make: https://github.com/ruflin/Elastica/issues/944 This will hopefully remove quite a bit of complexity and code from Elastica, so we can add it again at the other end :-)

About the repositories: I'm quite convinced we should go with 2 repositories, especially as soon as we add dependencies. But I'm happy to have a first version in Elastica to make development easier.

merk:
After taking into consideration here I've submitted 2 new PRs to split out the features.

1) A PR that extracts ResultSet building into its own service
2) A PR that builds on 1 to add a transformation interface with no implementations

I will open a new issue to discuss ownership and maintainership of other transformer implementations like the doctrine implementation I originally wrote here.

Issue #1001, opened 2015-12-04
Elastica\Filter\Nested should be returned back


kukulich:
It's not possible to filter nested objects without `Elastica\Filter\Nested`.

Eg. `Elastica\Query\Filtered::setFilter()` requires `\Elastica\Filter\AbstractFilter` so it's not possible to use `Elastica\Query\Nested`.

kukulich:
There's also an option to move all filters to `Elastica\Query` and change `Elastica\Query\Filtered::setFilter()` to accept `Elastica\Query\AbstractQuery`

ruflin:
@kukulich Good point. Would it be sufficient to just make setFilter accept also AbstractQuery? What do you mean by "move all filters" to Elastica\Query? Removing filters completely?

ruflin:
@kukulich Side note here: I think the points you are bringing up are the last major changes which are not properly done yet in Elastica 3.0.0. So I'm glad you got this discussion started.

kukulich:
All filters are deprecated in favour of queries so this may be done:

1. Add missing queries (formerly filters), eg. `ExistsQuery`, `MissingQuery`, `GeoDistanceQuery`, `ScriptQuery` etc.
2. Remove filters completely or in favour of easier migration mark them only deprecated
3. Make `setFilter()` accept only `AbstractQuery` or in favour of easier migration make it accept both `AbstractQuery` and `AbstractFilter`.
4. Do the same for aggregations.
5. Return `Elastica\Filter\Nested` and mark it deprecated if other filters will be deprecated too

Side note: Yes, I tried to migrate to Elastica 3.0.0 and found these issues :)

ruflin:
@kukulich That looks like the cleanest way to go. Mentioning @ewgRa here as he probably also has an opinion here.

A shortcut I was thinking of (in case we don't find the time to do the full change) could be to make AbstractFilter extend AbstractQuery and make setFilter() accept only AbstractQuery.

ewgRa:
Looks interesting, but I can't define how right it will be, especially without code cases, so feel free to ignore my 2 cents.

1. https://www.elastic.co/guide/en/elasticsearch/reference/2.0/query-dsl-geo-distance-query.html - seems it is still filter. So, in Elastica call it "Query" seems not so right way. It is still query and filter.

2. I'm always for provide good BC layer.

[...]

PR #1020, opened 2016-01-04
Readd Nested filter

ruflin:
See https://github.com/ruflin/Elastica/issues/1001

### DIFF
--- a/test/lib/Elastica/Test/Filter/NestedFilterWithSetFilterTest.php
+++ b/test/lib/Elastica/Test/Filter/NestedFilterWithSetFilterTest.php
@@ -1,0 +1,111 @@
+<?php
+namespace Elastica\Test\Filter;
+
+use Elastica\Document;
+use Elastica\Filter\Nested;
+use Elastica\Filter\Terms;
+use Elastica\Search;
+use Elastica\Test\Base as BaseTest;
+use Elastica\Type\Mapping;
+
+class NestedFilterWithSetFilterTest extends BaseTest
+{
+    protected function _getIndexForTest()
+    {
+        $index = $this->_createIndex();
+        $type = $index->getType('user');
+
+        $type->setMapping(new Mapping(null, array(
+            'firstname' => array('type' => 'string', 'store' => 'yes'),
+            // default is store => no expected
+            'lastname' => array('type' => 'string'),
+            'hobbies' => array(
+                'type' => 'nested',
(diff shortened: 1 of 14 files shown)


ruflin:
@kukulich Because of the tests that are broken I actually prefer to keep this PR currently out of 3.0.0. It is easy to add a feature later but removing and changing BC is worse I think. As we plan to ship 3.1 as soon as this is fixed, I don't think it should be a big issue.

kukulich:
@ruflin I don't know if I understand you correctly. Do you plan to release 3.0.0 without `Filter\Nested`?

ruflin:
@kukulich Yes, I just did. The reason is that it is very easy to add code after a release but hard to remove it. The tests for nested didn't pass because some results were empty instead of results. So I prefer that people are aware of the issue instead of just using it and not getting results. I hope to ship either 3.0.1 or 3.1.0 in the next days with a fix.

test/lib/Elastica/Test/Filter/NestedFilterWithSetFilterTest.php
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * @group functional
+     */
+    public function testShouldReturnTheRightNumberOfResult()
+    {
+        $filter = new Nested();
+        $this->assertEquals(array('nested' => array()), $filter->toArray());
+        $query = new Terms();
+        $query->setTerms('hobby', array('guitar'));
```
kukulich: @ruflin From the documentation "Note that any fields referenced inside the query must use the complete path (fully qualified)."

This should work `$query->setTerms('hobbies.hobby', array('guitar'));`

ruflin: @kukulich that is great. Seems to work.

ruflin:
@ewgRa Seems like @kukulich just provided a very simple way to fix this PR. I suggest to merge it and release 3.0.1 with it. What do you think?

ewgRa:
Do it, I will merge it to my branch, I think it is not a problem for me.

ruflin:
@ewgRa Perfect. Lets wait for it to get green.

ruflin:
@kukulich Merged. Thanks a lot for the fast feedback for the fix.

kukulich:
You're welcome. I'm happy that `Filter\Nested` returned :)

PR #1023, opened 2016-01-05
Upgrade to Guzzle 6

jeskew:
Not sure if this change can be made without a major version bump. Since Guzzle is a suggestion and not a hard dependency, it might be allowable.

### DIFF
--- a/lib/Elastica/Transport/Guzzle.php
+++ b/lib/Elastica/Transport/Guzzle.php
@@ -10,7 +10,8 @@ use Elastica\Request;
 use Elastica\Response;
 use GuzzleHttp\Client;
 use GuzzleHttp\Exception\TransferException;
-use GuzzleHttp\Stream\Stream;
+use GuzzleHttp\Psr7;
+use GuzzleHttp\Psr7\Uri;
 
 /**
  * Elastica Guzzle Transport object.
@@ -29,7 +30,7 @@ class Guzzle extends AbstractTransport
     /**
      * Curl resource to reuse.
      *
-     * @var resource Guzzle resource to reuse
+     * @var Client Guzzle client to reuse
      */
     protected static $_guzzleClientConnection = null;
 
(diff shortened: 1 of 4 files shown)


ruflin:
Thanks for the addition. On the one hand I would really like to drop support for 5.4 as it is EOL (http://php.net/eol.php) but on the other hand I know that some larger projects struggle because we dropped support for 5.3 :-(

But as you mentioned, guzzle is an optional requirement. So it would be nice to instead drop the full build for 5.4, to only skip the guzzle tests in 5.4. Is that possible? Best would be probably to not even install guzzle for the 5.4 build. We had a similar hack for 5.3 in the past: https://github.com/ruflin/Elastica/blob/2.3.1/.travis.yml#L26

ruflin:
@jeskew A good timing to get this in could be with 3.1.0

jeskew:
@ruflin It looks like a download failed in the last Travis run. Could you restart it?

lib/Elastica/Transport/Guzzle.php
```diff
@@ -126,16 +107,54 @@ public function exec(Request $request, array $params)
     }
 
     /**
+     * @param Request $request
+     * @param Connection $connection
+     *
+     * @return Psr7\Request
+     */
+    protected function createPsr7Request(Request $request, Connection $connection)
```
ruflin: Could you use _createPsr7Request here to follow the Elastica naming convention?

ruflin:
LGTM. See my comment above and can you update the CHANGELOG.md file?

jeskew:
@ruflin I wasn't sure how to classify the update to the changelog. Would you consider this a backward compatibility break?

ruflin:
@jeskew I would classify it under BC because for the people using 5.4 it is a BC break.

ruflin:
@jeskew Let me know if you move the entry yourself or if I should do it for you? Also in case squashing is no problem for you, could you squash the commits?

jeskew:
@ruflin squashed! Thanks for your help on this.

ruflin:
@jeskew Merged. Thanks for the contribution.

lib/Elastica/Transport/Guzzle.php
```diff
@@ -156,7 +175,12 @@ protected function _getBaseUrl(Connection $connection)
         if (!empty($url)) {
             $baseUri = $url;
         } else {
-            $baseUri = $this->_scheme.'://'.$connection->getHost().':'.$connection->getPort().'/'.$connection->getPath();
+            $baseUri = (string) Uri::fromParts([
+                'scheme' => $this->_scheme,
+                'host' => $connection->getHost(),
+                'port' => $connection->getPort(),
+                'path' => ltrim('/', $connection->getPath()),
```
Tobion: this seems broken `ltrim('/', $connection->getPath())` makes no sense.

ruflin: @Tobion Can you share some more thoughts here? I assume the intention was that the Path always has / as a prefix and the host doesn't.

ruflin: I didn't realise that the params were turned around. Definitively doesn't make sense. 

Did you stumble over this because it broke something? If no, it seems to just work and we can remove it? Otherwise, we should provide a fix.

Tobion: I didn't try the guzzle transport yet. Was just looking at it out of interest. But I assume the `ltrim` should be removed.

ruflin: Agree. Should we open an issue or PR directly?

Issue #949, opened 2015-10-08
Simplify usage of credentials with elasticsearch


ruflin:
Currently to use credentials with elasticsearch, the following is needed:

```
$client = new \Elastica\Client(
  array(
    'host' => 'es.host.com',
    'port' => '9200',
    'headers' => array(
        'Authorization ' =>  'Basic ' . base64_encode('admin' . ':' . '1234')
    )
  )
);
```

This should be simplified through introducing username and password params in connections.

juuuuuu:
Maybe this issue could be closed.

ruflin:
@juuuuuu Yep, closing. Thanks for the ping.

PR #1030, opened 2016-01-11
Adding username and password params to connection

levijackson:
Adding username and password params to connection (https://github.com/ruflin/Elastica/issues/949)

I wanted to write functional or at the very least unit tests for Transport\HttpTest.php but I'm unsure the best way to do that. I figured
... a functional test would require a second Elasticsearch instance to test against.
... a unit test isn't possible without pulling the curl_ commands into a wrapper class. I didn't want to blow the scope of this up though.

### DIFF
--- a/lib/Elastica/Client.php
+++ b/lib/Elastica/Client.php
@@ -38,6 +38,8 @@ class Client
         'log' => false,
         'retryOnConflict' => 0,
         'bigintConversion' => false,
+        'username' => null,
+        'password' => null,
     );
 
     /**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable changes to this project will be documented in this file based on the
 ### Added
 
 ### Improvements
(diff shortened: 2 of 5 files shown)


lib/Elastica/Client.php
```diff
@@ -38,6 +38,8 @@ class Client
         'log' => false,
         'retryOnConflict' => 0,
         'bigintConversion' => false,
+        'username' => null,
```
ruflin: I was first thinking if we should better add this inside the connections. But that would mean setting it up for each connection. I think it can be expected that all elasticsearch nodes have the same users so I'm happy to have it hear.

ruflin:
Yeah, it would probably need either an nginx setup or an elasticsearch node with shield. I'm fine with the current tests as long as you tested it manually?

levijackson:
Ok. Yep, I tested it manually with a server that required credentials as well as one that does not require credentials.

ruflin:
Merged. Thanks.

PR #1031, opened 2016-01-11
Filters deprecated

ewgRa:
PR for #1001.

Review it carefully please.

Two things left: changelog and QueryBuilder DSL.

I haven't experience with QueryBuilder, so any help would be appreciated.

1. QueryBuilder/DSL/Filter - must be deprecated? If yes, in QueryBuilder constructor it must be not created, so, need make it lazy or something.

2. DSL::TYPE_FILTER - must be deprecated?

3. Version::getFilters, Version::filters, Version::supports (with filter type) - must be deprecated? Seems not, because there is support of old versions.

### DIFF
--- a/lib/Elastica/Aggregation/Filter.php
+++ b/lib/Elastica/Aggregation/Filter.php
@@ -3,6 +3,7 @@ namespace Elastica\Aggregation;
 
 use Elastica\Exception\InvalidException;
 use Elastica\Filter\AbstractFilter;
+use Elastica\Query\AbstractQuery;
 
 /**
  * Class Filter.
@@ -13,13 +14,19 @@ class Filter extends AbstractAggregation
 {
     /**
      * @param string         $name
-     * @param AbstractFilter $filter
+     * @param AbstractQuery $filter
      */
-    public function __construct($name, AbstractFilter $filter = null)
+    public function __construct($name, $filter = null)
     {
         parent::__construct($name);
 
         if ($filter !== null) {
+            if ($filter instanceof AbstractFilter) {
+                trigger_error('Deprecated: Elastica\Aggregation\Filter passing filter as AbstractFilter is deprecated. Pass instance of AbstractQuery instead.', E_USER_DEPRECATED);
+            } elseif (!($filter instanceof AbstractQuery)) {
+                throw new InvalidException('Filter must be instance of AbstractQuery');
+            }
+
             $this->setFilter($filter);
         }
     }
@@ -27,12 +34,18 @@ class Filter extends AbstractAggregation
     /**
      * Set the filter for this aggregation.
      *
-     * @param AbstractFilter $filter
+     * @param AbstractQuery $filter
      *
      * @return $this
      */
-    public function setFilter(AbstractFilter $filter)
+    public function setFilter($filter)
     {
+        if ($filter instanceof AbstractFilter) {
+            trigger_error('Deprecated: Elastica\Aggregation\Filter\setFilter() passing filter as AbstractFilter is deprecated. Pass instance of AbstractQuery instead.', E_USER_DEPRECATED);
+        } elseif (!($filter instanceof AbstractQuery)) {
+            throw new InvalidException('Filter must be instance of AbstractQuery');
+        }
+
         return $this->setParam('filter', $filter);
     }
 
--- a/lib/Elastica/Filter/AbstractFilter.php
+++ b/lib/Elastica/Filter/AbstractFilter.php
@@ -4,12 +4,16 @@ namespace Elastica\Filter;
 use Elastica\Exception\InvalidException;
 use Elastica\Param;
 
+trigger_error('Deprecated: Filters are deprecated. Use queries in filter context. See https://www.elastic.co/guide/en/elasticsearch/reference/2.0/query-dsl-filters.html', E_USER_DEPRECATED);
+
 /**
  * Abstract filter object. Should be extended by all filter types.
  *
  * @author Nicolas Ruflin <spam@ruflin.com>
  *
  * @link https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-filters.html
+ *
+ * @deprecated Filters are deprecated. Use queries in filter context. See https://www.elastic.co/guide/en/elasticsearch/reference/2.0/query-dsl-filters.html
  */
 abstract class AbstractFilter extends Param
 {
--- a/lib/Elastica/Query/FunctionScore.php
+++ b/lib/Elastica/Query/FunctionScore.php
@@ -1,7 +1,9 @@
 <?php
 namespace Elastica\Query;
 
+use Elastica\Exception\DeprecatedException;
 use Elastica\Script\AbstractScript;
+use Elastica\Exception\InvalidException;
 use Elastica\Filter\AbstractFilter;
 
 /**
@@ -56,12 +58,13 @@ class FunctionScore extends AbstractQuery
 
     /**
      * @param AbstractFilter $filter
+     * @deprecated Elastica\Query\FunctionScore::setFilter is deprecated. Use setQuery instead
      *
      * @return $this
      */
     public function setFilter(AbstractFilter $filter)
     {
-        return $this->setParam('filter', $filter);
+        throw new DeprecatedException('Deprecated: Elastica\Query\FunctionScore::setFilter is deprecated. Use setQuery instead.');
     }
 
     /**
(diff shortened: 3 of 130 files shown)


lib/Elastica/Filter/AbstractFilter.php
```diff
@@ -4,12 +4,16 @@
...
 /**
  * Abstract filter object. Should be extended by all filter types.
  *
  * @author Nicolas Ruflin <spam@ruflin.com>
  *
  * @link https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-filters.html
+ *
+ * @deprecated Filters are deprecated. Use queries in filter context. See https://www.elastic.co/guide/en/elasticsearch/reference/2.0/query-dsl-filters.html
  */
 abstract class AbstractFilter extends Param
```
ruflin: Sholudn't AbstractFilter extend AbstractQuery?

lib/Elastica/Aggregation/Filter.php
```diff
@@ -13,13 +14,19 @@ class Filter extends AbstractAggregation
...
-     * @param AbstractFilter $filter
+     * @param AbstractQuery $filter
      */
-    public function __construct($name, AbstractFilter $filter = null)
+    public function __construct($name, $filter = null)
     {
         parent::__construct($name);
 
         if ($filter !== null) {
+            if ($filter instanceof AbstractFilter) {
```
ruflin: Could we move this to a function in Param for example? This would simplify the code. You could still pass the first part of the message to the function to have it noting what is deprecated.

lib/Elastica/Filter/AbstractFilter.php
ruflin: My initial thought was, that one extends the other which would prevent the check function above as we could still have the type specified in the function. Still thinking if this could make sense? That would remove the throw InvalidException part.

lib/Elastica/Query/FunctionScore.php
```diff
@@ -69,19 +76,27 @@ public function setFilter(AbstractFilter $filter)
      *
      * @param string         $functionType   valid values are DECAY_* constants and script_score
      * @param array|float    $functionParams the body of the function. See documentation for proper syntax.
-     * @param AbstractFilter $filter         optional filter to apply to the function
+     * @param AbstractQuery  $filter         optional filter to apply to the function
      * @param float          $weight         function weight
      *
      * @return $this
      */
-    public function addFunction($functionType, $functionParams, AbstractFilter $filter = null, $weight = null)
```
ruflin: Should we rename the param the $query instead f $filter to make it more obvious? This would apply to all changes.

lib/Elastica/Aggregation/Filter.php
ruflin: The other advantage is that we don't have to check for null every time. The deprecated message will be shown anyways if someone uses a filter, as somehow the class has to be initiated.

lib/Elastica/QueryBuilder.php
```diff
@@ -33,6 +33,7 @@ public function __construct(Version $version = null)
         $this->_version = $version ?: new Version\Latest();
 
         $this->addDSL(new DSL\Query());
+        // FIXME XXX: make it lazy
```
ruflin: If we don't make it "lazy" it means every time the query build is called, we get the deprecated error, right?

ruflin:
Thanks a lot for your work. I added some notes to the code but would like to summarise it here again. Overall the changes look very good. Here some changes I would suggest:

* DSL: I would ignore the DSL as far as possible at the moment. The main issue is that I assume in case we initiate the filter inside, it will always trigger an error even if filter is not used?
* I suggest we use `AbstractFilter extends AbstractQuery` I'm aware that this is not 100% correct as I assume query has a little bit more functionality, but it will heavily simplify the code as we can define the type passed to the functions `AbstractQuery` which will remove all checks and deprecated messages. In case someone uses an AbstractFilter he already gets the deprecated message because of the object
* For clarity we should probably rename all function params called $filter to $query when an AbstractQuery is passed
* I didn't check the classes copied from filter to queries as I assume it is just a copy of the filter
* I didn't check all test changes in detail

ewgRa:
Lets focus on this first:

> I suggest we use AbstractFilter extends AbstractQuery

I think on it when I start working, have similar idea, and decide that it will be broke BC in many cases.

For example in Elastica/Query.php there is a code:
```
            case $query instanceof AbstractQuery:
                return new self($query);
            case $query instanceof AbstractFilter:
                $newQuery = new self();
                $newQuery->setPostFilter($query);

                return $newQuery;
```

It will be broken in this case. I think if such extends will be used - it will be a mess for developers and this will be bad BC layer. Some developers also can have similar checks.

ruflin:
Good point, wasn't thinking of this one. I see that we could break some code here, but shouldn't we remove all usage of AbstractFilter in our code anyways? How would the above look like if there were no filters anymore?

ewgRa:
In this PR Elastica use AbstractFilter only as support BC layer, to trigger errors and some legacy tests.
After BC layer will be removed - we will haven't any AbstractFilter in our code. This BC layer - kind of "last call" for developers and chance to switch to new Elastica version without big pain.

Main idea was provide BC layer that say how to drop filter usage and warn about its usage. Now developer still can use Filters when they will switch to new version of Elastica and work with deprecation errors. They will have time to find all usage of deprecated stuff before it will be completely removed and change it step by step, make estimation how hard it will be, and can work with it without blocking development cycle. It makes switch to newer Elastica version not so painful.

In next major release (4.0?) filter BC layer will be removed.

Yes, there is question about bug fixing and new features in deprecated code. I think bug fixing must be provided even for deprecated code. New features - only for non deprecated, this will push developers switch to non deprecated functionality faster.

ruflin:
@ewgRa I think I fully agree on what you stated above. I was referring to this line here (that you posted above): https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Query.php#L74 That seems to be a "usage" of AbstractFilter in our code. Shouldn't we remove this and convert it to filters?

ruflin:
Based on the discussion in https://github.com/ruflin/Elastica/pull/724 I suggest to "ignore" the query part for the moment and wait for the PR. But we should still progress with this PR. @ewgRa See my comment above.

ewgRa:
> DSL: I would ignore the DSL as far as possible at the moment.

I revert back my changes for QueryBuilder. Now there is just small changes that follow changes in filters to be not broken.

> That seems to be a "usage" of AbstractFilter in our code. Shouldn't we remove this and convert it to filters?

It is part of BC layer. So, it must be there and than will be removed as part of BC layer. If we remove it now - developers have BC break, because instead of set post filter it will be set query. We must warn them, this is why AbstractFilter still used there.

> For clarity we should probably rename all function params called $filter to $query when an AbstractQuery is passed

I disagree. If elasticsearch have 'filter', we must have setFilter method. If we change setFilter($filter) to setFilter($query) - it will be like setSomething(int $int). Arguments must be named not by type, but by what they represent. If you still want rename it, at least lets ask opinion another developers to reduce subjective thinking in this problem. 

> I didn't check the classes copied from filter to queries as I assume it is just a copy of the filter

Mostly. Just for Indices query have setQuery instead of setFilter.

> I didn't check all test changes in detail

There is important change in Base::setUp, I enable show deprecation messages for all tests, and I've added some new methods for testing BC layer.

Is I miss something? Lets find agreements on general level and than I can switch to notes for commit.

ruflin:
@ewgRa I agree to all your points above. I think we are good :-)

Lumbendil:
Since method signatures will change, this should at least bump the minor version (reflection results would change) imho. I didn't see it mentioned, so I thought I'd say it.

ruflin:
@Lumbendil Definitively. Idea would be 3.1

lib/Elastica/Aggregation/Filter.php
ewgRa: Usually we need check filter on null if it can be null, like in this case, we set filter if it is only not null. So, we can't move it to function.

About code simplify and extract it to Param, I don't see big advantage, because function call will be with messages, and it save just 2-3 lines. And some places can have filter as array, that require another function with another checks. 

But negative effect here that exceptions and error will be triggered in Param and this will be not a last call at stack, so developer will have not a right place - where they need change something and they must check previous call stack to find real reason.

For me save 2-3 lines in temporary code will be compensated by wrong place in call stack.

If you still wish, I can do this, no problem, just confirm your wish.

lib/Elastica/Filter/AbstractFilter.php
ewgRa: As we agree in comments, it shouldn't.

lib/Elastica/Query/FunctionScore.php
ewgRa: As we agree in comments, we shouldn't.

ruflin:
@ewgRa LGTM. Merging is next step. Do you feel comfortable to squash the commits or should we go as with the current commits?

ewgRa:
I will squash it and make changes in changelog

ruflin:
:+1:

ewgRa:
Also interesting that Symfony team now not ask about squash.

ewgRa:
sorry, accidentally press button

ruflin:
@ewgRa Not sure if you meant that Symfony asks or doesn't? In any case, I think rebasing is great but also brings quite a few risks and can lead to a mess. That is why I normally ask if people feel comfortable with it, means did it in a past.

ewgRa:
No, no, all is fine with squash.

I just think that Symfony team also ask about squash, but later they change it and says: "Moderators earlier asked you to "squash" your commits. This means you will convert many commits to one commit. This is no longer necessary today, because Symfony project uses a proprietary tool which automatically squashes all commits before merging.".

I think that this is something that can be used in Elastica, but seems it is custom tool, far away from github.

ruflin:
@ewgRa Interesting, didn't know about it. Would have to have a closer look. But it sounds like a little bit too much magic for me :-)

ewgRa:
Done

ruflin:
@ewgRa Merged. Thanks for all the work.

Issue #960, opened 2015-10-18
get result as document


yehosef:
I would like to query for some documents, make changes to them, and then save them again.  Normally when getting a single doc by primary key, I get an elastica\document and make the changes via that.

I would also like to use this approach when retrieving a Elastica\ResultSet or Result.  I would like to either retrieve this as a Document or convert it to one (eg, including the version info).  Is this possible?  And if not, could there be a way to do it?

ruflin:
Currently this is not directly supported by Elastica. It was requested several times in the past but nobody implemented it so far. It shouldn't be too hard to implement it on top of ResultSet or Result. It would required to create document objects based on the retrieved values.

yehosef:
ok - .. you can leave open for a little and I'll see if I can do it.  I
have a working implementation.  The issue is that probably it needs to
ensure that the fields are not specified or else it will give partial
documents and saving them would not give the desired results.

ruflin:
:+1: Let me know if you need some help.

yehosef:
Thanks!  I'm building an ORM like structure using Elastica - this is the part in my constructor where I handle Elastic\Result

     elseif ($data instanceof Elastica\Result)
        {
            /** @var Elastica\Result $data */

            $doc = new Elastica\Document($data->getId(), $data->getData());

            $version = $data->getVersion();
            $version && $doc->setVersion($version);

            $this->doc = $doc;
            $this->doc->setDocAsUpsert(true);
        }

a few things I'm not sure about:
1) do I need the version check, or will a Result always have a version?
2) I'm not sure if I need the DocAsUpsert (I not even really sure what this does.. :)  
3) is there anything else I need to be passing into the new Document from the Result?


Like I said, I assume there should be some `$data->getFields()` check or something - if I specified the fields, I assume it would fail because I might not have all the data.

Any other ideas based on your experience with ES?

ps - thank you so much for Elastica - it's really amazing!

ruflin:
* As far as I know, version is only returned if version is set to true in the request. So you need to build in a check
* DocAsUpsert means the values in the document are used to "partially" update a document. Assuming you get "all" data as a response, you don't need this. There are some additional cases with script, but I think this would complicate the code too much (keep it simple and cover 80% of the use cases).
* You should also add index and type to the document

The line `$version && $doc->setVersion($version);` is quite hard to read. I would split it up to an if clause.

[...]

PR #1033, opened 2016-01-13
For issue #960

intech:
Fix https://github.com/ruflin/Elastica/issues/960

### DIFF
--- a/lib/Elastica/Result.php
+++ b/lib/Elastica/Result.php
@@ -197,7 +197,9 @@ class Result
         $doc = new \Elastica\Document();
         $doc->setData($this->getSource());
         $hit = $this->getHit();
-        unset($hit['_source']);
+        $this->hasParam('_source') unset($hit['_source']);
+        $this->hasParam('_explanation') unset($hit['_explanation']);
+        $this->hasParam('highlight') unset($hit['highlight']);
         $doc->setParams($hit);
         return $doc;
     }
--- a/lib/Elastica/Result.php
+++ b/lib/Elastica/Result.php
@@ -200,6 +200,7 @@ class Result
         if($this->hasParam('_source')) unset($hit['_source']);
         if($this->hasParam('_explanation')) unset($hit['_explanation']);
         if($this->hasParam('highlight')) unset($hit['highlight']);
+        if($this->hasParam('_score')) unset($hit['_score']);
         $doc->setParams($hit);
         return $doc;
     }
--- a/lib/Elastica/Result.php
+++ b/lib/Elastica/Result.php
@@ -186,6 +186,21 @@ class Result
     {
         return $this->getParam('_explanation');
     }
+    
+    /**
+     * Returns Document
+     * 
+     * @return \Elastica\Document
+     */
+    public function getDocument()
+    {
+        $doc = new \Elastica\Document();
+        $doc->setData($this->getSource());
+        $hit = $this->getHit();
+        unset($hit['_source']);
+        $doc->setParams($hit);
+        return $doc;
+    }
 
     /**
      * Magic function to directly access keys inside the result.
--- a/lib/Elastica/Result.php
+++ b/lib/Elastica/Result.php
@@ -197,9 +197,9 @@ class Result
         $doc = new \Elastica\Document();
         $doc->setData($this->getSource());
         $hit = $this->getHit();
-        $this->hasParam('_source') unset($hit['_source']);
-        $this->hasParam('_explanation') unset($hit['_explanation']);
-        $this->hasParam('highlight') unset($hit['highlight']);
+        if($this->hasParam('_source')) unset($hit['_source']);
+        if($this->hasParam('_explanation')) unset($hit['_explanation']);
+        if($this->hasParam('highlight')) unset($hit['highlight']);
         $doc->setParams($hit);
         return $doc;
     }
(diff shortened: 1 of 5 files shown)


ruflin:
@intech Thanks a lot for the contribution. Could you update the CHANGELOG.md file and add some basic tests?

intech:
@ruflin okey! No problems.

intech:
@ruflin Done!

ruflin:
The implementation looks good to me. Just to be one the same page: As far as I understand you didn't implement the full version we discussed here: https://github.com/ruflin/Elastica/issues/960 For example type and index or not set on the document, right? The current implementation is fine for me as it creates the base for future improvements. Just to clarify that this will not directly solve what @yehosef initially planned.

intech:
@ruflin all work perfect!
```php
$doc->set('name', 'value');
$index->updateDocument($doc);
```

webdevsHub:
`_id`, `_type`, `_index` and `_version` works with `$doc->setParams($result->getHit())`.

There might be strange behavior if `highlight` or `_explanation` is set and moved to document. You can constrcut a result with bascially everything, `new \Elastica\Result(['some' => 1, 'wired' => 2, 'keys' => 3]);`.

lib/Elastica/Result.php
```diff
@@ -186,6 +186,21 @@ public function getExplanation()
...
+     * Returns Document
+     * 
+     * @return \Elastica\Document
+     */
+    public function getDocument()
+    {
+        $doc = new \Elastica\Document();
+        $doc->setData($this->getSource());
+        $hit = $this->getHit();
+        unset($hit['_source']);
```
webdevsHub: `_source` might not be set in result, so this line might throw fatal error

webdevsHub:
An exception should be thrown if `$result->hasFields() === true` because you can not decide if `_fields` contains a full document

intech:
@webdevsHub you have more experience than me. How to do it better?
```php
$doc->setData($this->getSource());
```
replace to `getData()` ?

```php
unset($hit['_source']);
```
replace to
```php
if($this->hasParam('_source')) unset($hit['_source']);
```

test/lib/Elastica/Test/ResultSetTest.php
```diff
@@ -58,6 +58,35 @@ public function testArrayAccess()
...
+        $this->assertInstanceOf('Elastica\ResultSet', $resultSet);
+        
+        $documents = $resultSet->getDocuments();
+
+        $this->assertInternalType('array', $documents);
+        $this->assertEquals(3, count($documents));        
+        $this->assertInstanceOf('Elastica\Document', $documents[0]);
+        $this->assertInstanceOf('Elastica\Document', $documents[1]);
+        $this->assertInstanceOf('Elastica\Document', $documents[2]);
+        $this->assertFalse(isset($documents[3]));
```
ewgRa: Make no sense, since there is assertEquals - 3 an than all three you check in next lines.

Better here to test document content, like name => 'elastica search' is really in document.

intech: :+1:

lib/Elastica/Result.php
ewgRa: unset not raise notice if there is key not present in array:
php -r "\$a = array(); unset(\$a['a']);" - work without notice on 5.5.9 and I think on another php versions same.

test/lib/Elastica/Test/ResultSetTest.php
intech: Done!

lib/Elastica/Result.php
intech: http://php.net/manual/en/function.unset.php
> Note:
When using unset() on inaccessible object properties, the __unset() overloading method will be called, if declared.

ruflin:
Looking at the implementation I have the gut feeling that there are quite a few edge cases which will break this implementation. But it is probably quite hard to cover all of them. To move forward I see two options:

* Merge it as it is and wait for users to come up with the edge cases (by creating and issue)
* Spend more time on the PR and improve the test set, to cover more edge cases.

I'm tempted to go with approach one but would like to get the feedback from @ewgRa and @webdevsHub here.

intech:
@ruflin I propose to add to the whitelist \Elastica\Document with the list of params that can be there and solve this problem. And then act on the first point.

intech:
https://github.com/elastic/elasticsearch-php/blob/master/src/Elasticsearch/Endpoints/Source/Get.php#L55

ewgRa:
For me this PR is ok, it will work. Maybe I will implement it in different way, like DocumentBuilder::buildFromResult(Result), DocumentCollection::buildFromResult(ResultSet), but it is subjective. I just don't think add to Result and ResultSet more knowledge about Document.

I just wondering that there is #1017 that when will be implemented will provide same functionality.

webdevsHub:
I like the whitelist approach, but whitelists always cause problems because it needs constant maintenance. Another edge case could be on disabled source field. Would be nice to have integration tests for known edge cases (insert, retrieve and update a doc with edge case mapping/structure)

ruflin:
Lets go forward with the current state and revisit it combined with https://github.com/ruflin/Elastica/pull/1017 I agree that some great new things are ahead of us here.

@intech Could you squash your commits?

intech:
@ruflin what should be done? do not understand :(

ruflin:
@intech If you are not familiar with git rebase and squashing it is probably better to merge it as it is. It is easy to get it wrong the first time :-)

ruflin:
@intech Merged. Thanks.

intech:
@ruflin thanks! I do not understand because of his poor English, what I need to do. I was afraid to do wrong. :(

ruflin:
@intech no worries. Rebase / Squashing can be useful to have a clean history. Here some related search results in case you are interested: https://www.google.ch/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=squash%20commits%20rebase

PR #1037, opened 2016-01-17
Fix Index->deleteByQuery array lazy loading

raphaelMalie:
This method was broken since the array lazy loading push.

### DIFF
--- a/lib/Elastica/Index.php
+++ b/lib/Elastica/Index.php
@@ -159,7 +159,7 @@ class Index implements SearchableInterface
         }
         $query = Query::create($query);
 
-        return $this->request('_query', Request::DELETE, array('query' => $query->getQuery()->toArray()), $options);
+        return $this->request('_query', Request::DELETE, array('query' => $query->toArray()), $options);
     }
 
     /**


ruflin:
@raphaelMalie Thanks for the fix. Can you please update the CHANGELOG.md? Can you add a small tests that would fail before and is not successful to make sure we don't break it again?

xjewer:
same bug in this line https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Type.php#L490

ruflin:
@raphaelMalie Any updates here?

xjewer:
>This method was broken since the array lazy loading push.

raphaelMalie:
Hello, sorry I never wrote tests before and I have no time to learn it now :/ You can close this PR and apply the fix by yourself, it would be better if possible. Regards, Raphaël.

xjewer:
>This method was broken since the array lazy loading push.

it can happen, if create Query from array

```php
Query::create(
    [
        'query' => [
            'ids' => [
                'values' => [1],
            ],
        ],
    ]
);
```

so, array set through `$this->setRawQuery($query);` and when execute `$Query->getQuery()` we get array instead Query object

ruflin:
@raphaelMalie Thanks for the heads up. Will try to take your commit and add tests (or someone else who has the time :-) ).

@xjewer We should add a test and fix for that (if possible).

ruflin:
This was now fixed in https://github.com/ruflin/Elastica/pull/1049

PR #1039, opened 2016-01-22
Fix isset result for null value when key exist

bourvill:
Hi,
I think the interpretation of __isset is wrong.
Indeed a key can be null but in array.

Before the fix, the value given was "false" when in fact it is null and exist

Can imagine merge, and tag this fix in 3.0.2 ?

Thanx

### DIFF
--- a/lib/Elastica/Result.php
+++ b/lib/Elastica/Result.php
@@ -214,6 +214,6 @@ class Result
     {
         $source = $this->getData();
 
-        return array_key_exists($key, $source) && $source[$key] !== null;
+        return array_key_exists($key, $source);
     }
 }


ewgRa:
Thanks for PR.

I think current interpretation is right.
When developer have isset($result["something"]) he expect that it will work as isset documentation says. With your PR will be unexpected result.

What you want - it is just array_key_exists, it can be done by introduce another method (hasSourceParam for example) on my mind.

Are you agree?

@ruflin, I wondering, why there array_key_exists and null check, when it can be done by isset directly? Perfomance?

yobrx:
+1 with bourvill.

It's very laborious to check if variable is null in my project (template Twig, for example)

bourvill:
Hello @ewgRa,

In my case some documents can return a null value in some fields.

The previous test gives as result that fields doesn't exists, but the field exists in the document and return a null value.

The problem is that we test that field exists and the value is not null as it may be null.

I agree with the difference of isset function, but on class "Result" I think we should test simply existance of field, not the value.

ewgRa:
@bourvill 

>  but on class "Result" I think we should test simply existance of field, not the value.

Why? To have unexpected magic for a lot of developers who knows that isset return false if argument is null? They will be don't know anything about what inside of result, but will expect documented behavior from isset.

ruflin:
@ewgRa @bourvill Thanks for the discussion. I had to read up on isset myself to see what it actually does in the above cases: http://php.net/manual/en/function.isset.php 

I think using isset() or empty() should always return what is somehow expected, so I would agree with @ewgRa that there shouldn't be any magic on the Elastica side. Based on the example of @yobrx we also see, that the above could be a breaking change.

On the one hand I'm not sure if it is the job of Elastica to make template handling easier, but at the same time the library is here to simplify engineers life. I don't think we should change the behaviour of isset but I'm happy to a function like `hasSourceParam` or perhaps a better name that does what you would expect from isset.

@ewgRa Not sure why array_key_exists is used here, but I'm quite sure it isn't for performance reasons.

Issue #719, opened 2014-11-08
$fullResult parameter for Elastica\Type::count() 


akorthaus:
Hi!

Elastica\Search::count() has a 2nd parameter " $fullResult " to return a ResultSet object instead of the count. This is useful if you want to query aggregations only which should be cached by shard query cache:
http://www.elasticsearch.org/guide/en/elasticsearch/reference/1.4/index-modules-shard-query-cache.html

Is there a reason, why this is not implemented for Elastica\Type::count()?

Thank you!

best
Andreas

PR #1038, opened 2016-01-17
Options for count

intech:
Options do not work! You can not work properly without [_routing](https://www.elastic.co/blog/customizing-your-document-routing).

PR #1040, opened 2016-01-26
$fullResult parameter for Elastica\Type::count()

vinusebastian:
Addresses https://github.com/ruflin/Elastica/issues/719

### DIFF
--- a/lib/Elastica/Search.php
+++ b/lib/Elastica/Search.php
@@ -476,12 +476,11 @@ class Search
 
         $query = $this->getQuery();
         $path = $this->getPath();
-
         $response = $this->getClient()->request(
             $path,
             Request::GET,
             $query->toArray(),
-            array(self::OPTION_SEARCH_TYPE => self::OPTION_SEARCH_TYPE_COUNT)
+            array(self::OPTION_SEARCH_TYPE => self::OPTION_SEARCH_TYPE_QUERY_THEN_FETCH)
         );
         $resultSet = ResultSet::create($response, $query);
 
--- a/lib/Elastica/Type.php
+++ b/lib/Elastica/Type.php
@@ -358,17 +358,17 @@ class Type implements SearchableInterface
     /**
      * Count docs by query.
      *
-     * @param string|array|\Elastica\Query $query Array with all query data inside or a Elastica\Query object
-     *
-     * @return int number of documents matching the query
+     * @param string|array|\Elastica\Query $query       Array with all query data inside or a Elastica\Query object
+     * @param ResultSet                    $fullResult (default = false) By default only the total hit count is returned. If set to true, the full ResultSet including aggregations is returned.
+     * @return int number of documents matching the query|ResultSet
      *
      * @see \Elastica\SearchableInterface::count
      */
-    public function count($query = '')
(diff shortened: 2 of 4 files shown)


lib/Elastica/Search.php
```diff
@@ -476,12 +476,11 @@ public function count($query = '', $fullResult = false)
 
         $query = $this->getQuery();
         $path = $this->getPath();
-
         $response = $this->getClient()->request(
             $path,
             Request::GET,
             $query->toArray(),
-            array(self::OPTION_SEARCH_TYPE => self::OPTION_SEARCH_TYPE_COUNT)
+            array(self::OPTION_SEARCH_TYPE => self::OPTION_SEARCH_TYPE_QUERY_THEN_FETCH)
```
ruflin: Shouldn't this be dependent on which param is passed to count?

ruflin:
@vinusebastian Thanks for the PR. Looks good to me. Just one small question above. Can you also update the CHANGELOG.md?

ewgRa:
maybe better to create countResult?

also is OPTION_SEARCH_TYPE_QUERY_THEN_FETCH change not a break BC?

lib/Elastica/Search.php
vinusebastian: Yes, but Elasticsearch documentation says that `search type` of `count` is deprecated?  And that `count` doesn't have any advantage over `query then fetch` (provided size of query is 0)

https://www.elastic.co/guide/en/elasticsearch/reference/2.x/search-request-search-type.html

ruflin: Wasn't aware of that. LGTM.

vinusebastian:
@ewgRa Didn't follow what you said. Can you explain?

ruflin:
I think this PR also overlaps with: https://github.com/ruflin/Elastica/pull/1038

lib/Elastica/Type.php
```diff
@@ -358,17 +358,18 @@ public function search($query = '', $options = null)
...
-     * @return int number of documents matching the query
+     * @param string|array|\Elastica\Query $query       Array with all query data inside or a Elastica\Query object
+     * @param ResultSet                    $fullResult (default = false) By default only the total hit count is returned. If set to true, the full ResultSet including aggregations is returned.
+     * 
+     * @return int number of documents matching the query|ResultSet
      *
      * @see \Elastica\SearchableInterface::count
      */
-    public function count($query = '')
+    public function count($query = '', $fullResult = false)
```
ruflin: It looks like the approach here with an options array could be more flexible: https://github.com/ruflin/Elastica/pull/1038/files#diff-1d7207a5d6e8ecfaa5d6386b7e0edb99R368 Can you have a look at this together with @intech ?

ewgRa:
@vinusebastian never mind, seems I was wrong

lib/Elastica/Type.php
vinusebastian: Fair enough. Closing the PR

ruflin: @vinusebastian Idea was not necessarly to close this one, but have a chat with @intech on the best approach :-)

PR #1044, opened 2016-02-08
check aggregation parameters early

### LINKED ISSUE #1043: Add sanity checks for aggregations
I'm new to elastic search, and I'm playing with aggregations at the moment. The first bug that hits me quite hard is when trying to use `Cardinality` : I get a big error message telling me my query is invalid. When I realized that the name of the aggregation and the field are different things, I understand my error and use `setField`, but keep wondering why not simply have such a constructor for `Cardinality` (or event `AbstractSimpleAggregation` or higher) : `public function __construct($name, $field)`.

After reading the documentation, I see that aggregations can also be generated with scripts. Maybe a sanity check could be done at some point to make sure that one at least is provided?

greg0ire:
Instead of querying ES with something invalid, throw a meaningful
exception.
Fixes #1043

### DIFF
--- a/lib/Elastica/Aggregation/AbstractSimpleAggregation.php
+++ b/lib/Elastica/Aggregation/AbstractSimpleAggregation.php
@@ -1,6 +1,8 @@
 <?php
 namespace Elastica\Aggregation;
 
+use Elastica\Exception\InvalidException;
+
 abstract class AbstractSimpleAggregation extends AbstractAggregation
 {
     /**
@@ -32,6 +34,11 @@ abstract class AbstractSimpleAggregation extends AbstractAggregation
      */
     public function toArray()
     {
+        if (!$this->hasParam('field') && !$this->hasParam('script')) {
+            throw new InvalidException(
+                'Either the field param or the script param should be set'
+            );
+        }
         $array = parent::toArray();
 
         $baseName = $this->_getBaseName();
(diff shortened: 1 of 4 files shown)


lib/Elastica/Aggregation/AbstractSimpleAggregation.php
```diff
@@ -32,6 +32,11 @@ public function setScript($script)
      */
     public function toArray()
     {
+        if (!$this->hasParam('field') && !$this->hasParam('script')) {
+            throw new \LogicException(
```
ewgRa: In Elastica there is no usage of LogicException, I think better to change it to \Elastica\Exception\InvalidException to be on one wave with similar checks.

ewgRa:
Can you please also add test that exception throws when no script or field set?

And update changelog

greg0ire:
@ewgRa done!

test/lib/Elastica/Test/Aggregation/AbstractSimpleAggregationTest.php
```diff
@@ -0,0 +1,26 @@
...
+
+use Elastica\Aggregation\AbstractSimpleAggregation;
+use Elastica\Exception\InvalidException;
+
+class AbstractSimpleAggregationTest extends BaseAggregationTest
+{
+    public function setUp()
+    {
+        $this->aggregation = $this->getMockForAbstractClass(
+            AbstractSimpleAggregation::class,
```
ewgRa: "::class" introduced in PHP 5.5, in composer.json you can find information that Elastica support 5.4, can you fix it please?

greg0ire: Woops, thought it was 5.3, my bad. Fixed.

ewgRa:
thanks @greg0ire 

@ruflin fyi

ruflin:
Thanks @ewgRa and @greg0ire

Issue #1046, opened 2016-02-10
URL cannot include query-string parameters


olchick:
I'm trying to use the new Amazon Elasticsearch service however, I'm getting the following error when trying to use update document:

`When Content-Type:application/x-www-form-urlencoded, URL cannot include query-string parameters (after '?'): '/users/user/1/_update?version=2&retry_on_conflict=0'`

Also I found this post about [Amazon Elasticsearch Service Limits](http://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/aes-limits.html)

So it seems the Elasticsearch service doesn't like to receive query-string parameters.

I saw code in updateDocument function and there is always set $options even if I don't want use their. 

I use Elastica v.2.3.1

ruflin:
Passing this params per query is default in elasticsarch as far as I know: https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update.html#_parameters_3 Does Amazon have any recommendations what to do in these cases? This seems more to be an issue with the implementation of elasticsearch in AWS then Elastica. 

The improvement that could be made probably on the Elastica side is trying to not set any default params if none set here: https://github.com/ruflin/Elastica/blob/f35835e8308ec9e0b62842155603c32118a2ffc6/lib/Elastica/Client.php#L337

PR #1047, opened 2016-02-14
fix-update-document-options-not-set

olchick:
Comes from [1046](https://github.com/ruflin/Elastica/issues/1046)

Improvement for updateDocument method: not set option 'retry_on_conflict' if it is empty in Client's config.

### DIFF
--- a/lib/Elastica/Client.php
+++ b/lib/Elastica/Client.php
@@ -380,8 +380,9 @@ class Client
         }
 
         if (!isset($options['retry_on_conflict'])) {
-            $retryOnConflict = $this->getConfig('retryOnConflict');
-            $options['retry_on_conflict'] = $retryOnConflict;
+            if ($retryOnConflict = $this->getConfig('retryOnConflict')) {
+                $options['retry_on_conflict'] = $retryOnConflict;
+            }
         }
 
         $response = $this->request($path, Request::POST, $requestData, $options);


ewgRa:
@olchick For me it seems like broke BC.

For now Elastica always sends retry_on_conflict=0 by default. With your PR it will be taken from elasticsearch configuration and it can be not 0, for such developers it will be broke BC.

More right make something like "unsetConfig('retry_on_conflict')" or "setConfig('retry_on_conflict', null)" and than client can omit it when perform request.

Also what with 'version' field from original issue? I think if Amazon care about "retry_on_conflict" it will be also raise errors for "version", and your PR solve only retry_on_conflict.

Tobion:
@ewgRa retry_on_conflict=0 is the default in ES. How is that possible to change that?

ewgRa:
@Tobion https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update.html#_parameters_2

retry_on_conflict there

Or I misunderstand your question?

Tobion:
You said there is a configuration which can be changed and due that not sending the default value would be a BC break for those people.

> With your PR it will be taken from elasticsearch configuration and it can be not 0, for such developers it will be broke BC.

So what configuration are you talking about?

ewgRa:
@Tobion you are right. By default it is "the update will fail with a version conflict exception", if it is omitted, also like "default", and all this work as "0".

So, there is will be no BC break.

Tobion:
Good, then I think this is good to be merged.

Tobion:
Added changelog

PR #1048, opened 2016-02-17
fix phpdoc getQuery, it returns object instead array

xjewer:
This behavior was changed in #916


xjewer:
also, maybe it affected #1037

ruflin:
@xjewer Thanks for the fix. What you mean by affected #1037 ?

PR #1058, opened 2016-03-13
Lazy load status data

chuangbo:
In an elasticsearch instance with large number of indices, getting `_stats` is a huge expense, which `Elastica\Status` will load it for [every instance] (https://github.com/ruflin/Elastica/blob/e2daadc27ba263ab5b33c4afac10b1144078cd5d/lib/Elastica/Status.php#L44)

But some methods of `Elastica\Status` don't use `Status->_data`, e.g. `getIndicesWithAlias`. So it's possible to lazy load `_stats` data at when it's needed, to speed up stats independent function calls.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ All notable changes to this project will be documented in this file based on the
   `field` or `script` param.
 - `Elastica\Index->deleteByQuery($query, $options)` $query param can be a query `array` again
 - `Elastica\Query\MoreLikeThis->toArray()` now supports providing a non-indexed document as an input to perform the comparison.
+- `Elastica\Status` will lazy load the `_stats` at when it is needed.
 
 ### Deprecated
 
--- a/lib/Elastica/Status.php
+++ b/lib/Elastica/Status.php
@@ -24,7 +24,7 @@ class Status
      *
      * @var array Data
      */
-    protected $_data = array();
+    protected $_data = null;
 
     /**
      * Client object.
@@ -41,7 +41,6 @@ class Status
     public function __construct(Client $client)
     {
         $this->_client = $client;
-        $this->refresh();
     }
 
     /**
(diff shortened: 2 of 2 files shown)


ruflin:
It seems like this change somehow breaks the tests. Can you have a look?

Please also update the CHANGELOG.md file.

chuangbo:
It seems like the reason of tests failed is docker building error.
https://circleci.com/gh/ruflin/Elastica/7#tests

Can we try to re-run the tests on CircleCI and travis-ci? I don't have the re-run button here.

ruflin:
Please ignore CircleCI. I just restarted the tests for Travis CI.

ruflin:
@chuangbo Please ignore CircleCI. Here is the error on Travis which is in Status.php, so I assume it is related: https://travis-ci.org/ruflin/Elastica/jobs/116031553#L2048

chuangbo:
Thank you. The tests have been passed.

Actually this pull request looks more like a _hack_ to me, I think would it be better (api design) to move the `_stats` independent methods to somewhere else, or can we turn them to `static` function?

ruflin:
I agree it is currently far from clean, as we mix stats and status (which was not the same in the past). This came with the upgrade to elasticsearch 2.0 when we tried to keep it as BC compatible as possible. I'm good with this temporary hack for the moment but agree we should rethink the API here. I would even say Status should probably disappear as it does not exist anymore on the elasticsearch side.

@chuangbo Do you want to open a Github Issue with a proposal how to proceed?
Could you squash your commits into one before merging?

chuangbo:
Thank you, the commits have been rebased.

For the proposal issue, I'm afraid that I'm not quite familiar with elasticsearch, I was using a Magento plugin which based on Elastica and then we faced this performance issue. Sorry.

ruflin:
@chuangbo Thanks a lot for the contribution. No worries, I'm sure someone will pick it up.

PR #1065, opened 2016-03-24
Extract and Implement ResultSetBuilder

merk:
As per #1017, this PR contains the extraction of ResultSetBuilder.

Please consider this for merging.

This PR technically creates a BC break for anyone using ResultSet::$_class and ResultSet::create().

### DIFF
--- a/lib/Elastica/Client.php
+++ b/lib/Elastica/Client.php
@@ -49,6 +49,11 @@ class Client
     protected $_callback;
 
     /**
+     * @var Connection\ConnectionPool
+     */
+    protected $_connectionPool;
+
+    /**
      * @var \Elastica\Request
      */
     protected $_lastRequest;
@@ -62,11 +67,6 @@ class Client
      * @var LoggerInterface
      */
     protected $_logger;
-
-    /**
-     * @var Connection\ConnectionPool
-     */
-    protected $_connectionPool;
 
     /**
      * Creates a new Elastica client.
--- a/lib/Elastica/Index.php
+++ b/lib/Elastica/Index.php
@@ -6,6 +6,7 @@ use Elastica\Exception\InvalidException;
 use Elastica\Exception\ResponseException;
 use Elastica\Index\Settings as IndexSettings;
 use Elastica\Index\Stats as IndexStats;
+use Elastica\ResultSet\BuilderInterface;
 
 /**
  * Elastica index object.
@@ -36,9 +37,7 @@ class Index implements SearchableInterface
      * All the communication to and from an index goes of this object
      *
      * @param \Elastica\Client $client Client object
-     * @param string           $name   Index name
-     *
-     * @throws \Elastica\Exception\InvalidException
+     * @param string $name Index name
      */
     public function __construct(Client $client, $name)
     {
@@ -289,13 +288,14 @@ class Index implements SearchableInterface
 
     /**
      * @param string|array|\Elastica\Query $query
-     * @param int|array                    $options
+     * @param int|array $options
+     * @param BuilderInterface $builder
      *
-     * @return \Elastica\Search
+     * @return Search
      */
-    public function createSearch($query = '', $options = null)
+    public function createSearch($query = '', $options = null, BuilderInterface $builder = null)
     {
-        $search = new Search($this->getClient());
+        $search = new Search($this->getClient(), $builder);
         $search->addIndex($this);
         $search->setOptionsAndQuery($options, $query);
 
--- a/lib/Elastica/Multi/Search.php
+++ b/lib/Elastica/Multi/Search.php
@@ -17,14 +17,9 @@ use Elastica\Search as BaseSearch;
 class Search
 {
     /**
-     * @var array|\Elastica\Search[]
+     * @var MultiBuilderInterface
      */
-    protected $_searches = array();
-
-    /**
-     * @var array
-     */
-    protected $_options = array();
+    private $_builder;
 
     /**
      * @var \Elastica\Client
@@ -32,13 +27,25 @@ class Search
     protected $_client;
 
     /**
+     * @var array
+     */
+    protected $_options = array();
+
+    /**
+     * @var array|\Elastica\Search[]
+     */
+    protected $_searches = array();
+
+    /**
      * Constructs search object.
      *
      * @param \Elastica\Client $client Client object
+     * @param MultiBuilderInterface $builder
      */
-    public function __construct(Client $client)
+    public function __construct(Client $client, MultiBuilderInterface $builder = null)
     {
-        $this->setClient($client);
+        $this->_builder = $builder ?: new MultiBuilder();
+        $this->_client = $client;
     }
 
     /**
@@ -47,18 +54,6 @@ class Search
     public function getClient()
     {
         return $this->_client;
-    }
-
-    /**
-     * @param \Elastica\Client $client
-     *
-     * @return $this
-     */
-    public function setClient(Client $client)
-    {
-        $this->_client = $client;
-
-        return $this;
     }
 
     /**
@@ -149,7 +144,7 @@ class Search
             $this->_options
         );
 
-        return new ResultSet($response, $this->getSearches());
+        return $this->_builder->buildMultiResultSet($response, $this->getSearches());
     }
 
     /**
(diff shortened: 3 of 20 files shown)


lib/Elastica/Search.php
```diff
@@ -462,7 +463,7 @@ public function search($query = '', $options = null)
             $params
         );
 
-        return ResultSet::create($response, $query);
+        return $this->_client->getResultSetBuilder()->buildResultSet($response, $query);
```
merk: I would prefer to inject the ResultSetBuilder directly into the Search object, but I think given the way Elastica is structured that developers will be creating these objects directly and we dont want to introduce such a major change..

lib/Elastica/Client.php
```diff
@@ -62,19 +69,23 @@ class Client
      * @var LoggerInterface
      */
     protected $_logger = null;
+
     /**
-     * @var Connection\ConnectionPool
+     * @var BuilderInterface
      */
-    protected $_connectionPool = null;
```
ruflin: This breaks BC in case someone extended the Client. Not something that prevents us from moving forward, but worth mentioning in the CHANGELOG.md.

lib/Elastica/Index.php
```diff
@@ -295,7 +293,7 @@ public function exists()
      */
     public function createSearch($query = '', $options = null)
     {
-        $search = new Search($this->getClient());
+        $search = new Search($this->getClient(), $this->getClient()->getResultSetBuilder());
```
ruflin: As far as I understand, every Search can be done with a different Result Builder, but by default is the one used set in the client. So instead of passing it here, we could make Search fetching the default result builder from client if the param as null.

lib/Elastica/Multi/Search.php
```diff
@@ -27,18 +32,20 @@ class Search
...
      * Constructs search object.
      *
      * @param \Elastica\Client $client Client object
+     * @param MultiBuilderInterface $builder
      */
-    public function __construct(Client $client)
+    public function __construct(Client $client, MultiBuilderInterface $builder = null)
     {
-        $this->setClient($client);
+        $this->_builder = $builder ?: new MultiBuilder();
```
ruflin: In a way there are 2 default builders: One for normal requests, one for multi requests. In the client only the one for single requests can be set?

lib/Elastica/ResultSet.php
```diff
@@ -15,117 +15,45 @@
...
-    protected static $_class = 'Elastica\\ResultSet';
-
-    /**
      * Results.
      *
-     * @var array Results
+     * @var Result[] Results
      */
-    protected $_results = array();
+    private $_results = array();
```
ruflin: BC breaks as noted in the beginning.

lib/Elastica/Search.php
```diff
@@ -74,9 +81,11 @@ class Search
      * Constructs search object.
      *
      * @param \Elastica\Client $client Client object
+     * @param BuilderInterface $builder
      */
-    public function __construct(Client $client)
+    public function __construct(Client $client, BuilderInterface $builder = null)
     {
+        $this->_builder = $builder ?: new Builder();
```
ruflin: Shouldn't the fall back be the one set in the client as default (see previous comment)?

lib/Elastica/Tool/CrossIndex.php
```diff
@@ -65,7 +65,7 @@ public static function reindex(
         array $options = array()
     ) {
         // prepare search
-        $search = new Search($oldIndex->getClient());
+        $search = $oldIndex->createSearch();
```
ruflin: Unrelated note, but I think this functionality will become obsolete with the new Reindex API coming in ES 2.3

ruflin:
As mentioned before I think this definitively goes in the right direction. I did quite a brief review but I saw that the changes to the public APIs and tests are minimal which is great to make it easier for people to migrate (which in most cases is not needed). Could you update the CHANGELOG.md with the BC breaks?

lib/Elastica/Index.php
merk: The API might make more sense to have an optional ResultSetBuilder parameter in createSearch()? Personally I'd prefer to see not having an optional parameter and force it being set at creation time. It simplifies the code and any possibility of bugs (IME)

lib/Elastica/Multi/Search.php
merk: Yeah, its a problem. I'd prefer not to have any in the Client at all because the Client shouldn't know about this stuff. I'm contemplating removing the Client option and forcing it to be set in the constructor of both the normal and multi Search instances.

lib/Elastica/ResultSet.php
merk: In the last discussion I explained that this is a good change and we should add it to CHANGELOG and bump the major version - manipulating the ResultSet by inheritance isnt correct anymore - changing the implementation of the ResultSetBuilder is the way to manipulate it now.

lib/Elastica/Search.php
merk: As above, I think it would be better for the Client to know nothing about the Builder.

lib/Elastica/Client.php
merk: The connection pool remains protected, my habit of alphabetising properties strikes here, its just now on line 54.

merk:
Regarding the ResultSet $_results variable, to make it simpler we could leave it protected and deprecate it for a major bump instead? That would mean this change has almost no effect on any current users..

lib/Elastica/Index.php
ruflin: What do you exactly mean by "set a creation time"? You mean inside the search object (which seems to be the case also with an optional param) or here inside createSearch? 

As createSearch is "only"  a proxy method it should probably provide all options that can be passed to Search itself. So I somehow like the idea of providing an additional third param which is optional. 

I can see your point about simplifying the code on our side and preventing bugs, but it makes it harder for beginners to start with Elastica. Some "magic" is ok from my point of view, as long as it can be overloaded by setting the specific params.

lib/Elastica/Multi/Search.php
ruflin: I would suggest the following approach first:
* Client has no knowledge of the builders
* Each builder has its own default type

This is in a way identical to the current behaviour. If we later realise that it could be convenient to set a "default" builder for each type we can still figure out, which is the best place to set it.

So I would remove it here probably.

lib/Elastica/ResultSet.php
ruflin: Yeah, that is what I mean by my note, that it should be added to the CHANGELOG. I'm ok with the BC break.

ruflin:
@merk That would be great, if we could deprecate it first. Like this we could do the release as following:

* 3.x with new feature and deprecations
* 4.x with feature removed

This gives people the opportunity to update the lib to the most recent version first without breaking their app but they already see what they have to change.

lib/Elastica/Index.php
merk: I've removed knowledge of the Builder from the Client. Now, the only way to configure which Builder is used is by passing it into a Search constructor or via the proxy createSearch() method on an Index or a Type.

lib/Elastica/Multi/Search.php
merk: I think that in the case of a Multi\Search, we're not going to encounter many people overriding this builder. There's just not much need to customise what it does, but providing that capability means noone has to extend core classes anymore.

The MultiBuilder is implemented that each Search that it runs can have its own builder, which is why I've left the `MultiBuilderInterface` and `MultiBuilder` in the Multi namespace.

I cant really remove the MultiBuilder concept - it has a difference purpose to the ResultSet Builder. There may be a better name for what it is doing?

merk:
I have moved ResultSet properties back to protected with deprecations and updated the CHANGELOG file.

lib/Elastica/Multi/Search.php
ruflin: Lets keep it as it is for the moment, merge it and work from there.

ruflin:
@merk Sorry, took me some time here to get back to this one. Could you rebase on top of master so I can merge this one in?

ruflin:
@merk BTW: ping when you rebase, as Github does not send notifications for this :-(

merk:
Rebased.

ruflin:
Argh, seems you have to rebase again as it conflicts with the logging PR :-(

merk:
Squashed and rebased

ruflin:
Merged. Thanks.

PR #1066, opened 2016-03-24
Add resultset transforming

merk:
The first commit is contained in #1065 and should be merged first.

This PR adds the capability to "transform" results, which for FOSElasticaBundle allows us to insert the original object that generated the result into a new property on the Result object.

It could be that someone would want to somehow transform the $hit returned by ElasticSearch into something else and this mechanism allows that to occur. The primary magic occurs in 'TransformingBuilder' which decorates a real Builder and calls any registered Transformer for each ResultSet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to this project will be documented in this file based on the
 ### Bugfixes
 - Fix fatal error on `Query::addScriptField()` if scripts were already set via `setScriptFields()` #1086
 
+### Added
+- Added the concept of ResultSet Transformers. The Transformer adds more information to a Result, for example the original object or data that created the Result. #1066
+
 ## [3.2.0](https://github.com/ruflin/Elastica/compare/3.1.1...3.2.0)
 
 ### Backward Compatibility Breaks
--- a/lib/Elastica/ResultSet/ProcessingBuilder.php
+++ b/lib/Elastica/ResultSet/ProcessingBuilder.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace Elastica\ResultSet;
+
+use Elastica\Query;
+use Elastica\Response;
+use Elastica\ResultSet;
+
+class ProcessingBuilder implements BuilderInterface
+{
+    /**
+     * @var BuilderInterface
+     */
+    private $builder;
+
+    /**
+     * @var ProcessorInterface
+     */
+    private $processor;
+
+    /**
+     * @param BuilderInterface $builder
+     * @param ProcessorInterface $processor
+     */
+    public function __construct(BuilderInterface $builder, ProcessorInterface $processor)
+    {
+        $this->builder = $builder;
+        $this->processor = $processor;
+    }
+
+    /**
+     * Runs any registered transformers on the ResultSet before
+     * returning it, allowing the transformers to inject additional
+     * data into each Result.
+     *
+     * @param Response $response
+     * @param Query $query
+     * @return ResultSet
+     */
+    public function buildResultSet(Response $response, Query $query)
+    {
+        $resultSet = $this->builder->buildResultSet($response, $query);
+
+        $this->processor->process($resultSet);
+
+        return $resultSet;
+    }
+}
--- a/lib/Elastica/Result.php
+++ b/lib/Elastica/Result.php
@@ -191,11 +191,11 @@ class Result
     /**
      * Returns Document.
      * 
-     * @return \Elastica\Document
+     * @return Document
      */
     public function getDocument()
     {
-        $doc = new \Elastica\Document();
+        $doc = new Document();
         $doc->setData($this->getSource());
         $hit = $this->getHit();
         if ($this->hasParam('_source')) {
@@ -213,6 +213,17 @@ class Result
         $doc->setParams($hit);
 
         return $doc;
+    }
+
+    /**
+     * Sets a parameter on the hit.
+     *
+     * @param string $param
+     * @param mixed $value
+     */
+    public function setParam($param, $value)
+    {
+        $this->_hit[$param] = $value;
     }
 
     /**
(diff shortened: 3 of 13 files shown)


ruflin:
Thanks for splitting this into 2 different PR's, makes it easier to review. Lets get https://github.com/ruflin/Elastica/pull/1065 in and then review this one. As far as I can see this one will be quite small with 5-6 files to change. Also don't forget to update the CHANGELOG.md with the changes in the second commit.

merk:
Is `Transformation` the right name for the code provided in this PR? It might make more sense to make this more generalised - calling it a Result Processor?

I'm not sure what else one might do with the Result hits that come back from ES in this manner, but the primary use case for me is to inject original objects so they're available alongside the hit.

merk:
Rebased to keep up with master and renamed the concept of a Transformer to a ResultSet Processor.

merk:
FWIW, I'm not sure what Codecov means with the comment? It seems to be talking about old commits?

ruflin:
@merk Ignore @codecov-io , I just disabled it again. It seems they have a few issues in the recent days.

ruflin:
@merk I worry that we need to rebase again as soon as I merge one of the two PR. It seems Github started to ignore .gitattributes automatic merge for the changelog. Are you around the next hour to potentially rebase again?

merk:
This one might need more work once the others are merged.

I'm still not 100% the approach is correct, and will want to implement some real processors before submitting this for merging.

merk:
FWIW, the other merge conflicts weren't CHANGELOG - they were in Client.php

ruflin:
@merk Ok, other one is merged. Let me know when you feel comfortable to move forward here. I plan to do a release today or tomorrow with the changes we have in so far. Any objections?

merk:
This wont be ready today, so go ahead with a release, we'll get this one in the next release.

CHANGELOG.md
```diff
@@ -4,11 +4,14 @@ All notable changes to this project will be documented in this file based on the
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/3.1.1...HEAD)
 
 ### Backward Compatibility Breaks
+- Method \Elastica\ResultSet::create and property \Elastica\ResultSet::$class were removed. To change the ResultSet class, implement your own ResultSet Builder. #1065
```
ruflin: That needs probably some adjustment now and be moved up as we did a release in the meantime.

lib/Elastica/Multi/ResultSet.php
```diff
@@ -39,44 +37,13 @@ class ResultSet implements \Iterator, \ArrayAccess, \Countable
     /**
      * Constructs ResultSet object.
      *
-     * @param \Elastica\Response       $response
-     * @param array|\Elastica\Search[] $searches
+     * @param \Elastica\Response $response
+     * @param BaseResultSet[]
```
ruflin: Not sure how phpdoc is going to work with this, but I would assume the type should still be called ResultSet?

lib/Elastica/ResultSet/Builder.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+namespace Elastica\ResultSet;
+
+use Elastica\Query;
+use Elastica\Response;
+use Elastica\Result;
+use Elastica\ResultSet;
+
+class Builder implements BuilderInterface
```
ruflin: Should we call this perhaps "BaseBuilder" or "DefaultBuilder"?

lib/Elastica/ResultSet/ProcessingBuilder.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+namespace Elastica\ResultSet;
+
+use Elastica\Query;
+use Elastica\Response;
+use Elastica\ResultSet;
+
+class ProcessingBuilder implements BuilderInterface
```
ruflin: Perhaps here also "DefaultProcessingBuilder" or "BaseProcessingBuilder"?

ruflin:
@merk In general looks good to me. I added some more notes. Two more questions to make sure I understand it to 100%:

* Most of the users doing search and using ResultSet will keep working the same as the "default" builder is used and is an optional param.
* All implementation which access `_totalHits, _maxScore, _took, _timedOut` will break and will have to use the getters.

lib/Elastica/Multi/ResultSet.php
merk: It is aliased in the import, because there is already a ResultSet in the Multi namespace.

lib/Elastica/ResultSet/Builder.php
merk: I think Builder by itself is fine, but happy to rename to DefaultBuilder

lib/Elastica/ResultSet/ProcessingBuilder.php
merk: Since this is actually a decorator for a normal Builder instance, i wouldnt be calling this one a Default or Base - it is designed to wrap another builder regardless of the implementation.

merk:
Yes, the default builder is going to be used in most cases, there are very few reasons to reimplement it unless you need custom behaviour. The default builder is still used even if you're going to use the ProcessingBuilder.

I think we should undo those property removals and deprecate them as well, no point to cause BC breaks unless we have to.

merk:
Regarding your second point, you've already released 3.2.0 with these BC breaks. Do you want to release 3.2.1 reversing this (this was also part of the previous PR for the ResultSet builder, not this one for the Processor)?

CHANGELOG.md
merk: The notes here except for the processing line were part of the previous PR, github obviously didnt update the diffs when it was merged.

ruflin:
@merk Release 3.2.1 with reverting the BC breaks sounds like a great idea. Wasn't aware we could do it without :-)

ruflin:
Can you rebase on top of master?

merk:
Sent a PR with the BC break reversed

merk:
Rebased

PR #1073, opened 2016-04-04
Fix Type->deleteByQuery throwing fatal error for array queries

### LINKED ISSUE #1072: Type->deleteByQuery throws fatal error.
Similar to #1049, Type->deleteByQuery also throws `Call to a member function toArray() on a non-object` since $query->getQuery() returns array.

anandnalya:
Fixes #1072.


ruflin:
Merged. Thanks a lot for the contribution.

PR #1081, opened 2016-04-28
Fix bad _log call

merk:
Missed removing $response from one of the _log calls.


ruflin:
@merk Looks like travis is still working here :-) Will be online again in a few hours.

PR #1090, opened 2016-05-09
Clear code

ReenExe:
Change `not` - for more easy to read
Remove `else` when can without it
Use PHP 5.4 short array


lib/Elastica/Bulk/ResponseSet.php
```diff
@@ -42,16 +42,13 @@ public function getBulkResponses()
      */
     public function getError()
     {
-        $error = '';
-
         foreach ($this->getBulkResponses() as $bulkResponse) {
             if ($bulkResponse->hasError()) {
-                $error = $bulkResponse->getError();
-                break;
+                return $bulkResponse->getError();
```
ruflin: I personally prefer to have only one return statement per method. I'm aware that this is a personal preference but I would like to keep it the same across the whole library.

lib/Elastica/Aggregation/DateHistogram.php
```diff
@@ -143,14 +143,16 @@ public function setFormat($format)
...
     {
-        $bounds = array();
-        $bounds['min'] = $min;
-        $bounds['max'] = $max;
         // switch if min is higher then max
-        if (strtotime($min) > strtotime($max)) {
-            $bounds['min'] = $max;
-            $bounds['max'] = $min;
-        }
+        $bounds = strtotime($min) > strtotime($max)
```
ruflin: The key for me in writing code is to make it as readable as possible. I'm not sure if this change makes the code more readable (at least not for me).

ruflin:
@ReenExe Thanks a lot for the contribution. I like the changes to the short array notation as it makes the code more readable. We should probably have some conversation about the return changes :-)

lib/Elastica/Aggregation/DateHistogram.php
ReenExe: @ruflin, `before` - we have 1 or 2 operation to init array, `after` always 1

ruflin: I'm aware that this could be a potential performance improvement, but as I don't see this as something that could have a real impact on the overall performance of an app code readability is more important for me.

lib/Elastica/Bulk/ResponseSet.php
ReenExe: I without enough experience in big projects like that - **so if want revert as before** (but read enough books).
In this case `before` - 1 additional variable and 1 additional initialization, `break` - is same as `goto` to end of loop, so i change `goto` and `return` to 1 operation `return`.

ReenExe: And in this project already see more than one return, in my changes too - change return structure

ruflin: @ReenExe Like I mentioned before, the changes are not wrong at all and lots of projects do it this way. It is only about my personal preference. Unfortunately you are right and it is probably not followed in all parts of the code :-(

ReenExe: @ruflin, so what do - revert to one return? and other stay?

ruflin: @ReenExe I had a second look at our existing code and your changes. I made the decision to move away from the 1 return principle in Elastica as quite a few of your changes also make the code more readable. I will review your code a second time with this in mind.

Thanks for bringing up this topic and discussing it.

lib/Elastica/Cluster/Settings.php
```diff
@@ -53,15 +53,13 @@ public function getPersistent($setting = '')
...
-            }
+        if (empty($setting)) {
+            return $settings;
         }
 
-        return $settings;
+        if (isset($settings[$setting])) {
+            return $settings[$setting];
+        }
     }
```
ruflin: We should probably add here `return null` so it is clear nothing is returned if non matches. Yes, that was also missing before. Outcome will be the same, but I think it is nice for readability.

lib/Elastica/Cluster/Settings.php
```diff
@@ -78,29 +76,27 @@ public function getTransient($setting = '')
...
+                if (isset($settings[$key])) {
+                    $settings = $settings[$key];
+                } else {
+                    return;
+                }
+            }
+
+            return $settings;
+        }
     }
```
ruflin: See comment above about `return null`

lib/Elastica/Filter/GeoBoundingBox.php
```diff
@@ -39,15 +39,15 @@ public function __construct($key, array $coordinates)
      */
     public function addCoordinates($key, array $coordinates)
     {
-        if (!isset($coordinates[0]) || !isset($coordinates[1])) {
-            throw new InvalidException('expected $coordinates to be an array with two elements');
-        }
+        if (isset($coordinates[0], $coordinates[1])) {
```
ruflin: I liked the previous code more because the normal behaviour is on the top level of the function and special checks are made for the exception.

lib/Elastica/Index/Settings.php
```diff
@@ -84,22 +84,20 @@ public function get($setting = '')
...
 
-                return $settings;
+        if (strpos($setting, '.') !== false) {
+            // translate old dot-notation settings to nested arrays
+            $keys = explode('.', $setting);
+            foreach ($keys as $key) {
+                if (isset($settings[$key])) {
+                    $settings = $settings[$key];
+                } else {
+                    return;
```
ruflin: return null

ruflin:
@ReenExe I left some additional comments. Lets keep your return refactorings in.

ReenExe:
I agree with `return null` and change it.

lib/Elastica/Filter/GeoBoundingBox.php
ReenExe: Before was:
1. Check as broken - throw
2. Return normal

After:
1. Check normal - return normal
2. Throw

Maybe is `as for me` hard reading `not`.
But return as `was` before/

ruflin: thx

ReenExe:
Can i use short condition:
Example:
Before:
```php
$array = [];
...
if (!empty($array)) ...
```
To:
```php
$array = [];
...
if ($array) ...
```
?

ReenExe:
@ruflin I make changes after review

ruflin:
@ReenExe Can you also add a line to the CHANGELOG.md. Even though this has no functional changes, it will finally and for all break compatibility with PHP 5.3.

ReenExe:
@ruflin, Yes, fix build and update CHANGELOG.md

ReenExe:
@ruflin, hard understand - something wrong with build, i think - when failed build with PHP 5.4 - it`s my mistake, but when update only CHANGELOG.md - and failed 7.0 - this is `magic`

ReenExe:
Think - do better if you close (without merge), and I try fix in my own travis, agree?

ruflin:
@ReenExe I just restarted the build. TBH I don't think the failure is somehow related to your changes. Travis is sometimes flaky, but this more looks like a dependency got out of date?

ruflin:
It is somehow related to this error: 
```
The command '/bin/sh -c if php -r 'exit(version_compare(PHP_VERSION, "5.5.0", ">=") ? 0 : 1);'; then composer install; else composer install --no-dev; fi' returned a non-zero code: 2
```

The code for this can be found in the dockerfile here: https://github.com/ruflin/Elastica/blob/master/Dockerfile#L29

ReenExe:
@ruflin You right, i see https://github.com/composer/composer/issues/5308 and try without --no-dev

ruflin:
@ReenExe Yep, seems like that is the problem :-( We can wait on a fix release or alternatively, perhaps it is possible under PHP 5.4 to install an older version of composer like 1.0.3? https://github.com/ruflin/Elastica/blob/master/env/elastica/Docker54#L30

ReenExe:
@ruflin, Agree, but before try again with option and read discussion on `composer` Issues

ruflin:
@ReenExe Not 100% sure I got your last comment?

ReenExe:
@ruflin, Agree wait - day or two, and after use older composer version (1.0.3) or some composer option that emulate `old` (I'm from Ukraine and this my native language - so yes can mistake, sorry, think will be better with more expirence)

ruflin:
@ReenExe Sounds good to me. No worries, just wanted to make sure as english is a second language for most of us ;-)

ReenExe:
Think they fix it in version 1.1.1 - https://github.com/composer/composer/commit/8f0324f51e4f94c9be18a19dbbc80c52e7b260c0

ruflin:
I assume a release is coming very soon, so lets wait and then roll back the change.

lib/Elastica/Cluster/Settings.php
```diff
@@ -155,13 +154,9 @@ public function setReadOnly($readOnly = true, $persistent = false)
...
         $key = 'cluster.blocks.read_only';
 
-        if ($persistent) {
-            $response = $this->setPersistent($key, $readOnly);
-        } else {
-            $response = $this->setTransient($key, $readOnly);
-        }
-
-        return $response;
+        return $persistent
```
merk: In my opinion, the if block works better like this:

```php

if ($persistent) {
    return $this->setPersistent($key, $readOnly);
}

return $this->setTransient($key, $readOnly);
```

I find code that tries to follow Object Calisthenics to be nicer to read/maintain: http://williamdurand.fr/2013/06/03/object-calisthenics/

lib/Elastica/Document.php
```diff
@@ -97,11 +97,11 @@ public function __unset($key)
      */
     public function get($key)
     {
-        if (!$this->has($key)) {
-            throw new InvalidException("Field {$key} does not exist");
+        if ($this->has($key)) {
```
merk: I think the original way for these exceptions is clearer.

lib/Elastica/Index.php
```diff
@@ -235,36 +235,27 @@ public function refresh()
...
      * @return array Server response
      */
-    public function create(array $args = array(), $options = null)
+    public function create(array $args = [], $options = null)
     {
         $path = '';
-        $query = array();
+        $query = [];
 
         if (is_bool($options)) {
```
merk: We can simplify this block:

```php

if (is_bool($options) && $options) {
    // ...
} else if (is_array($options) {
    // ...
} 
```

lib/Elastica/Index.php
```diff
@@ -531,10 +524,19 @@ public function request($path, $method, $data = array(), array $query = array())
...
      */
-    public function analyze($text, $args = array())
+    public function analyze($text, $args = [])
     {
         $data = $this->request('_analyze', Request::POST, $text, $args)->getData();
 
         return $data['tokens'];
     }
+
+    private function silenceDelete()
```
merk: Is there any use in exposing this functionality as part of a parameter to the delete() method instead?

lib/Elastica/Index/Settings.php
```diff
@@ -77,30 +77,32 @@ public function get($setting = '')
         }
         $settings = $data['settings']['index'];
 
-        if (!$setting) {
-            // return all array
+        /**
```
merk: If there is duplicated code it should be moved into a Util class and be used from both places.

lib/Elastica/Query.php
```diff
@@ -356,8 +356,11 @@ public function addScriptField($name, AbstractScript $script)
      */
     public function addAggregation(AbstractAggregation $agg)
     {
-        if (!array_key_exists('aggs', $this->_params)) {
-            $this->_params['aggs'] = array();
+        /**
```
merk: I think the 'todo' about PHP versions should be removed from this PR and submitted as a separate PR of changes to change the supported PHP version instead.

The PR can then be merged during the release cycle where the decision is made to upgrade (and the debate about what version can be had in that PR)

merk:
This PR is going to cause all kinds of merge headaches if it stays outstanding for too long - It'd be better to get it merged sooner rather than later.

lib/Elastica/Query.php
ruflin: Agree. Didn't spot this one. And We are quite a long way away from removing 5.5 support ;-)

ruflin:
@ReenExe Composer 1.1.1 was released: https://github.com/composer/composer/releases/tag/1.1.1 I restarted the build. Can you rebase on top of master to get it in? It seems to have a conflict.

@merk 👍

ruflin:
Seems like the composer version release is still too fresh and it downloads the old version: https://travis-ci.org/ruflin/Elastica/jobs/129384382#L1211

ReenExe:
@ruflin, I make new clone from master and create new PR in new issue with same changes, it will be more easy for me that fix conflicts in huge project. So close and reopen in 19.05.2016

ruflin:
Looking forward to thew PR. Lets hope then it fetches the newest composer version. Otherwise you could just rebase and squash this one?

ReenExe:
Re-open in #1096

PR #1091, opened 2016-05-11
Fix BC break

merk:


lib/Elastica/ResultSet.php
```diff
@@ -62,6 +86,12 @@ public function __construct(Response $response, Query $query, $results)
         $this->_query = $query;
         $this->_response = $response;
         $this->_results = $results;
+
+        $data = $response->getData();
+        $this->_maxScore = isset($data['hits']['max_score']) ? (float) $data['hits']['max_score'] : 0;
```
ruflin: Could we keep this code in the getters and use here the getters instead? It means we use getData() 4 times but that should be ok?

ruflin:
Thanks. Failing test is because of a composer problem: https://github.com/composer/composer/issues/5308

merk:
IMO, its just neater to leave it this way until we remove the properties

lib/Elastica/ResultSet.php
ruflin: Doesn't matter too much to me :-) Will merge as it is.

PR #1093, opened 2016-05-15
Extract ClientInterface

merk:
Moving further towards support for elasticsearch-php's client library, this codifies the purpose of the Client by defining an interface we expect to see implemented and contains all methods required by the Elastica codebase on the Client.

This is also a potential solution to @aguvillalba's problem with #1070, allowing anyone to implement their own Client decorators to handle runtime initialisation if implementation of a ConnectionPoolInterface is not appropriate, or otherwise swap out a Client implementation for something else.

I'm also personally of the opinion that we should mark Client as final, but that might be a bit much :)

This is a BC break because any extension of a class with a typehint change will need to update its typehints.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ All notable changes to this project will be documented in this file based on the
 
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/3.2.0...HEAD)
 
+### Backward Compatibility Breaks
+- All classes that depend on \Elastica\Client by typehint have been updated to accept a ClientInterface. Any classes that extend Elastica classes that depend on this typehint will need to be updated.
+
 ### Backward Compatibility Fixes
 - Reintroduced properties in ResultSet removed in 3.2.0 as deprecated properties to be removed in 4.0
 
@@ -11,6 +14,7 @@ All notable changes to this project will be documented in this file based on the
 
 ### Added
 - Added the concept of ResultSet Transformers. The Transformer adds more information to a Result, for example the original object or data that created the Result. #1066
+- \Elastica\Client now implements ClientInterface and all Elastica classes accept any implementation of ClientInterface
 
 ## [3.2.0](https://github.com/ruflin/Elastica/compare/3.1.1...3.2.0)
 
--- a/lib/Elastica/Bulk.php
+++ b/lib/Elastica/Bulk.php
@@ -15,7 +15,7 @@ class Bulk
     const DELIMITER = "\n";
 
     /**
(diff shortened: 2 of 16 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ All notable changes to this project will be documented in this file based on the
 
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/3.2.0...HEAD)
 
+### Backward Compatibility Breaks
+- All classes that depend on \Elastica\Client by typehint have been updated to accept a ClientInterface. Any classes that extend Elastica classes that depend on this typehint will need to be updated.
```
ruflin: \Elastica\Client as type hint should still work, no?

merk: Unfortunately not:

```php
php > interface ClientInterface {}
php > class Client implements ClientInterface {}
php > class Type { function test(ClientInterface $client) {} }
php > class NewType extends Type { function test(Client $client) {} }
PHP Strict standards:  Declaration of NewType::test() should be compatible with Type::test(ClientInterface $client) in php shell code on line 1

Strict standards: Declaration of NewType::test() should be compatible with Type::test(ClientInterface $client) in php shell code on line 1
```

ruflin: :-( I'm curious how many people actually extended the client?

merk: Being the entry point to Elastica, enough I think - 

https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/Elastica/Client.php

merk: I think adding this as a 4.0 BC break is acceptable, it means you can swap in a php-elasticsearch wrapper that implements ClientInterface transparently.

ruflin:
LGTM. Can you rebase and ping me?

merk:
I'd like to consider a few more changes that could come with major BC breaks before we tag 4.0.

How do you want to handle merging this stuff? Do you want to branch off 3.2 and start merging into master which will become 4.0?

ruflin:
I suggest we do all our efforts currently in master, and I branch of 3.2 later from a tag if needed. There was only one request in the past to do an old branch (obviously because of the BC breaks with ES 2.0). I created the branch back then but never got the contributions that were promised. So as long as not request I suggest to keep the complexity to a minimum.

ruflin:
@merk What is your preferred way of pushing this one forward?

ruflin:
Closing this PR as it didn't have any movement for the past 7 years. @merk Appreciate that you took this on, please reopen if you plan to work on this again.

PR #1094, opened 2016-05-15
Tidy properties

merk:
Just some tidying of properties where they are set in the constructor and there is no point setting them in the property definition, removing `= null` since it is implicit and removing $_params; redefinitions in classes that extend Param.


ruflin:
LGTM

ruflin:
@merk I just made you a contributor to the repository, so you can also merge PR's. In general I like when PR's are reviewed by a second person, but sometimes I skip it, especially for my own ones as they are not always reviewed :-) Feel free to merge PR's you think need merging. I trust in your judgement.

merk:
No worries :)

PR #1096, opened 2016-05-20
review-again and Merge branch 'clear-code-before' into clear-code

ReenExe:


ReenExe:
@ruflin, Something wrong again, but with build, because I review code again with diff

lib/Elastica/Param.php
```diff
@@ -138,9 +138,12 @@ public function setParams(array $params)
     public function addParam($key, $value)
     {
         if ($key != null) {
-            if (!isset($this->_params[$key])) {
-                $this->_params[$key] = array();
-            }
+            /**
```
ruflin: Please remove this note. Lets keep the logic the same in this PR, so I suggest to uncomment the below again (especially as the tests fail).

lib/Elastica/Search.php
```diff
@@ -195,6 +195,13 @@ public function setOption($key, $value)
     {
         $this->_validateOption($key);
 
+        /**
```
ruflin: Please remove this comment (see comment above)

lib/Elastica/Search.php
```diff
@@ -236,10 +243,12 @@ public function addOption($key, $value)
     {
         $this->_validateOption($key);
 
-        if (!isset($this->_options[$key])) {
-            $this->_options[$key] = array();
-        }
-
+        /**
```
ruflin: Same here

ruflin:
In general LGTM. I don't think something is wrong with the build as the error is consistent across all 4 builds: https://travis-ci.org/ruflin/Elastica/jobs/131748215#L4437

Could it be related to the changes you made / commented out? There seems to be a true / false mix somewhere potentially. I didn't check in detail yet.

lib/Elastica/Query.php
```diff
@@ -349,9 +349,12 @@ public function addScriptField($name, AbstractScript $script)
      */
     public function addAggregation(AbstractAggregation $agg)
     {
-        if (!array_key_exists('aggs', $this->_params)) {
-            $this->_params['aggs'] = array();
-        }
+        /**
```
merk: Another one here.

If this is an issue, aggs should be set in the constructor.

lib/Elastica/QueryBuilder.php
```diff
@@ -51,11 +51,11 @@ public function __construct(Version $version = null)
      */
     public function __call($dsl, array $arguments)
     {
-        if (false === isset($this->_facades[$dsl])) {
```
merk: :-1: on this change - exceptions as gates has a nicer flow imo.

lib/Elastica/Response.php
```diff
@@ -317,11 +309,11 @@ public function getEngineTime()
     {
         $data = $this->getData();
 
-        if (!isset($data['took'])) {
-            throw new NotFoundException('Unable to find the field [took]from the response');
```
merk: Same here

merk:
Please remove the code rearrangement (exceptions being thrown elsewhere) to a separate PR so it can be discussed independently of the array syntax changes.

I'd also like to see the true/false changes as another separate PR since it doesnt relate to the syntax changes either, making it easier to discuss and track.

lib/Elastica/Param.php
ReenExe: I check this in 5.4 http://sandbox.onlinephpfunctions.com/code/8dc8d94b0ccf05c77dffaf2c3b69bda702463f29

merk: Its not the change thats the problem, its that there is no need to have it commented out with a TODO comment.

ruflin:
@ReenExe I like the suggestion from @merk Lets do in a first PR that array conversion. This is about 95% of the changes and should not change any logic. Then we can have a second PR with the other changes to discuss. This also prevents too many merge conflicts.

ruflin:
I tried to investigate further into the composer issue, and I'm not sure what happens. It uses now composer 1.1.1 but the 5.4 build still fails. See https://travis-ci.org/ruflin/Elastica/jobs/133096440#L1777 The strange part is that if I test it locally with PHP 5.6 the --no-dev works as expected, there on travis it seems it doesn't? I need to test it with 5.4 it seems. Any other ideas what the issue is?

ReenExe:
Compare with https://travis-ci.org/ruflin/Elastica/jobs/128814644 - see only
```bash
...
Problem 1
...
Installation request for aws/aws-sdk-php ~3.0 -> satisfiable by aws/aws-sdk-php
...
```

ruflin:
@ReenExe Sorry for the late answer. Not sure what you try to say with the above? Any idea how we could fix it?

ReenExe:
@ruflin I think problems with 5.4 begin after changes in composer.json after https://github.com/ruflin/Elastica/pull/1065

ReenExe:
@ruflin, Sorry my mistake in previous comment.
The problem stay in composer - because composer try require-dev aws/aws-sdk-php

ruflin:
@ReenExe Yep, it seems like the version didn't change in the PR linked. I'm currently trying to find a work around for this issue (again) :-)

ruflin:
Ok, I think I managed to find a work around for the PHP 5.4 issue.

@ReenExe I did also apply linting and interesting it converts all [] into array(...): See https://github.com/ruflin/Elastica/pull/1113 So I assume changing everything to the short syntax would only changing the entry here and apply linting? https://github.com/ruflin/Elastica/blob/master/.php_cs#L19 No manual work needed hopefully :-)

ReenExe:
@ruflin, I see, so sorry - but for me to resolve conflict need make new fork - and merge my changes to review it before make new PR - so i do it, and close this to reopen new.
It's good that build PHP 5.4 are success.

ruflin:
@ReenExe Ok, but try to make it the automated way (no manual work). Please push the array changes separate.

That means I can close this PR?

ReenExe:
@ruflin, Yes, close. I try find how automate or some like
```php

if ($condition) {
  throw new \Exception('some');
} else {
  return $b
}
```

to
```php
if ($condition) {
  throw new \Exception('some');
}
return $b
```

But think - for this automated need contribute to CS

ruflin:
@ReenExe Yeah, for automation I was only referring to the array stuff.

PR #1097, opened 2016-05-24
Add tests for Bulk hasIndex() and hasType()

ebernhardson:
The two functions wern't returning what i would expect, because
the tests check explicitly for the empty string and the default values
were null. Added a couple tests to assert their proper behavior and
then fixed the code to pass the tests.

### DIFF
--- a/lib/Elastica/Bulk.php
+++ b/lib/Elastica/Bulk.php
@@ -25,12 +25,12 @@ class Bulk
     protected $_actions = array();
 
     /**
-     * @var string
+     * @var string|null
      */
     protected $_index;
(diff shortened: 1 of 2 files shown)


merk:
This is incorrect - these values should not have empty string initialisers, the methods should be modified to return true/false based on null instead.

ruflin:
LGTM. I'm ok with merging this one. PHP 5.4 failure is unrelated to this change (composer problem ...)

@merk I assume the changes I see now were pushed after your comment.

merk:
Yep, looks good to me

PR #1099, opened 2016-05-25
Added var_export because sometime $error is an array

leopoiroux:


ruflin:
Which version of Elastica are you using? getErrorMessage should return always a string. See https://github.com/ruflin/Elastica/blob/f545465dd4d5acd020ebf1c1e7846fbefcc0f399/lib/Elastica/Response.php#L130

leopoiroux:
Hi, I use the last version (2.3.2)
Thks for your help

ruflin:
Version 2.3.2 of Elastica is not compatible with elasticsearch 2.x. Please update the version 3.x of Elastica.

ruflin:
Closing this issue as I think it is related to using incompatible version of Elastica and elasticsearch. Please reopen if that is not the case.

PR #1106, opened 2016-06-01
Fixed gzip encoding enable/disable on persistent connections

hectorgimenez:
When using a persistent connection and you enable compression with setCompression, make a request and then disable the compression, the next requests are sent without encoding as gzip (because compression is disabled) but the headers are persisted with the Content-Encoding: gzip, because we are reusing the same curl client and no headers are sent, so $headerConfig is empty and the CURLOPT_HTTPHEADER is never reached.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file based on the
 
 ### Bugfixes
 - Fix fatal error on `Query::addScriptField()` if scripts were already set via `setScriptFields()` #1086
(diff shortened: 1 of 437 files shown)


ruflin:
LGTM. @hectorgimenez Could you update the CHANGELOG.md file with this change?

hectorgimenez:
@ruflin Done.

ruflin:
@hectorgimenez Thanks. Seems like it conflicts with one of the merges I just did before. Could you rebase on top of master?

PR #1108, opened 2016-06-04
Support for non-string error responses.

tishma:
Hi there. I have a PHP 5.3 + ES 2.3.2 requirement. I'm doing good so far with Elastica 2.3.1, and  as long as my tests pass, I'm ok to use it. I've hit some incompatibilities and already have workarounds (outside Elastica) for those. However, your change to support non-string errors in ES 2.3.2 responses appears not to be breaking ES 1.x compatibility, so I've applied what appears to be minimum change on this branch. Could you please consider merging it, so there is one less fork to manage for anyone concerned. Thanks.

### DIFF
--- a/lib/Elastica/Response.php
+++ b/lib/Elastica/Response.php
@@ -1,350 +1,350 @@
 <?php
-    namespace Elastica;
+namespace Elastica;
 
-    use Elastica\Exception\JSONParseException;
-    use Elastica\Exception\NotFoundException;
+use Elastica\Exception\JSONParseException;
+use Elastica\Exception\NotFoundException;
+
+/**
+ * Elastica Response object.
+ *
+ * Stores query time, and result array -> is given to result set, returned by ...
+ *
+ * @author Nicolas Ruflin <spam@ruflin.com>
+ */
+class Response
+{
+    /**
+     * Query time.
+     *
+     * @var float Query time
+     */
+    protected $_queryTime = null;
 
     /**
-     * Elastica Response object.
+     * Response string (json).
      *
-     * Stores query time, and result array -> is given to result set, returned by ...
-     *
-     * @author Nicolas Ruflin <spam@ruflin.com>
+     * @var string Response
      */
-    class Response
+    protected $_responseString = '';
+
+    /**
+     * Error.
+     *
+     * @var bool Error
+     */
+    protected $_error = false;
+
+    /**
+     * Transfer info.
+     *
+     * @var array transfer info
+     */
+    protected $_transferInfo = array();
+
+    /**
+     * Response.
+     *
+     * @var \Elastica\Response Response object
+     */
+    protected $_response = null;
+
+    /**
+     * HTTP response status code.
+     *
+     * @var int
+     */
+    protected $_status = null;
+
+    /**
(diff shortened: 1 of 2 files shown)


merk:
Could you please undo the whitespace changes made to these files so its simpler to review?

tishma:
@merk Thank you for the response. It appears that not only whitespace has slipped in, but also a lot of other code from master. This second commit should be what I meant it to be - minimal change to support non-string errors.

lib/Elastica/Response.php
```diff
@@ -78,14 +78,56 @@ public function __construct($responseString, $responseStatus = null)
...
+        if (!$error) {
+            return '';
+        }
+
+        if (is_string($error)) {
+            return $error;
+        }
+
+        if (isset($error['root_cause'][0])) {
+            $error = $error['root_cause'][0];
```
ruflin: Do you really want to overwrite the error varialbe here? what if there are multiple root_causes?

ruflin:
@tishma So if I understand it right this is to make 2.3 more compatible with 1.x. I'm happy to get this in as longs as it does not introduce and BC breaks.

lib/Elastica/Response.php
tishma: Thanks for looking into this - the code was copied from the master branch (maybe you want to look there too). I did not really get much into semantics, but I don't see why root cause would be an array of strings.

tishma:
@ruflin Actually this is to make Elastica 2.3 branch more compatible with ES 2.3.2 server which appears to have introduced error objects instead of plain strings. I think your master branch code is defensive enough to fall back to old behaviour if error is just a string (which was always the case with ES 1.x), and robust enough to extract the most of error info when it's an object.

ruflin:
The main thing I'm concerned here is not the PR itself, but the impression it will give. Also with this fix Elastica 2.x is not compatible with elasticsearch 2.x. I'm aware that in your case you know about that, but most users don't and then just expect it to work.

What would the changelog state? Make errors messages compatible with elasticsearch 2.x? :-)

ruflin:
@tishma BTW: I have no issue in merging it into the 2.3 branch, I'm more concern about releasing it ...

ruflin:
@tishma Any thoughts on the comments above?

tishma:
@ruflin hey! Thanks for checking in. I thought what you said was already a 'no'. Certainly, you wouldn't be saying Elastica 2.3.x is now supporting ES 1.x. I already have my workaround, although my life would be somewhat easier if you decide to release. I leave it up to you. Thanks again.

ruflin:
Elastica 2.x does support Elasticsearch 1.x, but not elasticsearch 2.x ;-) That is what I worry that people confuse :-D

ruflin:
@tishma Lets see if there is going to be some more interest in this one (means waiting for either +1 on this PR or issue participants).

Tobion:
IMO this can be closed. Elastica 2.x is not meant to be compatible with ES 2.x. And there is not enough interest in making it this fully compatible.

ruflin:
Agree on closing this one as it is quite old and 5.0 happened in the meantime.

PR #1110, opened 2016-06-13
Changelog : Add missing links on #xxx.

ChristopheBoucaut:
I just add missing links in CHANGELOG. Easier to read the changelog :)


ruflin:
LGTM. I always left out the comments as Github does the linking here automatically: https://github.com/ruflin/Elastica/releases But if I copy it to elastica.io this make sense. Thanks.

PR #1111, opened 2016-06-14
Get the root reason from elasticsearch's json when available

maxgalbu:
Sometimes elasticsearch (i'm using 2.3.3) returns an error with two reasons:

```json
{
  "error": {
    "root_cause": [
      {
        "type": "index_creation_exception",
        "reason": "failed to create index"
      }
    ],
    "type": "illegal_argument_exception",
    "reason": "synonym requires either `synonyms` or `synonyms_path` to be configured"
  },
  "status": 400
}
```

But the current code uses only the "root_cause" reason, leaving me with an exception that says "failed to create index" and not much else. This change uses the "root" reason key if it exists.

### DIFF
--- a/lib/Elastica/Response.php
+++ b/lib/Elastica/Response.php
@@ -95,13 +95,18 @@ class Response
             return $error;
         }
 
+        $rootError = $error;
         if (isset($error['root_cause'][0])) {
-            $error = $error['root_cause'][0];
+            $rootError = $error['root_cause'][0];
         }
 
-        $message = $error['reason'];
-        if (isset($error['index'])) {
-            $message .= ' [index: '.$error['index'].']';
+        $message = $rootError['reason'];
+        if (isset($rootError['index'])) {
+            $message .= ' [index: '.$rootError['index'].']';
+        }
+
+        if (isset($error['reason']) && $rootError['reason'] != $error['reason']) {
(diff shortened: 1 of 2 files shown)


lib/Elastica/Response.php
```diff
@@ -96,13 +96,18 @@ public function getError()
             return $error;
         }
 
+        $root_error = $error;
         if (isset($error['root_cause'][0])) {
-            $error = $error['root_cause'][0];
+            $root_error = $error['root_cause'][0];
```
ruflin: Could you use `$rootError` for the name here to adhere to our naming standard?

ruflin:
LGTM. One minor comment about the naming. Could you please also update the CHANGELOG.md? And if possible, rebase it on master. That will fix the broken 5.4 build.

ruflin:
@maxgalbu Oh, Github started to send notifications for new commits. Not sure if this is complete yet, but you missed the CHANGELOG.md part ;-)

maxgalbu:
you're right, I had a brief moment of pause at work :)

maxgalbu:
Do you want a single commit, or multiple commits are ok?

ruflin:
Merged. Thanks a lot for the improvement. Unfortunately I already made the 3.2.1 release today in the morning, so it will go into the next minor.

PR #1122, opened 2016-06-22
Fixed gzip encoding enable/disable on persistent connections

hectorgimenez:
When using a persistent connection and you enable compression with setCompression, make a request and then disable the compression, the next requests are sent without encoding as gzip (because compression is disabled) but the headers are persisted with the Content-Encoding: gzip, because we are reusing the same curl client and no headers are sent, so $headerConfig is empty and the CURLOPT_HTTPHEADER is never reached.


ruflin:
@hectorgimenez Merged. Thanks for the contribution.

PR #1124, opened 2016-06-25
Bugfix array push to string

ReenExe:
Answer for possible future questions
1. First time use short array syntax in https://github.com/ruflin/Elastica/blob/e77bf54cf714b92828106a8d3863f1503c8e7d05/lib/Elastica/Transport/Http.php#L106
2. Fatal error on push as `array` to string http://sandbox.onlinephpfunctions.com/code/dbc5b51d01771f258fb814b12b7a5843052df15d

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file based on the
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/3.2.1...HEAD)
 
 ### Backward Compatibility Fixes
-- Use short array [] - so PHP 5.3 deprecated
+- Use short array syntax `[]` instead of long `array()` - so break compatibility with PHP 5.3
 
 ### Bugfixes
 - Set HTTP headers on each request preventing server error if persistent connection is enabled and compression enabled and later disabled for the same connection.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file based on the
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/3.2.1...HEAD)
 
 ### Backward Compatibility Fixes
+- Use short array [] - so PHP 5.3 deprecated
 
 ### Bugfixes
 - Set HTTP headers on each request preventing server error if persistent connection is enabled and compression enabled and later disabled for the same connection.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file based on the
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/3.2.1...HEAD)
 
 ### Backward Compatibility Fixes
-- Use short array syntax `[]` instead of long `array()` - so break compatibility with PHP 5.3
+- Use short array syntax `[]` instead of long `array()`
 
 ### Bugfixes
 - Set HTTP headers on each request preventing server error if persistent connection is enabled and compression enabled and later disabled for the same connection.
--- a/lib/Elastica/Query/Fuzzy.php
+++ b/lib/Elastica/Query/Fuzzy.php
@@ -57,7 +57,7 @@ class Fuzzy extends AbstractQuery
     {
         //Retrieve the single existing field for alteration.
         $params = $this->getParams();
-        if (count($params) < 1) {
+        if (empty($params)) {
             throw new InvalidException('No field has been set');
         }
         $keyArray = array_keys($params);
--- a/lib/Elastica/Query/Ids.php
+++ b/lib/Elastica/Query/Ids.php
@@ -64,7 +64,7 @@ class Ids extends AbstractQuery
     /**
      * Set type.
      *
-     * @param string|\Elastica\Type $type Type name or object
+     * @param array|string|\Elastica\Type $type Type name or object
      *
      * @return $this
      */
@@ -77,7 +77,7 @@ class Ids extends AbstractQuery
             return $this;
         }
 
-        $this->_params['type'] = $type;
+        $this->_params['type'] = is_array($type) ? $type : [$type];
 
         return $this;
     }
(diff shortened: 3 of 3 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file based on the
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/3.2.1...HEAD)
 
 ### Backward Compatibility Fixes
+- Use short array [] - so PHP 5.3 deprecated
```
merk: Using short arrays doesnt deprecate 5.3, nor is it a BC fix, it is a BC break

lib/Elastica/Query/Ids.php
```diff
@@ -91,11 +91,7 @@ public function setType($type)
      */
     public function setIds($ids)
     {
-        if (is_array($ids)) {
-            $this->_params['values'] = $ids;
-        } else {
-            $this->_params['values'] = array($ids);
-        }
+        $this->_params['values'] = is_array($ids) ? $ids : [$ids];
```
merk: You can just cast $ids to array:

http://sandbox.onlinephpfunctions.com/code/3dcdc5e1f7bd81e72c7da2548f5bb701e9622b70

CHANGELOG.md
ReenExe: @merk, how it's must be?
```
- Use short syntax [] instead long array()
```
Agree?

lib/Elastica/Query/Ids.php
ReenExe: Agree, change to cast

ReenExe:
@merk thanks, done after review, how about now?

CHANGELOG.md
merk: Given there is a requirement of >=5.4 in composer.json this change isnt necessary to list, imo.

ReenExe: Need do it after this https://github.com/ruflin/Elastica/pull/1090#issuecomment-218279606

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file based on the
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/3.2.1...HEAD)
 
 ### Backward Compatibility Fixes
+- Use short array syntax `[]` instead of long `array()` - so break compatibility with PHP 5.3
```
merk: The heading should be 'Backwards Compatibility Breaks', but this changelog entry should be with the composer.json update to >=5.4 since it is already not allowed to be used.

merk: Which is already in the changelog:

https://github.com/ruflin/Elastica/commit/11bcfe299b1b9a30d51b8dc928bf557f7e6dd7fb

ReenExe: @merk, You show me @ruflin's commit at 12 November 2015, but 10 May 2016 @ruflin write in my PR - add `break with PHP 5.3`.
I think need wait for @ruflin

ruflin: Officialy, PHP 5.3 was removed on 12. November 2015. But until now, it was still possible to use almost all of the library with PHP 5.3. In general if there was no big benefit of breaking 5.3, I didn't, but with introducing [] in all files it definitively doesn't work anymore with 5.3. That is why I mentioned it to add it. 

Following suggestion: Lets go with the suggestion from @merk as it will be confusing it "breaking 5.3" shows up again. Sorry about the confusion from my side.

ReenExe: @ruflin, for me need remove comment from `CHANGELOG`?

ruflin: Yes, but just the php 5.3 break part.

ReenExe: done

lib/Elastica/Query/Ids.php
```diff
@@ -77,7 +77,7 @@ public function setType($type)
             return $this;
         }
 
-        $this->_params['type'] = $type;
+        $this->_params['type'] = (array)$type;
```
ruflin: Isn't type a string?

ReenExe: @ruflin,

> 2. Fatal error on push as array to string http://sandbox.onlinephpfunctions.com/code/dbc5b51d01771f258fb814b12b7a5843052df15d

lib/Elastica/Query/Fuzzy.php
```diff
@@ -57,7 +57,7 @@ public function setFieldOption($param, $value)
     {
         //Retrieve the single existing field for alteration.
         $params = $this->getParams();
-        if (count($params) < 1) {
+        if (empty($params)) {
```
ruflin: Hm, reading this again. The outcome is the same but the previous version makes much more sense to be as we want to count the number of entries in $params. It doesn't matter to me if it is count(...) == 0 or size_of() == 0 or the existing one, but not replace it by empty. I don't see the benefit.

lib/Elastica/Query/Ids.php
ruflin: I think I still miss here something. `$this->_params['type']` is already a string, so your example does not apply I think.

ReenExe: Before my changes:
```php
$ids = new Ids();
$ids->setType('some');
$ids->addType('other'); // Fatal error; // because we try [] to string variable
```

ruflin: @ReenExe I think you just discovered a bug. I missed `addType`. Unfortunately we don't have any unit test for this. It seems we have tests for `setType` so the string part seems to work as expected. I'm curious of addType would even properly work with elasticsearch.

ruflin:
@ReenExe How should we proceed here? We "solved" the array to [] conversion but there is still good stuff in this PR, especially the bug you found. Should we open a separate PR for that?

ReenExe:
@ruflin, re-open with same changes?

ruflin:
It would be nice to have a PR with only the bug fix inside so we could add a potential test to make sure it doesn't happen again.

ReenExe:
@ruflin, re-open in #1146

PR #1125, opened 2016-06-25
use Scrutinizer for find and test before clear

ReenExe:
Also test changes http://sandbox.onlinephpfunctions.com/code/b91261fd29da86357e717e7343b37091d1ba3a61

### DIFF
--- a/lib/Elastica/Query/Fuzzy.php
+++ b/lib/Elastica/Query/Fuzzy.php
@@ -35,10 +35,10 @@ class Fuzzy extends AbstractQuery
      */
     public function setField($fieldName, $value)
     {
-        if (!is_string($value) or !is_string($fieldName)) {
+        if (!is_string($value) || !is_string($fieldName)) {
             throw new InvalidException('The field and value arguments must be of type string.');
         }
-        if (count($this->getParams()) > 0 and array_shift(array_keys($this->getParams())) != $fieldName) {
+        if ($this->getParams() && key($this->getParams()) != $fieldName) {
             throw new InvalidException('Fuzzy query can only support a single field.');
         }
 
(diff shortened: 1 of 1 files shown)


lib/Elastica/Query/Fuzzy.php
```diff
@@ -35,10 +35,10 @@ public function __construct($fieldName = null, $value = null)
      */
     public function setField($fieldName, $value)
     {
-        if (!is_string($value) or !is_string($fieldName)) {
+        if (!is_string($value) || !is_string($fieldName)) {
             throw new InvalidException('The field and value arguments must be of type string.');
         }
-        if (count($this->getParams()) > 0 and array_shift(array_keys($this->getParams())) != $fieldName) {
+        if ($this->getParams() && key($this->getParams()) != $fieldName) {
```
merk: Should we be doing a strict comparison here?

ReenExe: @merk, I stay logic `as is` before - only refactoring, and test it before in `sandbox`

merk: Whenever code is touched for changes like this without tests it is worth the effort to validate that there is no functionality changed by adding a unit test. Why cant your sandbox code be adapted to a simple unit test?

lib/Elastica/Query/Fuzzy.php
```diff
@@ -57,13 +57,13 @@ public function setFieldOption($param, $value)
     {
         //Retrieve the single existing field for alteration.
         $params = $this->getParams();
-        if (count($params) < 1) {
+        if (empty($params)) {
```
ruflin: I prefer to keep the count(..) here for the "meaning" of the code. Behaviour should be the same. Perhaps even more meaningful would be `count(...) == 0`

ruflin:
@ReenExe Would be nice if you could add a unit test. Otherwise perhaps remove the more critical change that was discussed.

ReenExe:
@ruflin, Ok, see please https://github.com/ruflin/Elastica/pull/1142

PR #1126, opened 2016-06-25
use key and current without create new array

ReenExe:
Test it http://sandbox.onlinephpfunctions.com/code/289e92423ceab61f9b94793a9a1557c4574bef4b

### DIFF
--- a/lib/Elastica/Suggest/Phrase.php
+++ b/lib/Elastica/Suggest/Phrase.php
@@ -171,10 +171,7 @@ class Phrase extends AbstractSuggest
             $generator = $array[$baseName]['candidate_generator'];
             unset($array[$baseName]['candidate_generator']);
 
-            $keys = array_keys($generator);
-            $values = array_values($generator);
-
-            $array[$baseName][$keys[0]][] = $values[0];
+            $array[$baseName][key($generator)][] = current($generator);
         }
 
         return $array;


merk:
This change needs a unit or functional test, since there is no test coverage of CandidateGenerator with toArray()

ReenExe:
```php

class Container
{
    private $array;

    public function __construct(array $array)
    {
        $this->array = $array;
    }

    public function getArray()
    {
        return $this->array;
    }
}

$container = new Container([
    'a' => 1,
    'b' => 2,
]);

foreach ($container->getArray() as $key => $value) {
    echo "$key => $value\n";
}

echo key($container->getArray()) === 'a' ? 'success' : 'fail';
echo "\n";
echo current($container->getArray()) === 1 ? 'success' : 'fail';
90
$container = new Container([
    1,
    2,
]);

foreach ($container->getArray() as $key => $value) {
    echo "$key => $value\n";
}

echo key($container->getArray()) === 0 ? 'success' : 'fail';
echo "\n";
echo current($container->getArray()) === 1 ? 'success' : 'fail';
```
`$generator` is copy of array that reseted - so we can use `key($generator)` and `current($generator)`

ruflin:
@ReenExe It would be great if you could directly use this code to create a unit test out of it. Let me know if you have troubles setting up the test environment.

ReenExe:
@ruflin, ok, find time to setup this week

ruflin:
@ReenExe ok, ping me if something doesn't work as expected. In case you have docker and docker-compose installed, `make tests` should do the job.

ruflin:
@ReenExe Did you have a chance to add some unit tests here?

ReenExe:
@ruflin, Sorry, agree to close and re-open when possible

ruflin:
@ReenExe We can also keep it open if you think you will get to it some day. Otherwise feel free to close it yourself.

ReenExe:
Find `weakness` in the code and write tests are more easier to me - than install environment, but hope do it

ruflin:
yeah, docker and windows :-) best is probably to try it inside a virtualbox. alternatively if you have phpunit and php locally, it should also work. If you test it just on your local PHP version and then travis tests all the other versions, that is fine for me.

ReenExe:
@ruflin, You give good idea for cheat - i have `main` project (Ubuntu + Vagrant + Elastica) and in subdirectory import `Elastica` for tests - possible will be success.  Will try this week

PR #1127, opened 2016-06-25
clear code same as `Elastica\Status`

ReenExe:
in both `$this->_data = $this->getResponse()->getData();`

### DIFF
--- a/lib/Elastica/Cluster.php
+++ b/lib/Elastica/Cluster.php
@@ -73,14 +73,7 @@ class Cluster
      */
     public function getIndexNames()
     {
-        $metaData = $this->_data['metadata']['indices'];
-
-        $indices = array();
-        foreach ($metaData as $key => $value) {
-            $indices[] = $key;
-        }
-
-        return $indices;
+        return array_keys($this->_data['metadata']['indices']);
     }
 
     /**


lib/Elastica/Cluster.php
```diff
@@ -73,14 +73,7 @@ public function getResponse()
...
     {
-        $metaData = $this->_data['metadata']['indices'];
-
-        $indices = array();
-        foreach ($metaData as $key => $value) {
-            $indices[] = $key;
-        }
-
-        return $indices;
+        return array_keys($this->_data['metadata']['indices']);
```
ruflin: Looks to me like it does the same. To we have a unit test that covers this? You can easily check in the coverage report.

ReenExe: @ruflin, Yes exists test in \Elastica\Test\StatusTest::testGetIndexNames https://github.com/ruflin/Elastica/blob/master/test/lib/Elastica/Test/StatusTest.php#L37

ruflin: Yep, looks good. That is what I was referring to: https://codecov.io/gh/ruflin/Elastica/src/bc7072afc02957035bd1c579e5dbd74e553950e4/lib/Elastica/Cluster.php#L74

ReenExe:
`codecov` - it's new tools for me - thanks

PR #1128, opened 2016-06-25
merge two condition to one

ReenExe:


merk:
To me this doesn't increase the readability of the code, and to me that is more important than the conciseness.

ruflin:
@merk @ReenExe I think that is a general question and also often is personal taste. I support here the opinion of @merk. Code should be like a book to read.

ReenExe:
@ruflin, my variant shorter - so what do? Close or merge?

ruflin:
@ReenExe Will close it. Thanks for looking into it anyways.

PR #1132, opened 2016-06-25
clear code

ReenExe:


lib/Elastica/Result.php
```diff
@@ -197,18 +197,10 @@ public function getDocument()
         $doc = new Document();
         $doc->setData($this->getSource());
         $hit = $this->getHit();
-        if ($this->hasParam('_source')) {
```
ruflin: I thought that in previous version of PHP this was throwing a notice, but I just tried it in PHP 5.4 in it looks fine.

PR #1133, opened 2016-06-25
remove unused `$index`

ReenExe:


lib/Elastica/Multi/MultiBuilder.php
```diff
@@ -38,16 +38,15 @@ private function buildResultSet(Response $childResponse, BaseSearch $search)
      */
     private function buildResultSets(Response $response, $searches)
     {
-        $resultSets = array();
-
         $data = $response->getData();
-        if (!isset($data['responses']) || !is_array($data['responses'])) {
-            return $resultSets;
+        if (empty($data['responses']) || !is_array($data['responses'])) {
```
ruflin: I don't think these two are here the same. If `$data['response] == false` the first one would not catch it.

Overall as we look for an `array` to in the end it seems to be the same. Do we have unit tests here to make sure that is the case?

ReenExe: 1. empty([]) // true
2. empty(false) // true
3. $data = [];
empty($data['responses']) // true

Your case:
$data['responses'] = false;
empty($data['responses']) // true

ruflin: Not sure I get this as:
```
$data['responses'] = false;
empty($data['responses']) // true
!isset($data['responses']) // false
```

Even if his currently the same affect, the question is what our intention is: Do we want to check if the array entry is set or checking if it is empty? These are conceptually two different things and I think it is the first one we are looking for. So I would not change this line.

ReenExe: If see all method - and use `!isset($data['responses'])` instead `empty($data['responses']))` - so we have `foreach ([] as $responseData) {}` so use `empty($data['responses']))` - remove this foreach.
The result of method - same, but in my - without `reset` and `empty foreach`

ruflin: Not sure I can follow your comment above?

ReenExe: In my vesrion - add `early return` that remove foreach on empty array

ruflin: I think the return already existed before?

ReenExe: Yes

ruflin: Not 100% I understand what you mean but best is to just push a new version of what you think should be done which makes it easier to discuss.

ReenExe: Ok, soon re-open

ReenExe:
#1147

PR #1134, opened 2016-06-29
Fix incorred AbstractUpdateAction::setVersionType annotation

ebernhardson:
This function should take a string, specifically 'external', so
annotate as taking a string rather than an integer.


ruflin:
Thanks for spotting and fixing this.

PR #1135, opened 2016-06-29
Fixup a variety of annotation inconsistencies

ebernhardson:
After noticing some incorrect annotations while running analysis against
my own project, i ran the same analysis against Elastica (using
etsy/phan). These annotation cleanups were all made by verifying usages
within Elastica itself.

### DIFF
--- a/lib/Elastica/Result.php
+++ b/lib/Elastica/Result.php
@@ -36,7 +36,7 @@ class Result
      *
      * @param string $name Param name
      *
-     * @return array Result data
+     * @return mixed Result data
      */
     public function getParam($name)
     {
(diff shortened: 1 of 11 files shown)


lib/Elastica/Result.php
```diff
@@ -36,7 +36,7 @@ public function __construct(array $hit)
      *
      * @param string $name Param name
      *
-     * @return array Result data
```
ruflin: Interesting that it swtiched this one to mix.

ruflin:
Thanks a lot. That is quite nice, wasn't aware of this tool. We should add this to our `make lint` command.

PR #1136, opened 2016-06-29
Remove unused extra parameter to _createElasticaResponse

ebernhardson:
This method does not have a second argument, as such remove it from the
caller.


ruflin:
Not sure how that sneaked in. Thanks for spotting and fixing.

PR #1140, opened 2016-07-11
use `php-cs-fixer` with option `short_array_syntax`

ReenExe:


ruflin:
Did you use `make lint` to create the result?

ReenExe:
I use ```bash
php php-cs-fixer fix
```

PR #1146, opened 2016-07-21
test before code - fail (#1124)

ReenExe:


ruflin:
@ReenExe Merged. Thanks for the fix and the tests.

PR #1150, opened 2016-07-25
Added the geo_centroid aggregation

coreation:


ruflin:
LGTM. Would be nice to have at least one integration test added to confirm it works as expected. Best is having a look at the existing integration tests and then do mostly copy / paste :-)

coreation:
@ruflin Added integration test and added the class to the Changelog. Not sure how I can see the merge conflict though.

ruflin:
Best is to rebase on top of master. Probably it is a conflict in CHANGELOG.md

CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ All notable changes to this project will be documented in this file based on the
 ### Bugfixes
 
 ### Added
-
+- `Elastica\Aggregations\GeoCentroid`
```
ruflin: Could you add a newline afterwards?

ruflin:
Changes LGTM

coreation:
Apparently I couldn't find a way to use rebase, my commits were on the master branch of my fork. Not that experienced with rebasing, so I did a merge upstream from the origin master, that indeed showed that the changelog had some differences before I changed it. Looks good now I think :)

CHANGELOG.md
```diff
@@ -38,6 +39,7 @@ All notable changes to this project will be documented in this file based on the
 - Fix php notice on `\Elastica\Index::getAliases()` if index has no aliases [#1078](https://github.com/ruflin/Elastica/issues/1078)
 
 ### Added
+- `Elastica\Aggregations\GeoCentroid`
```
ruflin: Seems like your merge now caused it to be added twice. This one should be removed.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file based on the
 
 ### Added
 - Elastica\QueryBuilder\DSL\Query::geo_distance
+- Elastica\Aggregation\GeoCentroid
```
ruflin: If you already edit the changelog, you could also add #1150 at the end (see change examples below)

ruflin:
Yeah, if you work in master directly rebasing on master is hard ;-) Merging cause a small issue above.

coreation:
You were right, I added it to the wrong version. Should be ok now I think.

ruflin:
Thanks for the contribution. I squashed to commits and put it into master.

coreation:
You're very welcome, loving the library so I'm glad I can give back (small as it may be ;) )

ruflin:
No exact plans for the next as there are already a few changes in an no BC breaks I could do a release quite soonish.

One thing I wanted to fix is the httpproxy vulnerability before the next release: https://httpoxy.org/ Unfortunately Elastica is also affected here (in case you use HHVM_VERSION) and Guzzle Transport: https://github.com/ruflin/Elastica/blob/c01c929424c8efd48eca8f773feecd6902b36c6c/lib/Elastica/Transport/Guzzle.php#L68

coreation:
Ok, thanks for the info, I'm afraid I have too little time to dive into that one :)

PR #1151, opened 2016-07-27
Allow strings for setMinimumMatch in Terms query

### LINKED ISSUE #1149: minimum_match accepts not only int but string
I was trying to run a `Terms` query with `minimum_match` and passing a percentage but was getting "wrong" results. Inspected the method and noticed that its only accepting `int`.

I can't find anything on the ElasticSearch docu about this (only finding `minimum_should_match`) but I ran manually a query against ES and confirmed that you can send "minimum_match" as a string (percentage).

Before doing a PR, I'd like to confirm this.

I'm using `Elastica 2.3.*` - Still being `int` in `3.*`

topikito:
This closes #1149

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file based on the
 
 ### Bugfixes
 - Set HTTP headers on each request preventing server error if persistent connection is enabled and compression enabled and later disabled for the same connection.
+- Removed `int` type hinting in `setMinimumMatch` (`Terms` Query): it should also allow `string`. [#1151](https://github.com/ruflin/Elastica/pull/1151)  
 
 ### Added
 - Elastica\QueryBuilder\DSL\Query::geo_distance
--- a/test/lib/Elastica/Test/Query/TermsTest.php
+++ b/test/lib/Elastica/Test/Query/TermsTest.php
@@ -36,14 +36,38 @@ class TermsTest extends BaseTest
         $this->assertEquals(3, $resultSet->count());
     }
 
+    public function provideMinimumArguments()
+    {
+        return [
+            [
+                3
+            ],
+            [
+                -2
+            ],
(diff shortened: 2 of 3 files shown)


ruflin:
LGTM. Can you also update the changelog?

ruflin:
I'm curious why Travis did not start on this PR?

topikito:
@ruflin Sure, will do

topikito:
@ruflin Done

topikito:
Seems like Travis is now willing to test the changelog

test/lib/Elastica/Test/Query/TermsTest.php
```diff
@@ -36,14 +36,38 @@ public function testFilteredSearch()
...
+            ],
+            [
+                '2<-25% 9<-3'
+            ]
+        ];
+    }
+
     /**
      * @group unit
+     * @dataProvider provideMinimumArguments
```
ruflin: It would be nice to also have this for an integration test to make sure it also works with the current elasticsearch version. but we can add this in an other PR.

ruflin:
Merged. Thanks

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file based on the
 
 ### Bugfixes
 - Set HTTP headers on each request preventing server error if persistent connection is enabled and compression enabled and later disabled for the same connection.
+- Removed `int` type hinting in `setMinimumMatch` (`Terms` Query): it should also allow `string`. [#1151](https://github.com/ruflin/Elastica/pull/1151)
```
ruflin: I will probably move this to improvements before the next release as I don't think it was a "bug". The limitation was more for historical reasons.

topikito: Sure - I just considered as a bug because the expected behaviour was not meeting the functional one.

PR #1152, opened 2016-07-27
[DX] Version method

oleg-andreyev:
adding `getVersion` method to Client.

**Case:** For large platforms it's critical to upgrade incrementally/step-by-step, for this purpose we need to know current version on ES

### DIFF
--- a/lib/Elastica/Client.php
+++ b/lib/Elastica/Client.php
@@ -68,6 +68,11 @@ class Client
     protected $_logger;
 
     /**
+     * @var string
+     */
+    protected $_version;
+
+    /**
      * Creates a new Elastica client.
      *
      * @param array           $config   OPTIONAL Additional config options
@@ -85,6 +90,21 @@ class Client
 
         $this->setConfig($config);
         $this->_initConnections();
+    }
+
+    /**
+     * Get current version
+     *
+     * @return string
+     */
+    public function getVersion()
+    {
+        if ($this->_version) {
+            return $this->_version;
+        }
+
+        $data = $this->request('/')->getData();
+        return $this->_version = $data['version']['number'];
     }
 
     /**
--- a/test/lib/Elastica/Test/ClientTest.php
+++ b/test/lib/Elastica/Test/ClientTest.php
@@ -29,8 +29,8 @@ class ClientTest extends BaseTest
     public function testGetVersion()
     {
         $client = $this->_getClient();
-        $this->assertNotEmpty($version = $client->getVersion());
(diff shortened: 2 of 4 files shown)


lib/Elastica/Client.php
```diff
@@ -88,6 +93,21 @@ public function __construct(array $config = [], $callback = null, LoggerInterfac
...
 
     /**
+     * Get current version
+     *
+     * @return string
+     */
+    public function getVersion()
+    {
+        if ($this->_version) {
+            return $this->_version;
```
ruflin: Is there a reason you "cache" the version? Do you expect that this call is happening more then once during the lifetime of a Client?

test/lib/Elastica/Test/ClientTest.php
```diff
@@ -26,6 +26,16 @@ public function testConstruct()
     /**
      * @group functional
      */
+    public function testGetVersion()
+    {
+        $client = $this->_getClient();
+        $this->assertNotEmpty($version = $client->getVersion());
+        $this->assertSame($version, $client->getVersion());
```
ruflin: Could we use here some like version compare to compare it >= 2.0.0? http://php.net/manual/en/function.version-compare.php

ruflin:
Thanks, definitively a useful addition. Definitively a useful addition.

lib/Elastica/Client.php
oleg-andreyev: It's possible. Imagine two _repositories_: **A**, **B** both are using ES, both should be B/C with 1.7 and 2.0.

test/lib/Elastica/Test/ClientTest.php
oleg-andreyev: I don't think we should hardcode version of ES that is used for tests.
I'll take a look, if it's possible to expose ES_VERSION and use it here.

ruflin: I think it is fine to "hardcode" the minimal compatible version. `>=2.0.0` will work for all future versions. I agree if we "hardcode" it, it should be in a variable like `MINIMAL_COMPATIBLE_VERSION` for reusability and this makes it easier to update it.

oleg-andreyev: I've added new ENV to Dockerfile and then checking it in test, see 0abec0e2961135d17a6d1ae87ef07c07683a5524

ruflin:
@oleg-andreyev Thanks. Merged. This will be part of the next release.

PR #1153, opened 2016-07-28
Error: Wrong parameters for Exception([string $exception [, long $cod…

xulien:
…e [, Exception $previous = NULL]]])

needed to https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1108

### DIFF
--- a/lib/Elastica/Transport/Http.php
+++ b/lib/Elastica/Transport/Http.php
@@ -192,7 +192,7 @@ class Http extends AbstractTransport
     {
         if ($this->getConnection()->hasConfig('curl')) {
             foreach ($this->getConnection()->getConfig('curl') as $key => $param) {
-                curl_setopt($curlConnection, $key, $param);
+                curl_setopt($curlConnection, constant($key), $param);
             }
         }
     }


ruflin:
It seems like this change breaks quite a few things in Elastica. Could you share some more details on what the intention of this change is and how it is supposed to work?

xulien:
hi, 

sorry, it was a quick and dirty push :confounded: 

I'm confronted to "Warning: curl_setopt() expects parameter 2 to be long, string given", so this PR is just a way to ensure the right type for curl option, who should be constant anyway or I missed something.
['CURLOPT_SSL_VERIFYPEER'] = false :red_circle: 
[CURLOPT_SSL_VERIFYPEER] = false :white_check_mark:

ruflin:
I'm not sure how your change should help with this? Based on the docs the second param should be an int: http://php.net/manual/en/function.curl-setopt.php

I don't think it is up to the Elastica code to fix wrong usage of constants. Perhaps I'm missing something here?

oleg-andreyev:
What version of bundle/elastica/elasticsearch you're using?

xulien:
@ruflin 
second param should be an int because it's the index into the options array, or (not in the doc) his constant equivalent.
https://github.com/php/php-src/blob/ae8cf3cb5462353f96b4e53b2500c4f3311bf32d/ext/curl/interface.c#L743

who know CURLOPT_SSL_VERIFYPEER is 64, CURLOPT_SSL_VERIFYHOST is 81 and CURLOPT_SSLCERT is 10025?

`<bad-faith>no one will use integer index instead equivalents constant</bad-faith>`
 
maybe something like that could do the jobs?

`
if (!is_int($key)) {
        $key = constant($key);
}
 curl_setopt($curlConnection, $key, $param);
`

@oleg-andreyev 
the latest (dev-master) for each one, but I'm not sure is relevant here

ruflin:
@xulien Not sure if I'm missing here something, but in the end a constant it just a value and in this case an int. In case someone uses `'CURLOPT_SSL_VERIFYPEER'` and passes a constant as a string instead of the `CURLOPT_SSL_VERIFYPEER` the constant itself, I would see this as something that should be fixed in the code of the implementation? 

The code below is the one that works:
```
curl_setup($conn, CURLOPT_SSL_VERIFYHOST, $param);
```

The following code we should not allow I think and also not fix, as it is some magic but from my perspective it is a problem on the code side.
```
curl_setup($conn, 'CURLOPT_SSL_VERIFYHOST', $param);
```

Please tell me if I'm missing something here.

xulien:
> The following code we should not allow I think and also not fix, as it is some magic but from my perspective it is a problem on the code side.

I would be tempted to say yes, but it's a little helpers, which allow to use string on curl options.

Like in my case, when you pass through symfony, as you give a constant who return an string... after two hour of quest without result to find why and an headache, you enjoy to simply pass a string.

ruflin:
@xulien Sorry to be a little bit stubborn here, but if symfony returns constants as strings, that sounds more like an issue in Symfony. But TBH I'm also missing the context here.

Wouldn't this be easy fixable by using `constan(...)` around the variables where you set the curl options?

ruflin:
I'm closing this one as I really think this is not a Elastica Http output issue. But I'm more then happy to discuss further and let me convince from the opposite :-)

Issue #1157, opened 2016-08-02
Missing multi_value_mode from decay function


istvankis:
Unable to set [multi_value_mode](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html#_multi_values_fields) for function_score decay function.

ruflin:
`multi_value_mode` is supported on decay functions — it is the last argument of `FunctionScore::addDecayFunction()`:

```php
$functionScore = new \Elastica\Query\FunctionScore();
$functionScore->addDecayFunction(
    \Elastica\Query\FunctionScore::DECAY_GAUSS,
    'location',
    '52.3760, 4.8940',
    '10km',
    null, // offset
    null, // decay
    null, // weight
    null, // filter
    \Elastica\Query\FunctionScore::MULTI_VALUE_MODE_AVG,
);
```

Valid values are the `FunctionScore::MULTI_VALUE_MODE_*` constants: `MIN`, `MAX`, `AVG`, `SUM`.

Closing as implemented.

PR #1158, opened 2016-08-02
Added multi_value_mode

istvankis:
Fix for #1157


ruflin:
Thanks for the contribution. Could you update the CHANGELOG and potentially add a small integration test to make sure it works as expected?

lib/Elastica/Query/FunctionScore.php
```diff
@@ -147,7 +148,8 @@ public function addDecayFunction(
         $offset = null,
         $decay = null,
         $weight = null,
-        $filter = null
```
ruflin: I think the addDecayFunction is becoming too many params. We should probably switch to a `$params` array to not have BC breaks in case one of the params disappears. Also this makes adding / non supported params easier.

ruflin: NOTE: Does not necessarly have to be fixed in this PR, just a note.

ruflin:
Thanks, merged. Thanks a lot for the addition.

PR #1161, opened 2016-08-18
Optimize memory usage for Http Adapter

romainneutron:
`$content = str_replace('\/', '/', $content);` consumes a lot of memory, especially with large payloads.

In my import loop, it was consuming 40MB per bulk.

This improves memory consumption

### DIFF
--- a/lib/Elastica/Transport/Http.php
+++ b/lib/Elastica/Transport/Http.php
@@ -122,13 +122,13 @@ class Http extends AbstractTransport
             }
 
             if (is_array($data)) {
-                $content = JSON::stringify($data, 'JSON_ELASTICSEARCH');
+                $content = JSON::stringify($data, JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES);
             } else {
                 $content = $data;
-            }
 
-            // Escaping of / not necessary. Causes problems in base64 encoding of files
-            $content = str_replace('\/', '/', $content);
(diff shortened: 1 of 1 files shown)


ruflin:
Thanks for the PR to make the code faster. Can you please update the CHANGELOG.md file? Adding a test to confirm the optimization of this can is quite tricky, but it looks very reasonable :-)

One question that this raised on my side is, why do we have JSON_ELASTICSEARCH at all? I looks like we could replace it everywhere with JSON_UNESCAPED_UNICODE https://github.com/ruflin/Elastica/blob/master/lib/Elastica/JSON.php#L59 My assumption is we needed it for some historical reasons (PHP version compatibility).

romainneutron:
Would you like a test to cover the optimization or the slashes unescaping? 
If you would like an optimization demonstration, I'd be glad to share a Blackfire profile

ruflin:
The changed code is already covered by quite a few other integration tests I assume as it is such a central part. So I worry not too much that it breaks the code. No need to share the Backfire profile, the changes make a lot of sense. I fully trust you here :-)

In case you have an easy test to add feel free to add it, but not required. Only additional thing required is the update to the CHANGELOG.md

ruflin:
@romainneutron I merged this change as I plan to do a release. I will add the changelog entry.

romainneutron:
Thanks a lot, sorry for being a bit slow on this one

ruflin:
@romainneutron No worries. Thanks a lot for the contribution.

PR #1169, opened 2016-08-31
Create transport from class which don't has prefix name Elastica\Transport

Gasol:
Allow we store configuration in JSON format, and simply restore to in-memory array and pass to Elastica\Client.

### DIFF
--- a/lib/Elastica/Transport/AbstractTransport.php
+++ b/lib/Elastica/Transport/AbstractTransport.php
@@ -99,14 +99,13 @@ abstract class AbstractTransport extends Param
             } else {
                 $transport = ucfirst($transport);
             }
-
-            $className = 'Elastica\\Transport\\'.$transport;
-
-            if (!class_exists($className)) {
-                throw new InvalidException('Invalid transport');
+            $classNames = ["Elastica\\Transport\\$transport", $transport];
+            foreach ($classNames as $className) {
+                if (class_exists($className)) {
+                    $transport = new $className();
+                    break;
+                }
             }
-
-            $transport = new $className();
         }
 
         if ($transport instanceof self) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable changes to this project will be documented in this file based on the
 - Elastica\Aggregation\SerialDiff
 
 ### Improvements
+- Elastica\Client constructor now accepts a transport of fully qualified name. [#1169](https://github.com/ruflin/Elastica/pull/1169)
 
 ## Deprecated
 
--- a/test/lib/Elastica/Test/Transport/AbstractTransportTest.php
+++ b/test/lib/Elastica/Test/Transport/AbstractTransportTest.php
@@ -21,6 +21,7 @@ class AbstractTransportTest extends \PHPUnit_Framework_TestCase
             [['type' => 'Http']],
             [['type' => new Http()]],
             [new Http()],
+            ['Elastica\Test\Transport\DummyTransport'],
         ];
     }
 
--- a/test/lib/Elastica/Test/Transport/DummyTransport.php
+++ b/test/lib/Elastica/Test/Transport/DummyTransport.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace Elastica\Test\Transport;
+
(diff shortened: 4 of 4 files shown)


ruflin:
I don't fully understand this change yet. Can you share some more details and example on this?

Gasol:
Assume we define configuration of Elastica in JSON format.

// config.json
```
{
    "servers": [
        {
            "host": "elasticsearch-1",
            "port": "9200",
            "transport": "ACME\Transport\Stream"
        }
    ],
    "timeout": "1s",
    "connectionStrategy": "RoundRobin"
}
```

We can simply call `json_decode` to get configuration in array structure, then pass it to constructor of `Elastica\Client`. If transport in configuration does not pre-defined in `Elastica\Transport\AbstractTransport`, We will get exception shows `Invalid transport`without this change.

For example:
```php
<?php
$data = file_get_contents('/path/to/config.json');
$config = json_decode($data, true);
$client = new Elastica\Client($config);
```

lib/Elastica/Transport/AbstractTransport.php
```diff
@@ -99,14 +99,13 @@ public static function create($transport, Connection $connection, array $params
             } else {
                 $transport = ucfirst($transport);
             }
-
-            $className = 'Elastica\\Transport\\'.$transport;
-
-            if (!class_exists($className)) {
-                throw new InvalidException('Invalid transport');
+            $classNames = ["Elastica\\Transport\\$transport", $transport];
```
ruflin: I'm kind of thinking if we should turn this around. Have the base assumption that $transport is the full class name, and only have as fall back the Elastica once. I would have to check where in our code we would have to adjust these paths.

lib/Elastica/Transport/AbstractTransport.php
```diff
@@ -99,14 +99,13 @@ public static function create($transport, Connection $connection, array $params
             } else {
                 $transport = ucfirst($transport);
             }
-
-            $className = 'Elastica\\Transport\\'.$transport;
-
-            if (!class_exists($className)) {
```
ruflin: we should still throw an invalid exception in case none of the classes matches.

ruflin:
Nice, I was never thinking of this use case. I left some minor comments in the PR. Could you also update the CHANGELOG.md?

lib/Elastica/Transport/AbstractTransport.php
Gasol: Maybe we could check whether namespace separator (\) exists or not? It means it could be FQCN or built-in transport of Elastica right?

Gasol: It does, If $transport is not a AbstractTransport ([Line 112](https://github.com/ruflin/Elastica/pull/1169/files#diff-dee909e4e66d40dd2dea7d7147d4bbc7L112)), It will throw InvalidException ([Line 119](https://github.com/ruflin/Elastica/pull/1169/files#diff-dee909e4e66d40dd2dea7d7147d4bbc7L119)).

lib/Elastica/Transport/AbstractTransport.php
Gasol: It does, If $transport is not a AbstractTransport ([Line 112](https://github.com/ruflin/Elastica/pull/1169/files#diff-dee909e4e66d40dd2dea7d7147d4bbc7L112)), It will throw InvalidException ([Line 119](https://github.com/ruflin/Elastica/pull/1169/files#diff-dee909e4e66d40dd2dea7d7147d4bbc7L119)).

lib/Elastica/Transport/AbstractTransport.php
Gasol: [Check namespace seperator](https://github.com/ruflin/Elastica/pull/1169#discussion_r77565586) is bad idea, We should not assume that class has a namespace.

Gasol:
@ruflin I have updated this PR for adding changelog.

lib/Elastica/Transport/AbstractTransport.php
ruflin: Got it, missed that line in the diff.

ruflin:
@Gasol Thanks, merged.

PR #1172, opened 2016-09-08
Add InnerHits functionality

wamania:
Add the innerHits functionality as described in
https://www.elastic.co/guide/en/elasticsearch/reference/2.4/search-request-inner-hits.html
for Nested aggregation and Child/parent relations

### DIFF
--- a/lib/Elastica/Query/InnerHits.php
+++ b/lib/Elastica/Query/InnerHits.php
@@ -1,0 +1,175 @@
+<?php
+namespace Elastica\Query;
+
+use Elastica\Script\AbstractScript;
+use Elastica\Script\ScriptFields;
+
+/**
+ * Nested query.
+ *
+ * @author Guillaume Affringue <wamania@yahoo.fr>
+ *
+ * @link https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-inner-hits.html
+ */
+class InnerHits extends AbstractQuery
+{
+    /**
+     * @return array
+     */
+    public function toArray()
+    {
+        $array = parent::toArray();
+
+        // if there are no params, it's ok, but ES will throw exception if json
+        // will be like {"top_hits":[]} instead of {"top_hits":{}}
+        if (empty($array['inner_hits'])) {
+            $array['inner_hits'] = new \stdClass();
+        }
+
+        return $array['inner_hits'];
+    }
+
+    /**
+     * The name to be used for the particular inner hit definition in the response.
+     * Useful when multiple inner hits have been defined in a single search request.
+     *
+     * @param string $name
+     *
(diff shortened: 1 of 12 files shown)


ruflin:
Very nice addition. Could you also update the CHANGELOG.md file with this?

wamania:
Done !

lib/Elastica/Query/InnerHits.php
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    /**
+     * Set script fields.
+     *
+     * @param array|\Elastica\Script\ScriptFields $scriptFields
+     *
+     * @return $this
+     */
+    public function setScriptFields($scriptFields)
```
ruflin: Should we require `ScriptFields` type here? One can just use `new ScriptsFileds(...)` when passing the argument. Like this we can verify the type and have less "magic".

ruflin:
LGTM. Only left one minor comment. Very nice functional tests you added.

lib/Elastica/Query/InnerHits.php
wamania: I used the same structure as Elastica\Aggregation\TopHits which is very similar.

`/**
     * Set script fields.
     *
     * @param array|\Elastica\Script\ScriptFields $scriptFields
     *
     * @return $this
     */
    public function setScriptFields($scriptFields)
    {
        if (is_array($scriptFields)) {
            $scriptFields = new ScriptFields($scriptFields);
        }

        return $this->setParam('script_fields', $scriptFields);
    }`

I think we should keep the same prototype.

ruflin: I think the scriptFields in the TopHits is probably for historical reasons as it was perhaps introduced before ScriptFields and only arrays were passed. 

Unfortunately we can't remove it from TopHits directly as it would break BC. We should add a note there that it is deprecated. Even though it is not consistent I prefer not to build up more legacy and only use ScriptFields here to be supported. Otherwise there will be in more BC breaks if we remove it one day.

wamania: ok, i agree

wamania:
look good now, argument forced to type ScriptFields, test ok

ruflin:
@wamania Merged. Thanks a lot for the addition.

wamania:
Happy to have been usefull, tell me if i can do some thing else.

ruflin:
Help is always welcome :-) The biggest task ahead of Elastica currently from my perspective is: https://github.com/ruflin/Elastica/issues/944 Already several people played around with different approaches (including me) but I think we didn't get to a final conclusion yet. More then happy to hear your thoughts on this one.

PR #1173, opened 2016-09-08
Fixing bad syntax on parent call for ResponseException

aequasi:
Without this, mapping errors throw this:

```
  Wrong parameters for Elastica\Exception\ResponseException([string $message [, long $code [, Throwable $previous = NULL]]])
```


aequasi:
Derp.

ruflin:
?

PR #1178, opened 2016-09-27
Add a constant for the expression language.

ChristopheBoucaut:
Just add a constant for the _expression_ language. It exists since ES 1.3 (maybe before) : https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-scripting.html


ruflin:
@ChristopheBoucaut Merged. Thanks for the addition.

PR #1181, opened 2016-10-04
Throw an exception if the setExecution method is called on RangerFilter.

ChristopheBoucaut:
The ```execution``` option for ```range filter``` was removed since Elasticsearch 2.0.

I try to find a changelog to confirm but in the [1.7 documentation](https://www.elastic.co/guide/en/elasticsearch/reference/1.7/query-dsl-range-filter.html#_execution), the execution option is mentioned but in the [2.x documentation](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/query-dsl-range-query.html), the execution option isn't mentioned in Range Query.

For this query, elasticsearch throws ```"reason": "[range] query does not support [execution]"``` : 
```json
GET index/type/_search
{
  "query": {
     "bool": {
       "filter": {
         "range": {
           "year": {
             "from": 1950,
             "to": 2000,
             "execution": "fielddata"
           }
         }
       }
     }
  }
}
```

Note : The Elastica\Filter\Range class is deprecated so I don't know if my change is relevant.

### DIFF
--- a/lib/Elastica/Filter/Range.php
+++ b/lib/Elastica/Filter/Range.php
@@ -3,6 +3,8 @@ namespace Elastica\Filter;
 
 trigger_error('Deprecated: Filters are deprecated. Use queries in filter context. See https://www.elastic.co/guide/en/elasticsearch/reference/2.0/query-dsl-filters.html', E_USER_DEPRECATED);
 
+use Elastica\Exception\InvalidException;
+
 /**
  * Range Filter.
  *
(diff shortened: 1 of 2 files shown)


ruflin:
@ChristopheBoucaut As we already deprecated it, the next step is to remove it. We definitively should do this for the next major release.

ChristopheBoucaut:
In this case, Elasticsearch doesn't just deprecated the option, it has already removed this option. So, Elasticsearch returns an error if you send the request. My change is just to notify the dev before (+ save time to understand the problem).

But I understand. This class is already deprecated so we shouldn't use it. So you can close this pull request =).

ruflin:
@ChristopheBoucaut  Closing this PR. I need to go through all deprecation notices and remove these for these files for the next major release.

ChristopheBoucaut:
Ok

PR #1183, opened 2016-10-07
Don't distribute tests, build files etc. for composer

willemstuursma:
Don't distribute tests, build files etc. when installing with Composer with `--prefer-dist` (the default). This saves lots of bandwith and file extraction when installing dependencies with default settings.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,7 +5,7 @@
 /.gitignore export-ignore
 /.php_cs export-ignore
 /.travis.yml export-ignore
-/CHANGELOG.md export-ignore merge=union
+/CHANGELOG.md merge=union
 /CONTRIBUTING.md export-ignore
 /Dockerfile export-ignore
 /docker-compose.yml export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,15 @@
-CHANGELOG.md  merge=union
+/.coveralls.yml export-ignore
+/.dockerignore export-ignore
+/.editorconfig export-ignore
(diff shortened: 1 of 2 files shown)


.gitattributes
```diff
@@ -1 +1,15 @@
-CHANGELOG.md  merge=union
```
ruflin: Is there a reason you removed this line?

ruflin:
Wasn't aware that composer takes gitattributes into account. The biggest difference will make the test directory. Can you please update the CHANGELOG.md?

willemstuursma:
Composer does not take it into account, but Github does. Zipballs actually are `git-exports` and hence obey `.gitattributes`. Zipballs are downloaded by Composer when installing with `--prefer-dist` (default).

.gitattributes
willemstuursma: ✔️

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file based on the
 
 ### Improvements
 - Add a constant for the expression language.
+- Composer installations will no longer include tests and other development files.
```
ruflin: This belongs under BC changes as in case someone was using some of these files (for whatever purpose) they will disappear during the next update ...

willemstuursma: Ok, moved the changes.

ruflin:
@willemstuursma Merged. Thanks for the contribution.

willemstuursma:
Thanks, appreciate it.

PR #1190, opened 2016-10-20
Fix fluent query

maxperrimond:
Little fix on query functions to be more fluent


ErikSaunier:
👍

ruflin:
@maxperrimond Thanks for the fix.

PR #1192, opened 2016-10-24
Remove tests on index.merge.policy.merge_factor

p365labs:
index.merge.policy.merge_factor has been removed from ES 1.7
https://www.elastic.co/guide/en/elasticsearch/reference/1.7/index-modules-merge.html

I have removed also setMergePolicyType and the getter as it seems (from the above link) that policy type has been deprecated from 1.6.0


ruflin:
LGTM. Can you update the CHANGELOG.md with this change?

p365labs:
updated the CHANGELOG.md

Issue #1184, opened 2016-10-11
Upgrade to Elasticsearch 5.0


ruflin:
Elasticsearch 5.0 brings several breaking changes which affect Elastica. This issue is to track the open tasks to make Elastica fully compatible with elasticsearch 5.0.
- [x] Remove all depercated methods as this is a major release
- [x] Remove filters as filters are replaced by query
- [ ] Testing
  - [ ] Fix tests which are currently skipped because of incompatibility
  - [ ] Fix `Set fielddata=true on [answer] in order to load fielddata in memory` tests
- [x] Implement percolate query and remove Percolator: https://www.elastic.co/guide/en/elasticsearch/reference/5.0/query-dsl-percolate-query.html
- [x] Remove image plugin support
- [x] Double check which PHP versions should be supported
- [x] Remove filtered query in favor of Bool queries
- [x] Add Painless as script language

Zyqsempai:
Hi there, i will be glad to help you with this migration, please give me some short description, for example for "Remove image plugin support", do you want to completely delete it from project? Or give some other task!

ruflin:
@Zyqsempai Great to hear. What I already did is removing all the filters and skip all the tests which do not pass at the moment. I added a message to each of this tests on why they don't pass (see https://travis-ci.org/ruflin/Elastica/jobs/167274669#L2135). We need to go through each of this test an figure out how it should be fixed. Most of them should be an easy fix.

To start simple: I think the image plugin is not under active development any more. So we should remove this one completely from the code and tests.

Zyqsempai:
ok then i will start with image plugin a then move on to tests

Zyqsempai:
So, what about this type of skipped tests, "script.inline is not enabled. This is required for this test", i checked your elasticsearch.yml and i found this "script.inline: on" i also tried to add something like this "script.engine.groovy.inline.aggs: on" and no results, do you know source of this problem?

Zyqsempai:
Also, this one "Skipped because of 5.0: unknown setting [index.merge.policy.merge_factor] did you mean [index.merge.policy.max_merge_at_once]" this is strange because according to this https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_20_setting_changes.html#_merge_and_merge_throttling_settings , this error mast appear after 2.4 version migration, anyway it seams like we can't fix it for Version > 2.4

ruflin:
[...]

Issue #1138, opened 2016-07-08
`filter` and `post_filter` are not the same - why did one get depracated in favour of the other????


Antoniossss:
What i have found out in the code:
```
   /**
     * Set Filter.
     *
     * @param \Elastica\Query\AbstractQuery $filter Filter object
     *
     * @return $this
     *
     * @link    https://github.com/elasticsearch/elasticsearch/issues/7422
     * @deprecated Use Elastica\Query::setPostFilter() instead, this method will be removed in further Elastica releases
     */
    public function setFilter($filter)
    {
        if ($filter instanceof AbstractFilter) {
            trigger_error('Deprecated: Elastica\Query::setFilter() passing filter as AbstractFilter is deprecated. Pass instance of AbstractQuery instead.', E_USER_DEPRECATED);
        } elseif (!($filter instanceof AbstractQuery)) {
            throw new InvalidException('Filter must be instance of AbstractQuery');
        }

        trigger_error('Deprecated: Elastica\Query::setFilter() is deprecated and will be removed in further Elastica releases. Use Elastica\Query::setPostFilter() instead.', E_USER_DEPRECATED);

        return $this->setPostFilter($filter);
    }

```

Why `setFilter` is depracated in favor of post_filter while they are NOT EQUIVALENT.

`filter` context is executed BEFORE `query` on filtered results, while `post_filter` is executed AFTER `query`. Moreover filtered query can be cached by ES while query cannot. 

What happend here???

Antoniossss:
According to https://github.com/elastic/elasticsearch/issues/7422 it is the same in fact. But still ES docs are not that clear about that. So how to get filtered query now ??

ruflin:
@Antoniossss Is your question on how to do in elasticsearch directly or in Elastica? Also which ES version are you referring to?

Antoniossss:
This was question about Elastica and ES 2.3.x. Sorry for confusion, but I just didnt noticed that there `Elastica\Query` is not extending `AbstractQuery` thus I was looking for `addFilter` method in wrong class. Anyway I will close this issue as I made unnecessary confusion :( sorry bout that.

ruflin:
Yeah, the naming there is for historical reasons not great.

PR #1196, opened 2016-10-29
Remove all depercated methods as this is a major release (#1184)

p365labs:
Following the #1184 , I've removed all the deprecated methods inside the project. For the moment I left the method _log inside Client class, as we already discussed. We will fix in a next future as @ruflin explained me. Can you have a look @ruflin and tell me if it looks ok or if I need to fix the commit ? tnx


p365labs:
@ruflin what kind of error is returning codecov?

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ All notable changes to this project will be documented in this file based on the
 - Remove Elastica/Query/Image.php and test/Elastica/Query/ImageTest.php, no more support for image-plugin.
 - Remove Elastica/Query/Filtered.php and test/Elastica/Query/FilteredTest.php and all uses from code.
 - Remove index.merge.policy.merge_factor, and set/get MergePolicy as it looks deprecated from ES 1.6
+- Remove all deprecated method and classes for the new release.
```
ruflin: Even though the list is going to be quite long, we probably have to be here more specific for people that upgrade and list all the methods that were removed. Please also link to the PR.

lib/Elastica/Aggregation/DateHistogram.php
```diff
@@ -11,34 +9,6 @@
...
-     */
-    public function setPreZone($preZone)
-    {
-        throw new DeprecatedException('Option "pre_zone" is deprecated as of ES 1.5 and will be removed in further Elastica releases. Use "time_zone" instead.');
-    }
-
-    /**
-     * Set post-rounding based on interval.
-     *
-     * @deprecated Option "post_zone" is deprecated as of ES 1.5 and will be removed in further Elastica releases. Use "time_zone" instead.
```
ruflin: We can directly use these comments for the changelog as I think they are quire good. Just replacing `deprecated` by `removed`.

lib/Elastica/Index.php
```diff
@@ -192,23 +192,6 @@ public function deleteDocuments(array $docs)
...
     /**
-     * Optimizes search index.
-     *
-     * Detailed arguments can be found here in the link
-     *
-     * @param array $args OPTIONAL Additional arguments
-     *
-     * @return array Server response
-     *
-     * @deprecated Replaced by forcemerge
```
ruflin: We cannot remove this yet as it was never really released as deprecated. I suggest to keep this in for the next major release and only remove it afterwards.

ruflin: Which also makes me realise, perhaps we should trigger here a deprecated message?

lib/Elastica/Query.php
```diff
@@ -118,21 +118,6 @@ public function getQuery()
...
 
     /**
-     * Set Filter.
-     *
-     * @param \Elastica\Query\AbstractQuery $filter Filter object
-     *
-     * @return $this
-     *
-     * @link    https://github.com/elasticsearch/elasticsearch/issues/7422
-     * @deprecated Use Elastica\Query::setPostFilter() instead, this method will be removed in further Elastica releases
```
ruflin: Need to check if this one was also just deprecated in one of the last commits.

lib/Elastica/Query/Builder.php
```diff
@@ -1,885 +0,0 @@
-<?php
```
ruflin: It's hard to see it going but I agree its time to remove it, especially as it was not really maintained anymore.

lib/Elastica/Index.php
p365labs: maybe a deprecation message could be better in a next release, considering it was never really released as deprecated, I'll just leave @deprecated. What do u guess ?

p365labs: anyway I'll follow your suggestion to trigger a deprecation

lib/Elastica/Query.php
p365labs: setFilter has been deprecated long time ago,
have a look here #1138 
I've also had a look on a bunch of commits even inside this one : 47938554412886f61a3b87d655cb7bf78faee24b
which is from 21/07/2015 so it looks like this is very old

ruflin:
@p365labs You probably have to rebase this PR because of https://github.com/ruflin/Elastica/pull/1197 got merged.

PR #1200, opened 2016-11-03
drop inline.script skip & errors (#1184)

p365labs:
- updated Dockerfile to load propertly elasticsearch.yml and groovy script (https://www.elastic.co/guide/en/elasticsearch/reference/5.0/_script_settings.html#_script_mode_settings)

skipped 3 tests :  after enabling inline.script we have three errors on aggregations, we need to have a look, now skipped

- Elastica\Test\Aggregation\BucketScriptTest::testBucketScriptAggregation --> [reduce] this test is failing
- Elastica\Test\Aggregation\MaxTest::testMaxAggregation --> compile error [reason: all shards failed]
- Elastica\Test\Aggregation\ScriptedMetricTest::testScriptedMetricAggregation --> compile error [reason: all shards failed]

Just to remember:  all the scripts are in Groovy which is not the default scripting language anymore, I suggest to add the new scripting language (painless) inside Script.php and make all the scripts in painless instead of groovy, or maybe adding  also tests for the new default scripting language.

### DIFF
--- a/env/elasticsearch/Dockerfile
+++ b/env/elasticsearch/Dockerfile
@@ -38,8 +38,8 @@ ENV ES_JAVA_OPTS="-Xms512m -Xmx512m"
 RUN ${ES_PLUGIN_BIN} install mapper-attachments
 
 # Copy config files
-COPY *.yml /usr/share/elasticsearch/config/
-COPY scripts/* /usr/share/elasticsearch/config/scripts/
+COPY elasticsearch.yml /opt/elasticsearch-${VERSION}/config/elasticsearch.yml
+COPY scripts/* /opt/elasticsearch-${VERSION}/config/scripts/
 
 COPY docker-entrypoint.sh /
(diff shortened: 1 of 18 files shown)


env/elasticsearch/Dockerfile
```diff
@@ -40,6 +40,8 @@ RUN ${ES_PLUGIN_BIN} install mapper-attachments
 # Copy config files
 COPY *.yml /usr/share/elasticsearch/config/
 COPY scripts/* /usr/share/elasticsearch/config/scripts/
+COPY elasticsearch.yml /opt/elasticsearch-${VERSION}/config/elasticsearch.yml
```
ruflin: I assume that means we don't need line 41 and 42 anymore?

p365labs: argh!!!.. I forget that ones :)

PR #1208, opened 2016-11-09
removed logging.yml from docker ES config. ES uses now log4j2.properties

p365labs:
old logging files not more in use in ES 5.0, now ES use log4j2.properties.

this is the warning returned : 

> ignoring unsupported logging configuration file [/opt/elasticsearch-5.0.0/config/logging.yml], logging is configured via [/opt/elasticsearch-5.0.0/config/log4j2.properties]

have a look here : 
[Elasticsearch commit which warn logging.yml](https://github.com/elastic/elasticsearch/blob/v5.0.0/core/src/main/java/org/elasticsearch/common/logging/LogConfigurator.java#L141)


ruflin:
Using defaults SGTM

PR #1213, opened 2016-11-09
 payload and output are removed from completion suggester

p365labs:
Updated  Elastica\Test\Suggest\CompletionTest.php
- [ES 5.0 completion suggester output removed](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_50_suggester.html#_simpler_completion_indexing)

- [payloads removed in ES 5.0](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_50_suggester.html#_completion_suggester_is_document_oriented)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file based on the
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/5.0.0-beta1...master)
 
 ### Backward Compatibility Fixes
+- Updated Elastica\Test\Suggest\CompletionTest now payload and output are removed
 
 ### Bugfixes
 
@@ -89,7 +90,7 @@ $indices[$indexName]
     - Elastica\Test\TypeTest::testUpdateDocumentWithIdForwardSlashes
     - Elastica\Test\TypeTest::testUpdateDocumentWithParameter
     - Elastica\Test\TypeTest::testUpdateDocumentWithFieldsSource
-  - Composer installations will no longer include tests and other development files.
+- Composer installations will no longer include tests and other development files.
 
 
 ## [3.2.3](https://github.com/ruflin/Elastica/compare/3.2.2...3.2.3)
--- a/test/lib/Elastica/Test/Suggest/CompletionTest.php
+++ b/test/lib/Elastica/Test/Suggest/CompletionTest.php
@@ -17,40 +17,29 @@ class CompletionTest extends BaseTest
         $index = $this->_createIndex();
         $type = $index->getType('song');
 
-        $this->_markSkipped50('Mapping definition for [fieldName] has unsupported parameters:  [payloads : true]');
         $type->setMapping([
             'fieldName' => [
                 'type' => 'completion',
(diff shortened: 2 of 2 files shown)


Zyqsempai:
@p365labs LOL)) wee forgot to tell each other on which issue we work, i just finished my PR!
@ruflin check my commit and decide, i think my is more complete i also changed some fields name according to new documentation. https://www.elastic.co/guide/en/elasticsearch/reference/5.0/search-suggesters-completion.html

p365labs:
@Zyqsempai you're right 👍 we forget to organise each other on which issue to work 😄 
regarding **prefix** I had a doubt, to be honest, I did not find any document stating that **text** has been removed or deprecated. 
And more having a look into 
[SuggestionBuilder.java](https://github.com/elastic/elasticsearch/blob/master/core/src/main/java/org/elasticsearch/search/suggest/SuggestionBuilder.java#L56) the text property is in every branch (5.0/5.x/master).
prefix is used in every example of the current release but not really explained (just a little bit on fuzzy suggests).
maybe I agree we should use prefix, I'm not sure if we have to leave text.

ruflin:
@Zyqsempai Did you see somewhere that text was removed / deprecated in the CHANGELOG?

Zyqsempai:
@ruflin Nope, i made my decision relying on that example https://monosnap.com/file/uCLeePKeotCPexILK87twJPM8OE1Li , but here is my thoughts about it, first of all it is about clarity and obviousness for users, we mast follow documentation! And maybe it's not deprecated in that version but it can be in next, especially if Elastic team will have many questions like this.

ruflin:
@p365labs @Zyqsempai I suggest for the prefix part we will use the PR from @Zyqsempai . I left there a comment.

CHANGELOG.md
```diff
@@ -89,7 +89,8 @@ $indices[$indexName]
     - Elastica\Test\TypeTest::testUpdateDocumentWithIdForwardSlashes
     - Elastica\Test\TypeTest::testUpdateDocumentWithParameter
     - Elastica\Test\TypeTest::testUpdateDocumentWithFieldsSource
-  - Composer installations will no longer include tests and other development files.
+- Composer installations will no longer include tests and other development files.
+- Updated Elastica\Test\Suggest\CompletionTest now payload and output are removed
```
ruflin: @p365labs Can you rebase on master and move that above the beta1 release line? I'm not even sure we need this changelog entry as it is only an update.

p365labs: @ruflin done.

p365labs: have a look now :) my fault

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file based on the
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/5.0.0-beta1...master)
 
 ### Backward Compatibility Fixes
+- Composer installations will no longer include tests and other development files.
```
ruflin: How did this line end up here?

ruflin:
LGTM

PR #1214, opened 2016-11-09
#1184 - Removed "payload option" from suggestions - removed from Elas…

Zyqsempai:
…ticSearch 5.0

      - Changed "text" field to "prefix" in suggestions - changed in ElasticSearch 5.0

### DIFF
--- a/lib/Elastica/Suggest/AbstractSuggest.php
+++ b/lib/Elastica/Suggest/AbstractSuggest.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace Elastica\Suggest;
 
 use Elastica\Exception\InvalidException;
@@ -16,9 +17,14 @@ abstract class AbstractSuggest extends Param implements NameableInterface
     protected $_name;
 
     /**
-     * @var string the text for this suggestion
+     * @var string the prefix for this suggestion
      */
-    protected $_text;
+    protected $_prefix;
+
+    /**
+     * @var string the regex for this suggestion
+     */
+    protected $_regex;
 
     /**
      * @param string $name
@@ -31,15 +37,29 @@ abstract class AbstractSuggest extends Param implements NameableInterface
     }
 
     /**
-     * Suggest text must be set either globally or per suggestion.
+     * Suggest prefix must be set either globally or per suggestion.
      *
-     * @param string $text
+     * @param string $prefix
      *
      * @return $this
      */
-    public function setText($text)
+    public function setPrefix($prefix)
     {
-        $this->_text = $text;
+        $this->_prefix = $prefix;
+
+        return $this;
+    }
+
+    /**
+     * Suggest regex must be set either globally or per suggestion.
+     *
+     * @param string $regex
+     *
+     * @return $this
+     */
+    public function setRegex($regex)
+    {
+        $this->_regex = $regex;
 
         return $this;
     }
@@ -65,6 +85,19 @@ abstract class AbstractSuggest extends Param implements NameableInterface
     }
 
     /**
+     * Expects one of the next params: max_determinized_states - defaults to 10000,
+     * flags are ALL (default), ANYSTRING, COMPLEMENT, EMPTY, INTERSECTION, INTERVAL, or NONE.
+     *
+     * @param array $value
+     *
+     * @return $this
+     */
+    public function setRegexOptions(array $value)
+    {
+        return $this->setParam('regex', $value);
+    }
+
+    /**
      * @param int $size maximum number of suggestions to be retrieved from each shard
      *
      * @return $this
@@ -78,7 +111,7 @@ abstract class AbstractSuggest extends Param implements NameableInterface
      * Sets the name of the suggest. It is automatically set by
      * the constructor.
      *
-     * @param string $name The name of the suggest.
+     * @param string $name The name of the suggest
      *
      * @throws \Elastica\Exception\InvalidException If name is empty
      *
@@ -110,8 +143,14 @@ abstract class AbstractSuggest extends Param implements NameableInterface
     public function toArray()
     {
         $array = parent::toArray();
+        if (isset($this->_prefix)) {
+            $array['prefix'] = $this->_prefix;
+        }
         if (isset($this->_text)) {
             $array['text'] = $this->_text;
+        }
+        if (isset($this->_regex)) {
+            $array['regex'] = $this->_regex;
         }
 
         return $array;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ All notable changes to this project will be documented in this file based on the
 - Updated Elastica\Test\Suggest\CompletionTest now payload and output are removed
 - Updated Elastica\Test\TypeTest::testGetDocumentWithFieldsSelection The stored_fields parameter will only return stored fields — it will no longer extract values from the _source
 - remove _shutdown for Node and Cluster as deprecated
+- Removed "payload option" from suggestions - removed from ElasticSearch 5.0
+- Changed "text" field to "prefix" in suggestions - changed in ElasticSearch 5.0
 
 ### Bugfixes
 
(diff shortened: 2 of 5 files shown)


lib/Elastica/Suggest/AbstractSuggest.php
```diff
@@ -16,9 +17,9 @@
     protected $_name;
 
     /**
-     * @var string the text for this suggestion
+     * @var string the prefix for this suggestion
      */
-    protected $_text;
+    protected $_prefix;
```
ruflin: I think both still exists. I'm not a big fan of local variables as we have `params` which brings more flexilibty. I was thinking if we should remove the variable completely and use params for prefix and text.

Zyqsempai: @ruflin Yes you'r right both params still exists, in other way https://github.com/ruflin/Elastica/pull/1213 this one wasn't successful, and idea to remove variable sounds good to me, but i think we should keep only only "prefix" param to avoid misunderstanding!

p365labs: Ok no I think to have a more clear overview of the problem :

in ES we have in [SuggestBuilder](https://github.com/elastic/elasticsearch/blob/master/core/src/main/java/org/elasticsearch/search/suggest/SuggestionBuilder.java#L56) these params :

- protected static final ParseField TEXT_FIELD = new ParseField("text");
- protected static final ParseField PREFIX_FIELD = new ParseField("prefix");
- protected static final ParseField REGEX_FIELD = new ParseField("regex");
- protected static final ParseField FIELDNAME_FIELD = new ParseField("field");
- protected static final ParseField ANALYZER_FIELD = new ParseField("analyzer");
- protected static final ParseField SIZE_FIELD = new ParseField("size");
- protected static final ParseField SHARDSIZE_FIELD = new ParseField("shard_size");

I think we should follow what ES is doing, prefix and text still exists in all latest branches "5.0/5.x/master ".

Anyway
we are dealign with the :
- field 
- size
- shard_size
with a **setParam**

we are dealign with the :
- text
- prefix

with a **property**

I think the idea of @ruflin is the more smoth and logic to use PARAMS; and maybe we are loosing analyzer :)  

what do you think guys ? @ruflin @Zyqsempai ?

p365labs: Ok I understand ur point @Zyqsempai but I think we have mixed problems : 

1 we need to solve the skipped test by CompletionTest.php class
2 we need clarify if text has been dropped or removed (it looks like in ES that text is not just a wrapper of prefix )
3 we, also, have not implemented **regex** in [suggesters completion](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-suggesters-completion.html#regex)

i propose this approach : 
1 - merging the #1213 (which just solves the tests)
2 - we should implement **regex** 
3 - text has not been deprecated and not even removed everywhere, I understand that maybe is for BC but there's nothing inside ES (branches that explains that it remains for BC). 

My suggestion is to keep **text** (until we had the evidence Elastic want to remove it) and **prefix** and add **regex**, and implement them through setParam as suggested by @ruflin which seems a more optimised solution.

So the idea is to split 1 from 2-3 which could be done together.

How does it sounds ?

ruflin: Plan SGTM. @Zyqsempai Any objections? If not, will merge #1213 and for 2 and 3 we can use this or future PR's.

Zyqsempai: @ruflin , @p365labs I agree, well then i will handle 2 and 3 with this PR.

p365labs: ok for me guys 👍

Zyqsempai:
@ruflin , @p365labs Hi guys, sorry for delay in preparing this one, so you can find not quiet beautiful decision with local variables))) i know it, may be it is too late evening now, but i think this is he best decision for that task, the main problem is level of parameters, if we refuse of using local variables in the AbstractSuggest class and replace them with params we facing problem that those params appear on the completion params level and not on suggest level, we can fix it by implementing global params in the Suggest class (like this one https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Suggest.php#L31 ) but in my opinion it will bad idea because we will need to instantiate and Suggest class and for example Complition class. Something like that)) i admit that i can be totally wrong i am glad to here other ideas.
Also i implemented regex))

ruflin:
@Zyqsempai Can you rebase your changes on top of master? I think currently i contains lots of changes from other PR's / commits. Best squash it into 1 commit.

Zyqsempai:
@ruflin now it is ok?

ruflin:
@Zyqsempai Looks like you still have 11 commits instead of 1 and you did some merging. I was more thinking of rebasing and squashing.

Zyqsempai:
@ruflin Strange, i did "soft reset" on my commits and recommited them as one, also before that i did  rebase with master, sorry i am not so good in GIt workflow, can you point me a better way?

ruflin:
My approach is normally:

```
git checkout master
git pull elastica-master
git checkout your-branch
git rebase -i master
```
The use `s` to squash all the commits into 1 except the first one. Afterwards you need `git push --force`. But as you already did some merging not sure if the above could become ugly because you will have quite a few conflicts during rebasing that you have to fix and then do `git rebase --continue`. 

Otherwise you create a new clean branch and just cherry pick the commits you want. But there also seems to be some mixing.

Sorry, not sure what the best to recommend is.

Zyqsempai:
@ruflin I think it works)) second one is rebase to solve Changelog conflict

lib/Elastica/Suggest/AbstractSuggest.php
```diff
@@ -31,15 +37,29 @@ public function __construct($name, $field)
...
+
+    /**
+     * Suggest regex must be set either globally or per suggestion.
      *
-     * @param string $text
+     * @param string $regex
      *
      * @return $this
      */
-    public function setText($text)
```
ruflin: If I look at the 5.0 examples, `text` shows up quite frequently: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-suggesters.html

lib/Elastica/Suggest/AbstractSuggest.php
```diff
@@ -16,9 +17,14 @@
     protected $_name;
 
     /**
-     * @var string the text for this suggestion
+     * @var string the prefix for this suggestion
+     */
+    protected $_prefix;
```
ruflin: It would be nice if we could move this to the already existing `_params` array.

lib/Elastica/Suggest/AbstractSuggest.php
```diff
@@ -65,6 +85,19 @@ public function setSize($size)
...
 
     /**
+     * Expects one of the next params: max_determinized_states - defaults to 10000,
+     * flags are ALL (default), ANYSTRING, COMPLEMENT, EMPTY, INTERSECTION, INTERVAL, or NONE.
+     *
+     * @param array $value
+     *
+     * @return $this
+     */
+    public function setRegexOptions(array $value)
```
ruflin: How is this different from setRegex?

lib/Elastica/Suggest/AbstractSuggest.php
```diff
@@ -110,9 +143,15 @@ public function getName()
     public function toArray()
     {
         $array = parent::toArray();
+        if (isset($this->_prefix)) {
```
ruflin: If we use _param this should not be needed.

Zyqsempai:
@ruflin Hi, here is something wrong with commit, in my last commits i returned back to place "_text",  about "params" array, i think that you missed my comment, please check it above, i described my point of view.  regexOptions is for inner options, i better show you, https://monosnap.com/file/eL7DUm2Ob8B4CylforcVwsOQXDB0Vl 
this cause all kind of problems  that i described in comment above.

ruflin:
@Zyqsempai Didn't set it but I think we should use `_setRawParam()` which is part of `Param` for this. Hopefully this should solve it?

Zyqsempai:
@ruflin Ok, i will try it!

PR #1215, opened 2016-11-11
store_fields don'textract values from the _source, use mapping 'store' (#1184)

p365labs:
The stored_fields parameter will only return stored fields — it will no longer extract values from the _source.

I've updated the test in \Elastica\Test\TypeTest.php::testGetDocumentWithFieldsSelection, there was the need to have "stored" ('store' => 'yes') the fields extracted from the Docuement.

https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_50_search_changes.html#_literal_fields_literal_parameter

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file based on the
 
 ### Backward Compatibility Fixes
 - Updated Elastica\Test\Suggest\CompletionTest now payload and output are removed
+- Updated Elastica\Test\TypeTest::testGetDocumentWithFieldsSelection The stored_fields parameter will only return stored fields — it will no longer extract values from the _source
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file based on the
 
 ### Backward Compatibility Fixes
 - Updated Elastica\Test\Suggest\CompletionTest now payload and output are removed
+- Updated Elastica\Test\TypeTest::testGetDocumentWithFieldsSelection The stored_fields parameter will only return stored fields — it will no longer extract values from the _source
```
ruflin: I think when we just update tests and don't to any chances to the actual code, we don't need the changelog normally. Going to merge it anyways ;-)

PR #1219, opened 2016-11-16
JSON_FORCE_OBJECT to prevent malformed requests

axel-tietje:
Sending data with an empty array causes an ES error, since it's stringified like so:
`{"query":{"match_all":[]}}` instead of `{"query":{"match_all":{}}}`. 
Using JSON_FORCE_OBJECT fixes that.

### DIFF
--- a/lib/Elastica/Transport/Http.php
+++ b/lib/Elastica/Transport/Http.php
@@ -122,7 +122,7 @@ class Http extends AbstractTransport
             }
 
             if (is_array($data)) {
-                $content = JSON::stringify($data, JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES);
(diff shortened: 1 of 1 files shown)


ruflin:
@axel-tietje Thanks for the contribution. Can you share where exactly you hit this issue? Because MatchAll does this already (just manually). The reason I would like to know more about this is change is to understand if there are some potential side affects.

axel-tietje:
I created a raw query like so:

    $client = new Elastica\Client($config);
    $search = new Elastica\Search($client);
    $search->addIndex('myindex')->addType('mytype');
    $rawQuery = ['query' => ['match_all' => []]];
    $query = new Elastica\Query($rawQuery);

axel-tietje:
Well, it has side effects, since other raw queries do not work any more, since everything is converted to objects. Looks like my personal workaround has to be

`$rawQuery = ['query' => ['match_all' => new \stdClass() ]];`

Please reject this pull request.

PR #1223, opened 2016-11-28
Regex for suggest completion

Zyqsempai:
@ruflin I will close this one https://github.com/ruflin/Elastica/pull/1214 if you do not mind.
This one is clear and elegant as you suggested, with setRawParams it works great.


lib/Elastica/Suggest/AbstractSuggest.php
```diff
@@ -110,9 +141,6 @@ public function getName()
     public function toArray()
```
ruflin: I think now we can completely remove the toArray as it will just take the parent one.

Zyqsempai:
@ruflin Done.

ruflin:
thx. merged

PR #1224, opened 2016-11-29
Properly pass query options with Multi\Search

nomoa:
According to the doc only few options need to be passed in the header of a
MultiSearch. Unfortunately the current behavior is to sent all query options
into the header. For instance terminate_after or timeout are ignored by elastic
if the search is sent via a MultiSearch.
This patch adds an explicit filter of query options to pass in the header and
write remaining ones in the search request body.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file based on the
 - remove _shutdown for Node and Cluster as deprecated
 
 ### Bugfixes
+- Query options such as "timeout" or "terminate_after" should not be ignored when using Multi\Search
 
 ### Added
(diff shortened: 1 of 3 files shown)


nomoa:
I'm very new to this build system so I may have done something wrong but the build failure seems unrelated (ERROR: Service 'elasticsearch' failed to build: Error: image elasticsearch/elasticsearch:latest not found)... Not sure how to fix that.

ruflin:
@nomoa It was a problem with our builds which is now fixed in master. Can you rebase on top of master?

nomoa:
@ruflin thanks!

ruflin:
@nomoa Thanks for the rebase. Could also add a CHANGELOG entry?

nomoa:
@ruflin sure, the commit should be updated with a new entry in the changelog.

ruflin:
@nomoa Merged. Thanks for the contribution.

nomoa:
Thanks!

PR #1230, opened 2016-12-14
Prepare 5.0.0 release

ruflin:


Zyqsempai:
@ruflin Need any help with release, something left?

ruflin:
@Zyqsempai There are still some tests which we skip on Travis, but I had a quick look at it and I think the issues is in the tests and no necessarly in Elastica itself. I was thinking to push the 5.0.0 release as we had no major bug reports in the last month and shipping 5.0.0 will also get more hands on the release. 

@Zyqsempai Anything from your side which you think could be an issue for the release?

Zyqsempai:
@ruflin Nope, i think you'r right! With major release i will update Elastica + ES in my project maybe than we will have some new issues.

PR #1236, opened 2017-01-16
Escape date math in request path

### LINKED ISSUE #1228: Slashes in URL need to be escaped when using date math expr in index names
My index name is `<log-{now/d-1d}>`. This index works fine for `_bulk` writes,
because it is placed inside the request body and doesn't need to be escaped.

However, if we use this index name for `_search` or other operations where index is specified in the URL, it needs to be escaped: `<log-{now%2Fd-1d}>` .

I image that the escaping should happen somewhere around here:

https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Transport/Http.php#L61

date_math expressions in index names are supported since `2.0`.
Our ES version is `2.4.1` and the package version is `3.2.3`

gskema:
Resolves #1228

I couldn't run the full test suite on Windows (Docker problems), so I pushed my solution + unit test (to get things moving). If someone could run this on Docker, that wouldd great. I'll try to run Docker suite at work.

Integration test for this would be declaring an index name with date math, then using it for both writing `_bulk` (index name ends up in JSON body, needs to be unescaped) and reading/doing something, e.g. `_refresh`, where index name is in URI (needs to be escaped).

### DIFF
--- a/lib/Elastica/Util.php
+++ b/lib/Elastica/Util.php
@@ -11,6 +11,72 @@ namespace Elastica;
  */
 class Util
 {
+    /** @var array */
+    protected static $dateMathSymbols = ['<', '>', '/', '{', '}', '|', '+', ':', ','];
+
+    /** @var array */
+    protected static $escapedDateMathSymbols = ['%3C', '%3E', '%2F', '%7B', '%7D', '%7C', '%2B', '%3A', '%2C'];
+
+    /**
+     * Checks if date math is already escaped within request URI.
+     *
+     * @param string $requestUri
+     *
+     * @return bool
+     */
+    public static function isDateMathEscaped($requestUri)
+    {
+        // In practice, the only symbol that really needs to be escaped in URI is '/' => '%2F'
+        return false !== strpos($requestUri, '%2F');
+    }
+
+    /**
+     * Escapes date math symbols within request URI.
+     * @see https://www.elastic.co/guide/en/elasticsearch/reference/5.x/date-math-index-names.html
+     *
+     * @param string $requestUri
+     *
+     * @return string
+     */
+    public static function escapeDateMath($requestUri)
+    {
+        if (empty($requestUri)) {
+            return $requestUri;
+        }
+
+        // Check if date math if used at all. Find last '>'. E.g. /<log-{now/d}>,log-2011.12.01/log/_refresh
+        $pos1 = strrpos($requestUri, '>');
+        if (false === $pos1) {
+            return $requestUri;
+        }
+
+        // Trim leading slash if present
+        if ($startsWithSlash = '/' === $requestUri[0]) {
+            $requestUri = ltrim($requestUri, '/');
+        }
+
+        // Find the position up to which we should escape.
+        // Should be next slash '/' after last '>' E.g. <log-{now/d}>,log-2011.12.01/log/_refresh
+        $pos2 = strpos($requestUri, '/', $pos1);
+        $pos2 = false !== $pos2 ? $pos2 : strlen($requestUri);
+
+        // Cut out the bit we need to escape: <log-{now/d}>,log-2011.12.01
+        $uriSegment = substr($requestUri, 0, $pos2);
+
+        // Escape using character map
+        $escapedUriSegment = str_replace(static::$dateMathSymbols, static::$escapedDateMathSymbols, $uriSegment);
+
+        // '\\{' and '\\}' should not be escaped
+        if (false !== strpos($uriSegment, '\\\\')) {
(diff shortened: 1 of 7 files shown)


lib/Elastica/Util.php
```diff
@@ -11,6 +11,73 @@
...
+        }
+
+        // Check if date math if used at all. Find last '>'. E.g. /<log-{now/d}>,log-2011.12.01/log/_refresh
+        $pos1 = strrpos($requestUri, '>');
+        if (false === $pos1) {
+            return $requestUri;
+        }
+
+        // Trim leading slash if present
+        if ($startsWithSlash = '/' === $requestUri[0]) {
```
gskema: Not sure if we should keep this check, as Http and HttpAdapter assume that `/` is present, but Guzzle transport trims it and reappends. Opinions?

lib/Elastica/Util.php
```diff
@@ -11,6 +11,73 @@
...
+     * Checks if date math is already escaped within request URI.
+     *
+     * @param string $requestUri
+     *
+     * @return bool
+     */
+    public static function isDateMathEscaped($requestUri)
+    {
+        // In practice, the only symbol that really needs to be escaped in URI is '/' => '%2F'
+        return false !== strpos($requestUri, '%2F');
```
gskema: Not sure but we may need `%2f` check here?

ruflin: We could make the url lower case first and then compare to `%2f`?

lib/Elastica/Util.php
ruflin: I would remove it as I'm not sure how this is related to the date math escaping?

lib/Elastica/Util.php
```diff
@@ -11,6 +11,73 @@
...
+     * @see https://www.elastic.co/guide/en/elasticsearch/reference/5.x/date-math-index-names.html
+     *
+     * @param string $requestUri
+     *
+     * @return string
+     */
+    public static function escapeDateMath($requestUri)
+    {
+        // No need to escape if a second slash is not present
+        if (empty($requestUri) || false === strpos($requestUri, '/', 1)) {
```
ruflin: Can you elaborate on the slash check? Not sure I get the relation between this and date math. Or is this a more general problem?

lib/Elastica/Util.php
```diff
@@ -11,6 +11,73 @@
...
+        }
+
+        // Trim leading slash if present
+        if ($startsWithSlash = '/' === $requestUri[0]) {
+            $requestUri = ltrim($requestUri, '/');
+        }
+
+        // Find the position up to which we should escape.
+        // Should be next slash '/' after last '>' E.g. <log-{now/d}>,log-2011.12.01/log/_refresh
+        $pos2 = strpos($requestUri, '/', $pos1);
```
ruflin: I think this will also work with multiple indices 👍

lib/Elastica/Util.php
gskema: I was wondering if all `$path` arguments start with `/`, but since `Http` and `HttpAdapter` assume they do, I'll remove this.

lib/Elastica/Util.php
gskema: I'll check if that's needed at all

lib/Elastica/Util.php
gskema: I wanted to add a quick check if we even need to escape the path. This covers `/_bulk`, `/_analyze` and such cases - quickly exits because there is no escaping to be done.

ruflin:
@gskema Change looks good to me. Do you plan to add some more code or is it ready to be merged from your perspective?

ruflin:
Oh, don't forget the CHANGELOG.md ;-)

gskema:
@ruflin I want to add an integration test, where we create and index, query it with `...{now/d}` through URI and push some bulk documents (index name ends up in JSON body, should be unescaped), using the same `Client`. I wanted to ask where should I put this test?

Also, under which version should I put the change note?

ruflin:
For the integration tests, I would put it in the IndexTest.php as I think that is where it is closest to (at least there I would search for it).

For the changelog, put it under Unreleased.

ruflin:
@gskema Ping me if you think it is ready to be merged.

gskema:
@ruflin I got confused on whether the request path begin with a slash or don't. After some time reading the code I think they never do. I you can confirm this, then it is good to be merged.
It'd be nice if someone would test on a real life ES :)

ruflin:
@gskema Merged as is. ES 5.x is more picky about nothing having double slashes etc. so I think if something is wrong in the above, we will know it quite soon.

Thanks a lot for this contribution.

PR #1238, opened 2017-01-26
Fix param documentation for flush(), add type hint

mweibel:
`Index::request()` has a type hint on the 4th parameter to be an array, so adding the type hint is ok.


ruflin:
@mweibel Thanks for the fix.

Issue #297, opened 2013-01-13
Implement Elastica_Alias to handle alias create and features


ruflin:
Alias has now much more capabilities in elasticsearch. Because of this it is planned to take the complexity Elastica_Index and move it into its own object.

See also discussion in pull request: https://github.com/ruflin/Elastica/pull/161

PR #1242, opened 2017-01-31
Finalize elasticsearch 5 migration

Tobion:
Some things that are missing for #1184 

Todo:

- [x] remove features that do not exist in ES 5.0
  -  e.g. missing, ttl, top_children
- [x] deprecate things that are also deprecated in ES 5.0
  -  e.g. `minimum_number_should_match`, `geo_distance_range`, `Query\Match::setFieldType` (see https://github.com/elastic/elasticsearch/pull/17536 )
- [x] Elastica\Script namespace pretty broken. Also stored script and referencing them by id is not implemented
  - refactored as much as possible without breaking BC. unfortunately the design of the scripts extending `AbstractUpdateAction` is not good. scripts can be part of an update action but they are not the action itself. so the `$id` parameter that I renamed to `$documentId` (which was hard to grasp) doesn't belong there at all. but changing that would require rewrite of the scripts logic.
  - added a `ScriptId` class to reference stored scripts. This was completely missing so far.
- [ ] fix error parsing to exception (it already has a todo in code)
- [x] `MatchPhrase` and `MatchPhrasePrefix` should NOT extend `Match` because they don't share the same options
- [x] fix skipped tests
- [x] There is not routing parameter when creating an index (and has never been). So this feature has always been useless but ES 5 now validates the wrong param. Creating aliases has the possibility to specify routing and a filter. But the alastica api doesn't provide those params yet, which could be added in the future (there is already an issue about this: #297)


ruflin:
@Tobion Thanks for taking this on. Can you ping me when I should have a look?

Tobion:
@ruflin This should be ready now. Please have a look. The last piece I need to look at is the error parsing that seems to be a little strange at the moment. But identifying errors/success seems to work according to the tests.

ruflin:
@Tobion Great work. Any chance you could update the CHANGELOG with some notes on the change? A summary as you have in the PR works for me.

lib/Elastica/Bulk/Action/AbstractDocument.php
```diff
@@ -9,20 +9,20 @@
...
 {
     /**
-     * @var \Elastica\Document|\Elastica\Script\AbstractScript
+     * @var Document|AbstractScript
      */
     protected $_data;
 
     /**
-     * @param \Elastica\Document|\Elastica\Script\AbstractScript $document
+     * @param Document|AbstractScript $document
```
ruflin: How does it know which is the correct path to `Document` as it is in a different path?

lib/Elastica/Index.php
```diff
@@ -249,40 +249,30 @@ public function refresh()
...
-        $query = [];
-
-        if (is_bool($options)) {
-            if ($options) {
-                try {
-                    $this->delete();
-                } catch (ResponseException $e) {
-                    // Table can't be deleted, because doesn't exist
-                }
+        if (is_bool($options) && $options) {
```
ruflin: Note for myself: Need to check this in detail again.

lib/Elastica/Query/GeohashCell.php
```diff
@@ -1,10 +1,12 @@
 <?php
 namespace Elastica\Query;
 
+trigger_error('Elastica\Query\GeohashCell is deprecated.', E_USER_DEPRECATED);
+
 /**
  * Class GeohashCell.
  *
- * @link https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-geohash-cell-query.html
+ * @link https://www.elastic.co/guide/en/elasticsearch/reference/2.4/query-dsl-geohash-cell-query.html
```
ruflin: This was removed in 5.0? If yes, we can also remove it here as it not work with 5.0 anyways, correct? But we should mention it in the changelog.

Tobion: I ran the query against 5.0 and it didn't give an error. So I assume it was only removed by accident from the documentation and is only deprecated, just as geo_distance_range.

lib/Elastica/Bulk/Action/AbstractDocument.php
Tobion: I don't understand the question. `Elastica\Document` is already imported via use statement.

Tobion:
I added a detailed changelog. I would really prefer not to split the PR because it would create alot of overhead and conflicts like in the changelog if I have to add each entry separately.

If you have questions let me know. We can also have a skype call if you want to chat about something.

lib/Elastica/Index.php
mfeltscher: Why not `$options === true`?

lib/Elastica/Bulk/Action/AbstractDocument.php
ruflin: The question probably is, if phpdoc automatically also detects this. So I assume the answer is yes.

ruflin:
Ok, let me try to review this PR in its entirety 👍

lib/Elastica/Index.php
Tobion: Would work as well.

lib/Elastica/Script/AbstractScript.php
```diff
@@ -2,28 +2,160 @@
...
+    const LANG_NATIVE = 'native';
+    const LANG_EXPRESSION = 'expression';
+    const LANG_PAINLESS = 'painless';
+
     /**
-     * @param array|null $params
-     * @param string     $id
+     * @var string
      */
-    public function __construct(array $params = null, $id = null)
```
ruflin: This seems to be a BC break, right?

test/lib/Elastica/Test/Base.php
```diff
@@ -128,11 +128,6 @@ protected function _createIndex($name = null, $delete = true, $shards = 1)
         return $index;
     }
 
-    protected function _markSkipped50($message)
```
ruflin: AWESOME

test/lib/Elastica/Test/Query/GeoDistanceRangeTest.php
```diff
@@ -1,66 +1,12 @@
...
 
-use Elastica\Document;
-use Elastica\Query;
 use Elastica\Query\GeoDistanceRange;
-use Elastica\Query\MatchAll;
-use Elastica\Test\Base as BaseTest;
+use Elastica\Test\DeprecatedClassBase;
 
-class GeoDistanceRangeTest extends BaseTest
+class GeoDistanceRangeTest extends DeprecatedClassBase
```
ruflin: I like that idea.

lib/Elastica/Script/AbstractScript.php
Tobion: Only for the AbstractScript which also contains the language as well. But I don't think anyone implemented an own subclass. The existing subclasses like ScriptFile still have the same constructor. So I don't think anyone is affected by this.

Tobion:
@ruflin Do you not try to reflect the versioning of elastica with elasticsearch compatibility? So then the new version would be 5.0.1 and not 5.1. I'd like to create another PR afterwards with some tiny changes to upgrade the dependency to elasticsearch 5.1

lib/Elastica/Script/AbstractScript.php
ruflin: I also assume that everyone that used it hopefully contributed it back to Elastica. For completeness I would still like to mention it under breaking changes. That doesn't mean I will not create 5.1 out of it ;-)

ruflin:
Yes I try to reflect compatibility. Is there anything inside which breaks with 5.1? All seems to also work with 5.2: https://github.com/ruflin/Elastica/pull/1245 If we do 5.1 I agree we should have elasticsearch first set to 5.1.

Tobion:
There does not seem to be anything that break with 5.1. But the field to reference scripts by id has been renamed from `id` to `stored`. Both versions work but stored seems to be the documented way now, see https://github.com/elastic/elasticsearch/pull/21179

So the only thing I'd like to do for 5.1 is to change this line: https://github.com/ruflin/Elastica/pull/1242/files#diff-154aad1be01fbe69e4f5790bcf273893R57

lib/Elastica/Script/AbstractScript.php
Tobion: Done

ruflin:
@Tobion Merged 🎉  THANKS

PR #1243, opened 2017-02-01
Added support for search option 'request_cache'

ilanrivers:


ruflin:
@ilanrivers Thanks for the contribution. Could you also add a quick integration test for this one?

ilanrivers:
@ruflin Where exactly would this best fit? I do not see any other integration test for the other options.

ruflin:
I think it should be in https://github.com/ruflin/Elastica/blob/master/test/lib/Elastica/Test/SearchTest.php There seem to be at least some tests which cover other constants.

Tobion:
I don't think this needs a test because it's difficult to test if ES used a cache or not internally which this flag controls. So this is good to merge IMO.

Zyqsempai:
Guys we need to merge it, it is a really long awaited feature.
Tests are not needed here, they can be only formal, "Request passed or not".

ruflin:
I'm ok with merging it. But we need a CHANGELOG entry. Either @ilanrivers has time to do it or @Zyqsempai or @Tobion perhaps have the time to do a follow up PR with the CHANGELOG? Or I can do it next week ... @Tobion Feel free to merge it.

Tobion:
Added the changelog

PR #1249, opened 2017-02-14
Fix typo in deprecated warning for Index::optimize

nomoa:
The warning points to Query::forcemerge but should be Index::forcemerge.


ruflin:
@nomoa Thanks for the fix.

Issue #457, opened 2013-09-09
Index Exists will return false if an index exists but is marked as read-only.


paulrblakey:
Index->exists will return false if an index exists but is marked as read-only.

PR #1251, opened 2017-02-16
Fix reading bool index setting & deprecate ElasticsearchException

Tobion:
- Use the explicit getter Response::getStatus instead of the curl transfer info
- Fix reading bool index setting, e.g. when it uses `'off'` instead of `'0'` (where just casting to bool doesn't work)
- Deprecate ElasticsearchException which is irrelevant since elasticsearch now exposes the errors as a structured array instead of a single string. So the whole purpose of the class is gone.
- Add test to show #457 is solved by now

### DIFF
--- a/lib/Elastica/Exception/ElasticsearchException.php
+++ b/lib/Elastica/Exception/ElasticsearchException.php
@@ -1,6 +1,8 @@
 <?php
 namespace Elastica\Exception;
 
+trigger_error('Elastica\Exception\ElasticsearchException is deprecated. Use Elastica\Exception\ResponseException::getResponse::getFullError instead.', E_USER_DEPRECATED);
+
 /**
  * Elasticsearch exception.
  *
@@ -28,15 +30,13 @@ class ElasticsearchException extends \Exception implements ExceptionInterface
     /**
      * Constructs elasticsearch exception.
      *
-     * @param int   $code  Error code
-     * @param array $error Error object from elasticsearch
+     * @param int    $code  Error code
+     * @param string $error Error message from elasticsearch
      */
     public function __construct($code, $error)
     {
-        $this->_error = $error;
-        // TODO: es2 improve as now an array
-        $this->_parseError(json_encode($error));
-        parent::__construct(json_encode($error), $code);
+        $this->_parseError($error);
+        parent::__construct($error, $code);
     }
 
     /**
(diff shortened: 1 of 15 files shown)


lib/Elastica/Response.php
```diff
@@ -133,13 +126,7 @@ public function getFullError()
      */
     public function getErrorMessage()
     {
-        $error = $this->getError();
-
-        if (!is_string($error)) {
```
Tobion: getError always returns string. So this is useless

lib/Elastica/Response.php
```diff
@@ -28,13 +28,6 @@ class Response
     protected $_responseString = '';
 
     /**
-     * Error.
-     *
-     * @var bool Error
-     */
-    protected $_error = false;
```
Tobion: Property that doesn't even have a getter and is not really used

lib/Elastica/Index/Settings.php
```diff
@@ -185,13 +194,12 @@ public function setBlocksWrite($state = true)
      */
     public function getBlocksMetadata()
     {
-        // TODO will have to be replace by block.metadata.write once https://github.com/elasticsearch/elasticsearch/pull/9203 has been fixed
```
Tobion: This has been merged and makes it possible to use the correct `read_only` setting again which was changed in #738

lib/Elastica/Exception/ElasticsearchException.php
```diff
@@ -28,15 +30,13 @@ class ElasticsearchException extends \Exception implements ExceptionInterface
     /**
      * Constructs elasticsearch exception.
      *
-     * @param int   $code  Error code
-     * @param array $error Error object from elasticsearch
+     * @param int    $code  Error code
+     * @param string $error Error message from elasticsearch
```
Tobion: This has been string before and is still receiving a string in https://github.com/ruflin/Elastica/pull/1251/files#diff-38e97e4d9ad047da8141046011e72f90R66

lib/Elastica/Index/Settings.php
```diff
@@ -103,6 +103,25 @@ public function get($setting = '')
     }
 
     /**
+     * Returns a setting interpreted as a bool.
```
ruflin: Nice addition. I hope that ES will also clean this up in the future and only have 1 option of returning bools.

lib/Elastica/Response.php
ruflin: Not sure if we introduced this to make master "partially" compatible with 2.x. Happy to remove it now and lets readd it in case things break ...

PR #1254, opened 2017-02-20
Fix AnalyzeAPI with 'explain' parameter

thePanz:
The ElasticSearch AnalyzeAPI has an "explain" parameter to gather further details about the performed analysis on a text content.
Unfortunately the ElasticSearch response is different when such parameter is set or not set.
See https://www.elastic.co/guide/en/elasticsearch/reference/5.2/_explain_analyze.html
and https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-analyze.html.

The format "explain = true" is consistent in the 5.x major release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ All notable changes to this project will be documented in this file based on the
 
 ### Improvements
 
+- added support for the "explain" flag of AnalyzeAPI [PR-1254](https://github.com/ruflin/Elastica/pull/1254)
+
 ### Deprecated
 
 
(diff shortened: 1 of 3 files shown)


Tobion:
👍 would be nice if you could add a note to the changelog file

thePanz:
@Tobion : Changelog added

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ All notable changes to this project will be documented in this file based on the
 
 ### Improvements
 
+- added support for the "explain" flag of AnalyzeAPI
```
ruflin: @thePanz Could you also add the PR number behind it? We missed it in the recent changelog entries but I want to add it again as it makes reading changelogs easier as you can jump to the PR directly.

thePanz: @ruflin PR number and link added. Does it work OK for you?

ruflin:
@thePanz LGTM. Thanks

PR #1255, opened 2017-02-21
Re-enable php 5.5 support

ebernhardson:
This set of patches adjusts composer.json to specifically target php 5.5, and then adjusts the test suite to be compatible with running php5.5.  Tested by running `TARGET=55 make tests`.


composer.json
```diff
@@ -12,7 +12,7 @@
         }
     ],
     "require": {
-        "php": ">=5.6.0",
+        "php": ">=5.5.0 <5.6.0",
```
ruflin: Good idea, so people use this for ONLY 5.5

ruflin:
@ebernhardson Do you want to readd the 5.5 target in the .travis file so all PR's against this branch get also tested by travis?

ebernhardson:
Not sure why github closed it when i pushed ... anyways I updated the patch to have .travis.yml run with env: TARGET="55"

.travis.yml
```diff
@@ -8,9 +8,7 @@ branches:
     - master
 
 env:
-  - TARGET="56"
-  - TARGET="70"
-  - TARGET="71"
+  - TARGET="55"
```
ruflin: Good idea to only have 5.5. Makes it even more obvious this is only for 5.5.

ruflin:
@ebernhardson Thanks, merged. Not sure why these commits didn't trigger travis :-( I checked the travis config but it should build all branches :-( Lets see what happens with the next PR.

ruflin:
@ebernhardson Posted to early. I think I found the issue in the `.travis.yml` file: https://github.com/ruflin/Elastica/blob/master/.travis.yml#L6 Can you also update the branch here?

PR #1258, opened 2017-02-24
Enable travis on the 5.1-php55 branch

ebernhardson:
The .travis.yml file currently limits travis to only
running on the master branch. Update to include the
new branch supporting php 5.5 as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ services:
 branches:
   only:
     - master
+    - 5.1-php55
 
 env:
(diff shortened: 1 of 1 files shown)


ebernhardson:
I think this will do the trick, but I'm frightfully unfamiliar with travis so it's mostly a guess

ruflin:
This PR should probably go against the php55 branch and not agains master?

ebernhardson:
indeed it should have, will fix

Issue #1261, opened 2017-03-02
Search::count() modifies query


z38:
When calling `Elastica\Search::count($query)` with a `Elastica\Query`, the query gets modified. Specifically, the param `size` is set to zero.

Example:
```php
$client = new Elastica\Client();
$search = new Elastica\Search($client);

$query = new Elastica\Query();
var_dump($query->hasParam('size')); // false
$search->count($query);
var_dump($query->hasParam('size')); // true
```

I'm not sure whether this is the intended behavior, please feel free to close this issue in case it is expected to work like this.

ruflin:
I think the idea here was to set a default. Will it have any affect on your querey? Does it also reset if you set a size before?

z38:
As it's not documented, I didn't expect that the query gets modified and assumed it will be the same after counting. The param is set to `0` in every case.

I was trying to figure out if this should be fixed here or in FOSElasticaBundle (which is tracked in [#1215](https://github.com/FriendsOfSymfony/FOSElasticaBundle/issues/1215) now).

XWB:
@z38 Everyone using Elastica could face this issue, not just `FOSElasticaBundle`.

XWB:
https://github.com/ruflin/Elastica/pull/1264

PR #1264, opened 2017-03-07
Do not modify query in Search::count

XWB:
Related to https://github.com/ruflin/Elastica/issues/1261

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file based on the
 - Fix reading bool index settings like `\Elastica\Index\Settings::getBlocksWrite`. Elasticsearch returns all settings as strings and does not normalize bool values.
   The getters now return the right bool value for whichever string representation is used like 'true', '1', 'on', 'yes'.
 - Fix for QueryBuilder version check `\Elastica\QueryBuilder\Version\Version240.php` added all new query types to queries array.
+- Do not modify the original query in `\Elastica\Search::count`.
 
 ### Added
 
--- a/lib/Elastica/Search.php
+++ b/lib/Elastica/Search.php
@@ -475,7 +475,8 @@ class Search
     {
         $this->setOptionsAndQuery(null, $query);
 
-        $query = $this->getQuery();
+        // Clone the object as we do not want to modify the original query.
+        $query = clone $this->getQuery();
         $query->setSize(0);
         $path = $this->getPath();
 
--- a/test/Elastica/SearchTest.php
+++ b/test/Elastica/SearchTest.php
@@ -578,4 +578,20 @@ class SearchTest extends BaseTest
         $results = $search->search($query, [Search::OPTION_SEARCH_IGNORE_UNAVAILABLE => true]);
         $this->assertInstanceOf(ResultSet::class, $results);
     }
+
+    /**
+     * @group functional
+     */
+    public function testQuerySizeAfterCount()
+    {
+        $client = $this->_getClient();
+        $search = new Search($client);
+
+        $query = new Query(new MatchAll());
+        $query->setSize(25);
+
+        $search->count($query);
+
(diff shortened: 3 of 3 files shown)


z38:
It might be better to clone the query and keep the removed line. With the current fix, the first 10(?) results get transmitted over the wire.

ruflin:
@XWB Thanks for the PR. I think I understand the problem now better. It becomes a problem if count is used as the query itself is modified. Shouldn't we use a count query (https://www.elastic.co/guide/en/elasticsearch/reference/current/search-count.html) here if fullResult is not set? I'm curious why we introduce fullResult here anyways.

Currently at Elastic{ON} so will not have time this week to dig into this.

ruflin:
@XWB Had a quick look at it, and it seems we should use a _count query if fullResult is false. Or do I miss something here?

Tobion:
This is not the correct solution. Setting the size to 0 is an important optimization as also pointed out by @z38 and the main advantage of using the shortcut `count()` at all.
But I agree the query should not be modified. Please clone the query before setting the size and please also add a test for this.

Tobion:
@XWB do you have time to finish this?

XWB:
@Tobion Will do.

ruflin:
@XWB Can you also update the CHANGELOG.md?

XWB:
@ruflin @Tobion Done.

lib/Elastica/Search.php
```diff
@@ -475,7 +475,7 @@ public function count($query = '', $fullResult = false)
     {
         $this->setOptionsAndQuery(null, $query);
 
-        $query = $this->getQuery();
+        $query = clone $this->getQuery();
```
ruflin: Could you add a quick note (docs comment) on what it is cloned? Just to make sure nobody will remove it in the future ...

XWB:
@ruflin Done.

ruflin:
@Tobion Do you want to have a look before merging?

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file based on the
 - Fix reading bool index settings like `\Elastica\Index\Settings::getBlocksWrite`. Elasticsearch returns all settings as strings and does not normalize bool values.
   The getters now return the right bool value for whichever string representation is used like 'true', '1', 'on', 'yes'.
 - Fix for QueryBuilder version check `\Elastica\QueryBuilder\Version\Version240.php` added all new query types to queries array.
+- Do not modify the original query in `\Search::count`.
```
Tobion: `\Elastica\Search`

test/Elastica/SearchTest.php
```diff
@@ -578,4 +578,29 @@ public function testIgnoreUnavailableOption()
...
+    /**
+     * @group functional
+     */
+    public function testQuerySizeAfterCount()
+    {
+        $index = $this->_createIndex();
+        $search = new Search($index->getClient());
+        $type = $index->getType('test');
+
+        $doc = new Document(1, ['id' => 1, 'username' => 'ruflin']);
```
Tobion: No need to add a document since it's not relevant for this test.

XWB:
@Tobion Changes are underway.

XWB:
And done.

test/Elastica/SearchTest.php
```diff
@@ -578,4 +578,26 @@ public function testIgnoreUnavailableOption()
         $results = $search->search($query, [Search::OPTION_SEARCH_IGNORE_UNAVAILABLE => true]);
         $this->assertInstanceOf(ResultSet::class, $results);
     }
+
+    /**
+     * @group functional
+     */
+    public function testQuerySizeAfterCount()
+    {
+        $index = $this->_createIndex();
```
Tobion: you also don't need an index or a type for this test.

XWB:
@Tobion I simplified the test.

PR #1267, opened 2017-03-12
move test files to decrease directory depths

### LINKED ISSUE #715: Migrate file structure to PSR-4
See https://github.com/ruflin/Elastica/pull/714 for more details

Tobion:
The `JSONTest` and the rest of the test files were at different locations. I moved all the files to the shorter directory structure which decreases the depths.

Fixes #715 

Doing the same thing for the src is not worth it and will just complicate git history.


composer.json
```diff
@@ -13,11 +13,11 @@
     ],
     "require": {
         "php": ">=5.6.0",
-        "psr/log": "~1.0"
```
ruflin: Could we move the changes to these 3 lines to an other PR? I prefer not to mix the heavy moving of files with these "dependency" logic changes.

Tobion: removed

ruflin:
@Tobion Merged. Thanks.

PR #1272, opened 2017-03-15
Fix geo_bounds PR

Tobion:


test/Elastica/Aggregation/GeoBoundsTest.php
```diff
@@ -8,7 +8,7 @@
 
 class GeoBoundsTest extends BaseAggregationTest
 {
-    protected function _getIndexForTest()
+    private function getIndexForTest()
```
ruflin: That is an interesting discussion we should probably have. I was stubborn enough to keep Elastica with `_` for all private and protected function as I personally liked it. But I think in the meantime, most projects moved away from it?

Tobion: Yes. There is no point in adding that anymore since there are visibility modifiers.

ruflin: The reason I kept the `_` is not because of the modifiers, they exist for quite some time. I kept it for humans as it makes it very easy to see if a method is public or not. Similar to Golang that uses upper and lower case for it. As I said, I still like the `_` but I think also it makes sense to follow what other large projects are doing to have a "unified" code base.

What are others doing?

Tobion: I haven't seen PHP projects using underscore prefix in a long time.

ruflin: Welcome to Elastica ;-) Seems like we should make this switch across the product then. Perhaps we can split it up in 2 parts. First migrate all tests and the library. On the library side it will be obviously a BC break so we should do it as soon as we decide that we are potentially not going to do any further 5.x releases. ES 6.0.0-alpha1 should be out in a few weeks. That sounds like a good timing from my point of view.

Tobion: Sounds good. But IMO we should only combine it with some other changes. Only  renaming properties does not justify a BC break.

ruflin: @Tobion You mean to do it across the whole project? Or you would do it step by step and only if the methods have to be changed anyways?

For the overall, I'm quite confident we will have some in 6.0. And for the change I don't worry too much about the users, I mostly worry about other libraries like FOSElastica where this is potentially going to break things.

PR #1273, opened 2017-03-16
Keep track of neccessary scroll pages to skip useless last request

### LINKED ISSUE #1234: Unnecessary request to ES when iterating over ResultSets using Scroll
There is one unnecessary call to ES API when using Elastica\Scroll. The last call (in next method) always sets `$this->_currentResultSet` with empty Elastica\ResultSet. It could be omitted.
The solution is quite simple. In rewind method number of total pages could be count
`$this->totalPages = ceil($scroll->current()->getTotalHits() / $scroll->current()->count());`

Then in next method request should be sent only if there is next page `$this->currentPage <= $this->totalPages;`
In next method `$this->currentPage` should be incremented.

Tobion:
Fixes #1234

### DIFF
--- a/lib/Elastica/Scroll.php
+++ b/lib/Elastica/Scroll.php
@@ -32,12 +32,14 @@ class Scroll implements \Iterator
 
     /**
      * 0: scroll<br>
-     * 1: scroll id<br>
-     * 2: search type.
+     * 1: scroll id
      *
      * @var array
      */
-    protected $_options = [null, null, null];
+    protected $_options = [null, null];
+
+    private $totalPages = 0;
+    private $currentPage = 0;
 
     /**
      * Constructor.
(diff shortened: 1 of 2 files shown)


lib/Elastica/Scroll.php
```diff
@@ -32,12 +32,14 @@ class Scroll implements \Iterator
 
     /**
      * 0: scroll<br>
-     * 1: scroll id<br>
-     * 2: search type.
+     * 1: scroll id
      *
      * @var array
      */
-    protected $_options = [null, null, null];
```
ruflin: search type was never used?

Tobion: It got removed before already with es upgrade

Tobion:
Added changelog

ruflin:
@Tobion Thanks, merged.

PR #1276, opened 2017-03-23
Add MatchNone the inverse of MatchAll

nomoa:
This query is sometimes handy and I think Elastica should support it.


ruflin:
@nomoa Merged. Thanks for the contribution.

PR #1277, opened 2017-03-23
Use endpoints official client

ewgRa:
This PR start using Endpoints from official client.

There is still places where raw request used, but they require more work and can be done later, during some refactoring. For example official client for Bulk and similar multi requests require SerializerInterface, and there is research needed, can Elastica use Serializer from official client.

This PR cover most cases.


lib/Elastica/Index.php
```diff
@@ -406,7 +424,10 @@ public function addAlias($name, $replace = false)
 
         $data['actions'][] = ['add' => ['index' => $this->getName(), 'alias' => $name]];
 
-        return $this->getClient()->request($path, Request::POST, $data);
+        $endpoint = new Update();
```
ruflin: Do we also have to add the index name here?

lib/Elastica/Index.php
```diff
@@ -420,11 +441,10 @@ public function addAlias($name, $replace = false)
      */
     public function removeAlias($name)
     {
-        $path = '_aliases';
+        $endpoint = new \Elasticsearch\Endpoints\Indices\Alias\Delete();
+        $endpoint->setName($name);
```
ruflin: Do we also need the index name here? It was part of the request before ...

lib/Elastica/Node/Info.php
```diff
@@ -208,16 +208,14 @@ public function refresh(array $params = [])
...
-        $path = '_nodes/'.$this->getNode()->getId();
+        $endpoint = new \Elasticsearch\Endpoints\Cluster\Nodes\Info();
+        $endpoint->setNodeID($this->getNode()->getId());
 
         if (!empty($params)) {
-            $path .= '/';
-            foreach ($params as $param) {
-                $path .= $param.',';
-            }
+            $endpoint->setMetric(join(',', $params));
```
ruflin: Nice.

lib/Elastica/Index.php
ewgRa: No, it is in "$data".

Actually here we need to use https://github.com/elastic/elasticsearch-php/blob/master/src/Elasticsearch/Endpoints/Indices/Alias/Put.php, but I don't quickly find a way to work with "$replace" argument there. Any suggestion?

lib/Elastica/Index.php
ewgRa: No, it is not needed, since we call $this->requestEndpoint(...) it will add index name there.

lib/Elastica/Node/Info.php
ewgRa: Wow, I just checked official client code, "join" even not needed, we call setMetric with array, https://github.com/elastic/elasticsearch-php/blob/master/src/Elasticsearch/Endpoints/Cluster/Nodes/Info.php#L31 . Will change it at evening and looks like we haven't test for this case.

ewgRa:
I think "meta" is not needed. It is easy to find places that can be ported to Endpoints. Just check "usage" of "request" function in Type, Index and Client. For now mostly such places left in Client.

lib/Elastica/Node/Info.php
ruflin: Even better.Ping me when done.

lib/Elastica/Index.php
ruflin: Not sure I can follow the above comment?

ewgRa: I just mean that we need use Put endpoint there, but I don't find a way how to process $replace flag there. So, it can be tricky and need more time and can be done later. And asking maybe you see quick solution for this. So, we can do this later.

ewgRa:
Looks loke all done, ping @ruflin

lib/Elastica/Index.php
ruflin: Got it, also not sure how to deal with it. Lets merge it as is.

ruflin:
@ewgRa Merged. Thanks.

PR #1283, opened 2017-04-06
Add set collate functionality for phrase suggester

vgaldikas:


ruflin:
Thanks for the PR. This is getting interesting. I had a quick look to better understand how collate works and it requires a Template Query (https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-template-query.html) which is deprecated in 5.0 and replaced by a Search Template: https://www.elastic.co/guide/en/elasticsearch/reference/5.3/search-template.html

This makes the whole thing a little bit more complex then I expected as it is not just a standard query and we don't have support for template queries yet. Seems like we should add support for it first?

vgaldikas:
@ruflin Yes, this is what confused me. 

So should I create a separate issue for SearchTemplate, and have a go at cracking that first, and then come back to this? Or do the whole lot on this branch?

And in Regards to search template, should it still be -> SearchTemplate extends AbstractQuery, and then setCollate would take SearchTemplate as parameter? Or should it be a separate thing all together?

ruflin:
+1 on separating this out in a different Github issue / PR. I think the TemplateQuery would extend AbstractQuery. I'm not so sure about the Search Template as it seems to be separate request. 

The easier path definitively seems to be going with TemplateQuery first for the 5.x releases and potentially change it for 6.0 in case it is removed.

vgaldikas:
@ruflin ok, that sounds fair enough. i will have ago at this first, but my first impression is that implementing that may be way over my head, maybe once I start digging it will prove to be simpler then I imagined :)

ruflin:
@vgaldikas Keep me posted on this one. Best open a new github issue for it to have the discussions and questions directly there. Happy to help.

ruflin:
Closing this as it didn't have any movement for the past 6 year. I still think it would be interesting to get these features in eventually.

PR #1294, opened 2017-04-30
using official nginx image and use of php:7.1 instead of php:7.1-rc

p365labs:
This small PR will correct some configurations on docker images

- remove the use of ruflin/nginx:1.9 with the official version:latest.
- update the docker image for Travis for php:7.1. Updating the php:7.1-rc to php:7.1

### DIFF
--- a/env/nginx/Dockerfile
+++ b/env/nginx/Dockerfile
@@ -1,4 +1,4 @@
-FROM nginx:1.9
+FROM nginx:1.13
 MAINTAINER Nicolas Ruflin <spam@ruflin.com>
 
 # Copy config files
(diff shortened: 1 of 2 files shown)


env/nginx/Dockerfile
```diff
@@ -1,5 +1,4 @@
-FROM nginx:1.9
-MAINTAINER Nicolas Ruflin <spam@ruflin.com>
+FROM nginx
```
ruflin: I prefer to have a specific version in here instead of latest. I had the experience in the past that with latest a new version gets pulled automatically which can break the CI. As I want tests to be reproducible, a fixed version should be kept in.

p365labs: sounds good to me :) update to PR coming.

PR #1296, opened 2017-05-04
Use correct Put class for updating settings. Fixes #1295

### LINKED ISSUE #1295: [5.2] Index::setSettings broken?
It seems that in 5.2+ `setSettings` on an `Index` is broken. It tries to set the settings on the whole cluster instead of only the index.

In 5.1 this still works.

If I take a short look at the code it seems `Elasticsearch\Endpoints\Cluster\Settings\Put` is the wrong class and we should use `Elasticsearch\Endpoints\Indices\Settings\Put` instead?

nickygerritsen:
This fixes #1295 and adds a test that verifies this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ All notable changes to this project will be documented in this file based on the
 
 ### Bugfixes
 
+- Fix updating settings of an index. [#1296](https://github.com/ruflin/Elastica/pull/1296)
+
 ### Added
  - Parameter `filter_path` for response filtering (e.g. `$index->search($query, ['filter_path' => 'hits.hits._source'])`)
 ### Improvements
--- a/lib/Elastica/Index.php
+++ b/lib/Elastica/Index.php
@@ -7,7 +7,7 @@ use Elastica\Index\Settings as IndexSettings;
 use Elastica\Index\Stats as IndexStats;
 use Elastica\ResultSet\BuilderInterface;
 use Elasticsearch\Endpoints\AbstractEndpoint;
-use Elasticsearch\Endpoints\Cluster\Settings\Put;
+use Elasticsearch\Endpoints\Indices\Settings\Put;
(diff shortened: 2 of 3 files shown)


ewgRa:
Strange, that another tests in this SettingsTest class are written in a same way, but they do not fail...

I haven't ability to check it right now, but did your test fail, if you revert changes in "lib/Elastica/Index.php" back?

Maybe there is wrong tests if it is not fail...

nickygerritsen:
Yeah my test failed before. The other tests in this class are different: they use `getSetting` and then `set` on that object instead of directly using `setSettings` on the index. The later is what was broken (and not tested at all yet).

ewgRa:
@nickygerritsen Indeed :+1: 

ping @ruflin

ruflin:
@nickygerritsen Could you put a note in the CHANGELOG.md file?

nickygerritsen:
Done!

ruflin:
@nickygerritsen Thanks for the fix
@nickygerritsen @ewgRa Do we need some follow up PR's to fix more stuff?

ewgRa:
@ruflin I don't think so, it was kind of typo, and fully fixed and covered by test now.

Tobion:
@ruflin can you make a patch release as this bug could affect many people?

ruflin:
@Tobion Yes, that is possible. Put it on my TODO list for next week. Hope that works.

mweibel:
@ruflin +1 on the patch release. Affects us right now :) Thanks! 👍

ruflin:
Here you go: https://github.com/ruflin/Elastica/releases/tag/5.2.1

mweibel:
Thanks! I noticed 5.2.1 has a BC break if you override `Elastica\Client::request()`

see: https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1258

ruflin:
Oh, good point :-( Should we add this to the CHANGELOG? Thanks for fixing it in the bundle.

PR #1297, opened 2017-05-05
refactor: variable $_config contain the key 'servers' will much more …

matrixstylite:
…readable

that can indicate user who can config servers without to look for source code
without that see the source code on line 127 on file lib/Elastica/Client.php, I am as a user not sure if there is some code to parse servers config info


ruflin:
Servers was removed / deprecated a long time ago (2012-12-23). It's still in the code for BC reasons but TBH should be removed in the next major release. Use `connections` instead.

PR #1298, opened 2017-05-09
Docker config improvements, properly create an Elasticsearch cluster and share the FS to be able to test SNAPSHOTS

p365labs:
This PR solves many problems : 

- now Docker is configured to properly create an Elasticsearch Cluster
- I've tweak some ES configuration to make ES create a cluster properly
- I've removed a container used just only the filesystem sharing, as it is possibile
to share a volume from the docker-compose file
- now the 2 Elasticsearch nodes shares a filesystem on which we can test snapshot generation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,10 +13,23 @@ env:
   - TARGET="71"
 
 before_install:
-  # Docker-compose installation
-  - curl -L https://github.com/docker/compose/releases/download/1.8.1/docker-compose-`uname -s`-`uname -m` > docker-compose
+  # list docker-engine versions, usefull to update the docker engine on travis and trusty
+  #- apt-cache madison docker-engine
+
+  # update docker engine to the desired version
+  - sudo apt-get -o Dpkg::Options::="--force-confnew" install -y docker-engine=1.13.1-0~ubuntu-trusty
+
+  - sudo rm /usr/local/bin/docker-compose
+  - curl -L https://github.com/docker/compose/releases/download/1.13.0/docker-compose-`uname -s`-`uname -m` > docker-compose
   - chmod +x docker-compose
   - sudo mv docker-compose /usr/local/bin
+
+  # check running "docker engine" and "docker-compose" version on travis
+  - docker --version
+  - docker-compose --version
+
+before_script:
+  - sudo sysctl -w vm.max_map_count=262144
 
 script:
   - make tests TARGET=$TARGET
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ endif
 
 .PHONY: clean
 clean:
+	docker-compose down -v
 	rm -r -f ./build
 	rm -r -f ./vendor
 	rm -r -f ./composer.lock
@@ -57,16 +58,16 @@ phpunit:
 tests:
 	# Rebuild image to copy changes files to the image
 	make elastica-image
-	make setup
+	make start
 	mkdir -p build
-	docker-compose run elastica make phpunit
-	docker cp elastica_elastica_run_1:/elastica/build/coverage/ $(shell pwd)/build/coverage
+	docker run --network=elastica_esnet elastica_elastica  make phpunit
+	docker cp elastica:/elastica/build/coverage/ $(shell pwd)/build/coverage
 
 # Makes it easy to run a single test file. Example to run IndexTest.php: make test TEST="IndexTest.php"
 .PHONY: test
 test:
 	make elastica-image
-	make setup
+	make start
 	mkdir -p build
 	docker-compose run elastica phpunit -c test/ ${TEST}
 
@@ -93,15 +94,9 @@ phploc:
 build:
 	docker-compose build
 
-.PHONY: setup
-setup: build
-	#docker-compose scale elasticsearch=3
-	# TODO: Makes the snapshot directory writable for all instances. Nicer solution needed.
-	#docker-compose run elasticsearch chmod -R 777 /tmp/backups/
-
 .PHONY: start
 start:
-	docker-compose up
+	docker-compose up -d
 
 .PHONY: stop
 stop:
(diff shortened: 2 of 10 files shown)


.travis.yml
```diff
@@ -14,9 +14,20 @@ env:
 
 before_install:
   # Docker-compose installation
-  - curl -L https://github.com/docker/compose/releases/download/1.8.1/docker-compose-`uname -s`-`uname -m` > docker-compose
-  - chmod +x docker-compose
-  - sudo mv docker-compose /usr/local/bin
+  - sudo apt-get update
+  - sudo apt-get install -o Dpkg::Options::="--force-confold" --force-yes -y docker-engine
```
ruflin: Can you add a comment in the code on what this is needed?

.travis.yml
```diff
@@ -14,9 +14,20 @@ env:
...
+  - docker-compose --version
```
ruflin: I would like to set this to a specific version as changes then to break related to docker releases :-)

Makefile
```diff
@@ -95,7 +96,7 @@ build:
 
 .PHONY: setup
 setup: build
-	#docker-compose scale elasticsearch=3
+	docker-compose up -d
```
ruflin: This already starts the service. I would not expect this from `setup`. Does this belong into `start`

.travis.yml
```diff
@@ -14,9 +14,20 @@ env:
...
+  - docker --version
+
+before_script:
+#  - composer self-update
+#  - composer install
+##  - docker-compose build
+  - sudo sysctl -w vm.max_map_count=262144
+  - docker-compose up -d
```
ruflin: could we use the makefile commands here?

docker-compose.yml
```diff
@@ -1,35 +1,84 @@
...
-  #image: ruflin/elastica
-  ports:
-    - "9200:9200"
-  links:
-    - nginx
-    - elasticsearch
-  environment:
-    - ES_HOST=elasticsearch
-    - PROXY_HOST=nginx
+version: '2.0'
```
ruflin: Glad to be on version 2.0

test/Elastica/Base.php
```diff
@@ -123,7 +123,7 @@ protected function _createIndex($name = null, $delete = true, $shards = 1)
 
         $client = $this->_getClient();
         $index = $client->getIndex('elastica_'.$name);
-        $index->create(['index' => ['number_of_shards' => $shards, 'number_of_replicas' => 0]], $delete);
+        $index->create(['index' => ['number_of_shards' => $shards, 'number_of_replicas' => 2]], $delete);
```
ruflin: As you have only 2 instances of elasticsearch, this should probably be set to 1 and not 2?

.travis.yml
p365labs: yep!  the docker-compose up -d has been removed from travis now is in the make file

.travis.yml
p365labs: in this case docker-compose is already installed on travis, and I updated to the latest docker version as here : https://docs.travis-ci.com/user/docker/#Installing-a-newer-Docker-version

p365labs: if u prefer having a version I should revert the updates. maybe using docker-compose 1.8.X, and the default docker-engine

docker-compose.yml
p365labs: 👍

test/Elastica/Base.php
p365labs: that's a good point don't know why I choose 2 :) I'm executing tests again on my local dev-machine than I will push it to remote

Makefile
p365labs: mmm ok but *start* is never called in the Makefile. If you want the *docker-compose up -d* inside the *start* I should change also *tests* and *test* the have inside **make setup** with **make start**. How does it sounds ?

.travis.yml
p365labs: ok @ruflin  as u suggested I reverted some old code which installed a specific version of docker-compose I just upgraded to a more recent version, and I find a way to update the docker engine to a fixed version *1.13.x.* 
This would allow us to use in future docker-compose file version **3.0** and **3.2**

Makefile
p365labs: @ruflin moved to start

.travis.yml
p365labs: @ruflin see the comment below: I get the previous code for docker-compose install ***back***

.travis.yml
ruflin: That is great. I would really like to use a more recent version, but have a fixed version.

Makefile
```diff
@@ -95,13 +96,11 @@ build:
 
 .PHONY: setup
 setup: build
-	#docker-compose scale elasticsearch=3
-	# TODO: Makes the snapshot directory writable for all instances. Nicer solution needed.
-	#docker-compose run elasticsearch chmod -R 777 /tmp/backups/
+	#docker-compose up -d
```
ruflin: Seem slike we can remove `setup` now completely?

Makefile
```diff
@@ -10,6 +10,7 @@ endif
 
 .PHONY: clean
 clean:
+	docker-compose down
```
ruflin: Nice, that also removes the network. Should we add `-v` to also remove the volumes?

env/data/Dockerfile
```diff
@@ -1,9 +0,0 @@
-# Data image for elasticsearch instances
```
ruflin: Happy that we got rid of it.

.travis.yml
```diff
@@ -13,11 +13,24 @@ env:
   - TARGET="71"
 
 before_install:
-  # Docker-compose installation
-  - curl -L https://github.com/docker/compose/releases/download/1.8.1/docker-compose-`uname -s`-`uname -m` > docker-compose
+  # list docker-engine versions, usefull to update the docker engine on travis and trusty
+  #- apt-cache madison docker-engine
+
+  # update docker engine to the desired version
+  - sudo apt-get -o Dpkg::Options::="--force-confnew" install -y docker-engine=1.13.1-0~ubuntu-trusty
```
p365labs: this way you can still update the docker engine as docker compose

p365labs: another thing I squashed everything in one commit, I think it would be better

.travis.yml
p365labs: checkout the new commit now u can set the docker version

ruflin:
Thanks a lot for this PR. I know it's quite a bit of work to get this all right and green.

PR #1299, opened 2017-05-09
lib/Elastica/Search.php should not be executable

jbboehr:


ruflin:
Does this only apply to `Search.php`. Did you check if other files have the same issue?

jbboehr:
That was the only file that was executable that shouldn't have been:

```
$ find -executable -type f -not -wholename '*/.git/*'
./docker-entrypoint.sh
./lib/Elastica/Search.php
./env/elasticsearch/docker-entrypoint.sh

```

ruflin:
Merged. Thanks for checking and fixing this.

PR #1302, opened 2017-05-11
Always set Content-Type with Http Transport

### LINKED ISSUE #1301: Elastica should set the Content-Type header when sending json via Http
Not setting `Content-Type` header in the request headers will now produce (elasticsearch 5.3) a deprecation warning in elastic logs.
I think that Elastica should set this header when sending json.
The deprecation warning is :
`[2017-05-11T13:40:27,556][WARN ][o.e.d.r.RestController   ] Content type detection for rest requests is deprecated. Specify the content type using the [Content-Type] header.`

nomoa:
Added a new param to Client::request and Request::_construct to allow
setting the content-type of the data being sent.
This defaults to "application/json" and is only overidden by Bulk and MultiSearch
where it should be application/x-ndjson.

Sadly I'm not sure how to test this properly.
I'm not a big a big fan of adding a new param like that. Ideally we should refactor
the $data param of Request::_construct to accept a array (defaults to application/json)
or a new RequestBody object where the client could set the content-type alongside the
the string.

closes #1301

### DIFF
--- a/lib/Elastica/Request.php
+++ b/lib/Elastica/Request.php
@@ -15,6 +15,8 @@ class Request extends Param
     const PUT = 'PUT';
     const GET = 'GET';
     const DELETE = 'DELETE';
+    const DEFAULT_CONTENT_TYPE = 'application/json';
+    const NDJSON_CONTENT_TYPE = 'application/x-ndjson';
 
     /**
      * @var \Elastica\Connection
@@ -29,10 +31,11 @@ class Request extends Param
      * @param array      $data       OPTIONAL Data array
      * @param array      $query      OPTIONAL Query params
      * @param Connection $connection
+     * @param string     $contentType Content-Type sent with this request
      *
      * @return \Elastica\Request OPTIONAL Connection object
      */
-    public function __construct($path, $method = self::GET, $data = [], array $query = [], Connection $connection = null)
+    public function __construct($path, $method = self::GET, $data = [], array $query = [], Connection $connection = null, $contentType = self::DEFAULT_CONTENT_TYPE )
     {
         $this->setPath($path);
         $this->setMethod($method);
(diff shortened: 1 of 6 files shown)


lib/Elastica/Request.php
```diff
@@ -159,6 +163,24 @@ public function getConnection()
     }
 
     /**
+     * Set the Content-Type of this request
+     * @param string $contentType
+     */
+    public function setContentType($contentType)
+    {
+        $this->setParam('contentType', $contentType);
```
ruflin: `setParam(..)` should already return $this so you can put the return in front of the call directly.

lib/Elastica/Request.php
```diff
@@ -29,10 +31,11 @@ class Request extends Param
      * @param array      $data       OPTIONAL Data array
      * @param array      $query      OPTIONAL Query params
      * @param Connection $connection
+     * @param string     $contentType Content-Type sent with this request
      *
      * @return \Elastica\Request OPTIONAL Connection object
      */
-    public function __construct($path, $method = self::GET, $data = [], array $query = [], Connection $connection = null)
+    public function __construct($path, $method = self::GET, $data = [], array $query = [], Connection $connection = null, $contentType = self::DEFAULT_CONTENT_TYPE )
```
nomoa: Ideally we should no longer accept a String as the $data object.
Giving an array it'll be unambiguously 'application/json' but giving a string we don't know how to decide between json and the bulk format.
Creating a new class like RequestBody and enforce clients to use it when they plan to send strings directly might be a cleaner approach imo.

ruflin: As the header will be required in 6.0 and alpha1 just came out, I think our tests will hopefully show very quickly if we have to fix some more things here. For the moment I think this fixes the most urgent issue.

lib/Elastica/Request.php
nomoa: done

ruflin:
@nomoa Merged. Thanks a lot for the fix.

PR #1303, opened 2017-05-12
Fix a problem with Travis build, installing libpcre to avoid failing …

p365labs:
It looks like that from yesterday evening the build is failing due to a lack of ***libpcre3-dev*** in the docker container with php 5.6. I'm still investigating what happened, I suspect something changed in the *debian:jessie* container. But I'm still inspecting, in the mean time this problem should be fixed installing the lib in the container.

### DIFF
--- a/env/elastica/Docker56
+++ b/env/elastica/Docker56
@@ -9,6 +9,7 @@ RUN apt-get update && apt-get install -y \
 	graphviz \
 	nano \
 	php5-xsl \
+	libpcre3-dev \
 	zlib1g-dev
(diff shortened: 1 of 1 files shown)


p365labs:
I found maybe some more info [https://github.com/docker-library/php/commit/6844e717a56a5dd8ad87a236a96bea069cc635fd](https://github.com/docker-library/php/commit/6844e717a56a5dd8ad87a236a96bea069cc635fd), on the official DOCKER php56 Dockerfile

ruflin:
Interesting. Thanks for the fix.

PR #1306, opened 2017-05-12
Functional test in HasParentTest intermittently fails due to a lack in ES configuration

p365labs:
The problem with the docker image from ES version 5.2.2 is that is shipped with x-pack installed, I disabled security but I forgot to disable monitor which sometimes make this test failing. 

At this time it's not possible to remove completely the xpack plugin (except with a workaround) due
to an issue in the official docker ES image : [https://github.com/elastic/elasticsearch-docker/issues/36](https://github.com/elastic/elasticsearch-docker/issues/36). At the moment I've disabled *security* and *monitoring* from the elasticsearch.yml. The problem has been resolved (maybe) in the 5.4, my suggestion is to completely remove the plugin when we will switch to that version.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,6 +18,7 @@ services:
       - bootstrap.memory_lock=true
       - network.host=172.18.0.4
       - xpack.security.enabled=false
+      - xpack.monitoring.enabled=false
       - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
     ulimits:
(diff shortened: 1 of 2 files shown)


ruflin:
The alternative would be to enable it for all tests and use the default username / password. Like this we also test the authentication. This will probably work well in 5.x. But in 6.0 probably random passwords will be introduced which will make this much more complex.

Issue #1287, opened 2017-04-13
Support for parent id query missing


mfn:
Docs: https://www.elastic.co/guide/en/elasticsearch/reference/5.3/query-dsl-parent-id-query.html

AFAICS there's nothing supporting this in Elastica.

E.g. `new ParentId(string $type, $id, bool $ignoreUnmapped = false)`

ruflin:
Yes, this query is currently missing. Interested to add is a contribution?

mfn:
Fixed in https://github.com/ruflin/Elastica/pull/1308

Thanks 😄

PR #1308, opened 2017-05-15
Added support to parent_id query

giovannialbero1992:
added ParentId Query by reference [#1287](https://github.com/ruflin/Elastica/issues/1287)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ All notable changes to this project will be documented in this file based on the
  - Parameter `filter_path` for response filtering (e.g. `$index->search($query, ['filter_path' => 'hits.hits._source'])`)
  - Add support for Health parameters for Cluster\Health endpoint (new prop : delayed_unassigned_shards, number_of_pending_tasks, number_of_in_flight_fetch, task_max_waiting_in_queue_millis, active_shards_percent_as_number)
  - Add support for querystring in Type. this allow to use `update_all_types` in type mapping in order to resolve conflicts between fields in different types. [Conflicts between fields in different types](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-put-mapping.html#merging-conflicts) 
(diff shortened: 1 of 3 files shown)


lib/Elastica/Query/ParentId.php
```diff
@@ -0,0 +1,27 @@
+<?php
+namespace Elastica\Query;
+
+final class ParentId extends AbstractQuery
```
ruflin: Why did you add final here?

CHANGELOG.md
```diff
@@ -13,7 +13,7 @@ All notable changes to this project will be documented in this file based on the
 ### Added
 
  - Parameter `filter_path` for response filtering (e.g. `$index->search($query, ['filter_path' => 'hits.hits._source'])`)
-
+ - Added `\Elastica\Query\ParentId` to avoid join with parent documents [#1287](https://github.com/ruflin/Elastica/issues/1287)
```
ruflin: Could also add here a newline before the next title?

lib/Elastica/Query/ParentId.php
giovannialbero1992: I usually apply `final` by default to classes that are not specifically designed to be extended. 
This avoids breaking an eventual child class when refactoring this class.
If you want i can remove `final` from this class.

ruflin: I would prefer to remove it as we don't have it in any other files and I don't see a reason, why we should prevent someone extendings this class if we wants to (even though it is unlikely to happen).

giovannialbero1992: Ok, I've removed final

ruflin:
@giovannialbero1992 Seems like you have changelog conflict :-(

giovannialbero1992:
@ruflin done! ;-)

ruflin:
@giovannialbero1992 Thanks. merged.

PR #1313, opened 2017-05-19
Adding the missing Content-Type key in the headers for ES 5.3+

pengyue:
Refactor on the Transport/Http.php would be better, this is a quick fix for https://github.com/nomoa/Elastica/commit/c8896cec23d23d92fdeb0c5fcf1b03ee0d7c4cb7


nomoa:
@pengyue thanks for catching this!

ruflin:
@pengyue Thanks for the fix.

Issue #1311, opened 2017-05-17
Deprecate and remove CrossIndex Feature


ruflin:
Elasticsearch now supports reindex which makes the crossindex feature obsolote. The feature should be deprecated / removed and potential support for reindex on the elastica side should be added.

giovannialbero1992:
I'll do a PR on this in this days ! :)

PR #1315, opened 2017-05-21
Added Reindex and deprecated CrossIndex

giovannialbero1992:
[#1311](https://github.com/ruflin/Elastica/issues/1311)

### DIFF
--- a/lib/Elastica/Reindex.php
+++ b/lib/Elastica/Reindex.php
@@ -1,0 +1,139 @@
+<?php
+namespace Elastica;
+
+use Elastica\Query\AbstractQuery;
+
+class Reindex
+{
+    const VERSION_TYPE = 'version_type';
+    const VERSION_TYPE_INTERNAL = 'internal';
+    const VERSION_TYPE_EXTERNAL = 'external';
+    const OPERATION_TYPE = 'op_type';
+    const OPERATION_TYPE_CREATE = 'create';
+    const CONFLICTS = 'conflicts';
+    const CONFLICTS_PROCEED = 'proceed';
+    const TYPE = 'type';
+    const SIZE = 'size';
+    const QUERY = 'query';
+
+    /**
+     * @var Index
+     */
+    protected $_oldIndex;
+
+    /**
+     * @var Index
+     */
+    protected $_newIndex;
+
+    /**
+     * @var array
+     */
+    protected $_options;
+
+    public function __construct(Index $oldIndex, Index $newIndex, array $options = [])
+    {
+        $this->_oldIndex = $oldIndex;
+        $this->_newIndex = $newIndex;
+        $this->_options = $options;
+    }
+
+    public function run()
+    {
+        $body = $this->_getBody($this->_oldIndex, $this->_newIndex, $this->_options);
+
+        $reindexEndpoint = new \Elasticsearch\Endpoints\Reindex();
+        $reindexEndpoint->setBody($body);
+
+        $this->_oldIndex->getClient()->requestEndpoint($reindexEndpoint);
+        $this->_newIndex->refresh();
+
+        return $this->_newIndex;
+    }
+
+    protected function _getBody($oldIndex, $newIndex, $options)
+    {
+        $body = array_merge([
+            'source' => $this->_getSourcePartBody($oldIndex, $options),
+            'dest' => $this->_getDestPartBody($newIndex, $options)
+        ], $this->_resolveBodyOptions($options));
+
+        return $body;
+    }
+
+    protected function _getSourcePartBody(Index $index, array $options)
+    {
+        $sourceBody = array_merge([
+            'index' => $index->getName(),
+        ], $this->_resolveSourceOptions($options));
+
+        $sourceBody = $this->_setSourceQuery($sourceBody);
+        $sourceBody = $this->_setSourceType($sourceBody);
+
+        return $sourceBody;
+    }
+
+    protected function _getDestPartBody(Index $index, array $options)
+    {
+        return array_merge([
+            'index' => $index->getName(),
+        ], $this->_resolveDestOptions($options));
+    }
+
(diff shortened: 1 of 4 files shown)


lib/Elastica/Reindex.php
```diff
@@ -0,0 +1,117 @@
...
+    const CONFLICTS_PROCEED = 'proceed';
+    const TYPE = 'type';
+    const SIZE = 'size';
+    const QUERY = 'query';
+
+    public static function reindex(Index $oldIndex, Index $newIndex, array $options = [])
+    {
+        $body = self::getBody($oldIndex, $newIndex, $options);
+
+        $reindexEndpoint = new \Elasticsearch\Endpoints\Reindex();
```
ruflin: glad to see you used the endpoints 👍

lib/Elastica/Reindex.php
```diff
@@ -0,0 +1,117 @@
...
+    const VERSION_TYPE_EXTERNAL = 'external';
+    const OPERATION_TYPE = 'op_type';
+    const OPERATION_TYPE_CREATE = 'create';
+    const CONFLICTS = 'conflicts';
+    const CONFLICTS_PROCEED = 'proceed';
+    const TYPE = 'type';
+    const SIZE = 'size';
+    const QUERY = 'query';
+
+    public static function reindex(Index $oldIndex, Index $newIndex, array $options = [])
```
ruflin: Why did you use static method here? I'm thinking the implementation would become simpler if you would use a Reindex object where old, new and options are passed.

giovannialbero1992: I've used static method because there are not external depenendencies but maybe, to be compliant with OOP I should remove static. Do you suggest to pass parameters into constructor or in reindex method ?

ruflin: I was thinking passing params to constructor and then having something like `$reindex->Run()` or `Start` or something similar. Perhaps Start() could also take additional options. Not 100% sure yet how the most common use case will look like. Any thoughts?

giovannialbero1992: I think that call `new` and `->run()` (or `->start()`) are ever near to constructor and there is no need to change options after have called `new Reindex(...)`. Maybe in a first time we can see how community uses reindex and if there will be need to change the use, we will change it consequently.

giovannialbero1992:
@ruflin I've removed static from class and renamed method in `run`. What do you think about this version?

lib/Elastica/Reindex.php
```diff
@@ -0,0 +1,139 @@
...
+    const CONFLICTS = 'conflicts';
+    const CONFLICTS_PROCEED = 'proceed';
+    const TYPE = 'type';
+    const SIZE = 'size';
+    const QUERY = 'query';
+
+    /**
+     * @var Index
+     */
+    private $oldIndex;
```
ruflin: We always use $_oldIndex for all the variable names which are private / protected but perhaps the time has come to break this as not much other projects are left with the old pattern.

lib/Elastica/Reindex.php
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+     * @var Index
+     */
+    private $newIndex;
```
ruflin: Please make these all protected and not private. In case someone wants to extend Reindex in his code, he should be able to.

lib/Elastica/Reindex.php
```diff
@@ -0,0 +1,139 @@
...
+        $reindexEndpoint = new \Elasticsearch\Endpoints\Reindex();
+        $reindexEndpoint->setBody($body);
+
+        $this->oldIndex->getClient()->requestEndpoint($reindexEndpoint);
+        $this->newIndex->refresh();
+
+        return $this->newIndex;
+    }
+
+    private function getBody($oldIndex, $newIndex, $options)
```
ruflin: Make these also protected, not private.

giovannialbero1992:
@ruflin, I've changed it :)

lib/Elastica/Reindex.php
giovannialbero1992: I changed to be compliant with old pattern... maybe we can change the pattern into another commit on all protected/private variables :)

lib/Elastica/Reindex.php
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+     * @var Index
+     */
+    protected $_newIndex;
+
+    /**
+     * @var array
+     */
+    private $_options;
```
ruflin: This one should also be protected.

lib/Elastica/Reindex.php
```diff
@@ -0,0 +1,139 @@
...
+        $reindexEndpoint = new \Elasticsearch\Endpoints\Reindex();
+        $reindexEndpoint->setBody($body);
+
+        $this->_oldIndex->getClient()->requestEndpoint($reindexEndpoint);
+        $this->_newIndex->refresh();
+
+        return $this->_newIndex;
+    }
+
+    protected function getBody($oldIndex, $newIndex, $options)
```
ruflin: This would be _getBody(

lib/Elastica/Reindex.php
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    private function getDestPartBody(Index $index, array $options)
+    {
+        return array_merge([
+            'index' => $index->getName(),
+        ], $this->resolveDestOptions($options));
+    }
+
+    private static function resolveSourceOptions(array $options)
```
ruflin: This should also be not static and start with _* Also make it protected if there is no need to make it private.

giovannialbero1992:
@ruflin I've changed all names with '_' and removed a missing static from method.
`_getDestPartBody`, `_getSourcePartBody` and `_getBody` now are protected. Thanks for your reviews :)

ruflin:
@giovannialbero1992 Thanks for your contribution.

Issue #565, opened 2014-03-05
Type->deleteById() throws NotFoundException for missing documents


rayward:
Just wanted to initiate some discussion around whether this method should consider a missing document an error.

https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Type.php#L423

In my opinion, its not an error and shouldn't throw the `NotFoundException`. Trying to delete a document that doesn't exist still ends up with the same result, the document no longer exists. This seems fine to me.

Elastic doesn't consider it an error as it doesn't populate the `error` field. For comparison, neither does MySQL. It will say `Query OK, 0 rows affected (0.00 sec)`.

Also, there's an inconsistency with the Bulk delete, as it doesn't generate any exceptions for missing documents.

If someone wants to know whether the document did exist and was actually deleted then they can read the response data and check the `found` element themselves.

tastendruecker:
I just stumbled over this problem and would like to pick up the discussion again. I agree with rayward: As the result is the same, no exception should be thrown. My suggestion would be to remove the corresponding block.
@ruflin Would that be okay with you? Would you remove that code?

ruflin:
I'm ok with changing it, but I would like to have still a way to identify if the doc was actually deleted or not. This could be through a bool or an int that is returned couting how many docs were deleted. Feel free to open a PR for further discussions.

tastendruecker:
Isn't the returned $response object supposed to carry that information? (Please bare with me as I haven't much experience with Elastica. I just depend on using it within a project that had been designed before I entered the project team.)

ruflin:
@tastendruecker You are probably right. I'm not sure what the exact response will be. Can you open a PR? Then we can directly check it in the tests :-)

PR #1317, opened 2017-05-30
Type.php: Removed Exception thrown in deleteById()

tastendruecker:
An exception was thrown if no data set corresponding to param $id was found (and hence was not deleted). Removed this behavior because it is a very harsh reaction to a more or less uncritical situation because the result is the same: The data requested to be deleted no longer exists. The information wether or not a record set was deleted is supposed to be part of the returned $response.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,8 @@ All notable changes to this project will be documented in this file based on the
 
 ### Improvements
 
- - Added support for `other_bucket` and `other_bucket_key` paramters on `Elastica\Aggregation\Filters`
+ - Added support for `other_bucket` and `other_bucket_key` parameters on `Elastica\Aggregation\Filters`
+ - Removed NotFoundException thrown in `\Elastica\Type::deleteById` [#565](https://github.com/ruflin/Elastica/issues/565)
 
 ### Deprecated
  - Deprecated `Tool\CrossIndex` use `\Elastica\Reindex` instead [#1311](https://github.com/ruflin/Elastica/issues/1311)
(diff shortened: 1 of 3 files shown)


ruflin:
I remember we discussed this in an other PR / Issue recently. Could you link to the issue?

Also this requires a CHANGELOG entry as it is a breaking change and probably you have to adjust the tests.

tastendruecker:
Hi @ruflin! We discussed this enhancement in [this issue](https://github.com/ruflin/Elastica/issues/565).

I'll have a look at the tests.

test/Elastica/TypeTest.php
```diff
@@ -266,27 +266,6 @@ public function testDeleteById()
             $this->assertTrue(true);
         }
 
-        try {
```
ruflin: Instead of removing the tests, could we add one that it now works as expected?

CHANGELOG.md
```diff
@@ -20,7 +20,8 @@ All notable changes to this project will be documented in this file based on the
 
 ### Improvements
 
- - Added support for `other_bucket` and `other_bucket_key` paramters on `Elastica\Aggregation\Filters`
+ - Added support for `other_bucket` and `other_bucket_key` parameters on `Elastica\Aggregation\Filters`
+ - Removed NotFoundException thrown in `\Elastica\Type::deleteById` [#565](https://github.com/ruflin/Elastica/issues/565)
```
ruflin: This is actually a breaking change.

ruflin:
@tastendruecker Any update on this one?

thePanz:
Implemented in #1732 
Closing this one

PR #1319, opened 2017-06-01
Add span queries

### LINKED ISSUE #1316: Missing support for span queries
`span_*` queries are currently unimplemented in Elastica; this is evident looking into the QueryBuilder class which throw a `NotImplementedException` for every `span_` method.

Documentation on the span queries can be found here: https://www.elastic.co/guide/en/elasticsearch/reference/current/span-queries.html

alekitto:
Fixes #1316 

- [x] span_term
- [x] span_multi
- [x] span_first
- [x] span_near
- [x] span_or
- [x] span_containing
- [x] span_within
- [x] span_not
- [ ] field_masking_span

### DIFF
--- a/lib/Elastica/Query/SpanContaining.php
+++ b/lib/Elastica/Query/SpanContaining.php
@@ -1,0 +1,49 @@
+<?php
+namespace Elastica\Query;
+
+/**
+ * SpanContaining query.
+ *
+ * @author Alessandro Chitolina <alekitto@gmail.com>
+ *
+ * @link https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-span-containing-query.html
+ */
+class SpanContaining extends AbstractSpanQuery
+{
+    /**
+     * Constructs a SpanContaining query object.
+     *
+     * @param AbstractSpanQuery $little OPTIONAL
+     * @param AbstractSpanQuery $big OPTIONAL
+     */
+    public function __construct(AbstractSpanQuery $little = null, AbstractSpanQuery $big = null)
+    {
+        if (null !== $little) {
+            $this->setLittle($little);
+        }
+
+        if (null !== $big) {
+            $this->setBig($big);
+        }
+    }
+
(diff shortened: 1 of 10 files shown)


ruflin:
https://github.com/ruflin/Elastica/pull/1320 is now merged. This probably needs rebase now.

alekitto:
I've rebased and squashed my work.
The only thing left is the `field_masking_span`, but it's barely documented (there's only an example), so i think we could leave this out, as the implementation of the query class would be probably wrong.

lib/Elastica/Query/SpanContaining.php
```diff
@@ -0,0 +1,45 @@
...
+            $this->setBig($big);
+        }
+    }
+
+    /**
+     * @param AbstractSpanQuery $little
+     */
+    public function setLittle(AbstractSpanQuery $little)
+    {
+        $this->setParam('little', $little);
```
ruflin: We normal have a `return $this->setParam('little', $little);` in here so all setters return the object itself. This allows for chaining functions.

alekitto:
Ok, fixed all the setters in SpanContaining, SpanNot, SpanWithin and SpanNear (missed in #1320)

ruflin:
@alekitto In the diff Builder.php shows up and I'm not sure why?

ruflin:
BTW: Could you also add an entry to the changelog?

alekitto:
Builder.php is an empty file, I probably deleted it without noticing.
The Changelog already contains a line for Span* queries addition. Do I have to add another one for these three?

ruflin:
Argh, it seems some things got mixed up in the Changelog :-( The reason an additional entry is needed because just yesterday happened a release. So this goes into the next release.

ruflin:
I just pushed a change to the CHANGELOG to bring it up-to-date. Make sure to get the most recent one in before updating it.

alekitto:
👍 Done.

PR #1324, opened 2017-06-08
WIP: Expose getNumberOfReplicas() from index settings

thePanz:
The index settings exposes most of the available settings, while for the numberOfReplicas the "get" method is missing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file based on the
 ### Bugfixes
 
 ### Added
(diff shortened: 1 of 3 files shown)


ruflin:
LGTM. Could you add a changelog entry?

thePanz:
@ruflin Changelog added, plus:
- fixed a typo in the test name (it was "shards" instead of "replicas")
- removed usage of `empty`, as we can set `0` (zero) replicas

ruflin:
Thanks, merged.

PR #1325, opened 2017-06-09
Send scroll_id inside a json rather that text

nomoa:
It's likely to cause issues with elastic 5.3 as Content-Type handling becomes
more and more strict.

eg.:
curl -H"Content-Type: application/json"  -XGET localhost:9200/_search/scroll?scroll=1m -d 'DXF...'
now fails with Failed to parse request body on 5.3.x.
Elastic 5.1 was lenient and accepted this body and content-type.
Since we default Content-Type to application/json this is likely cause issues
in Elastica as well.
Using an array instead of plain string should force the use of a JSON body which is coherent
with the content-type we set.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ All notable changes to this project will be documented in this file based on the
 
 ### Bugfixes
 
+- Send the `scroll_id` inside a json body instead of plain text [#1325](https://github.com/ruflin/Elastica/pull/1325)
+
 ### Added
 
 ### Improvements
--- a/lib/Elastica/Search.php
+++ b/lib/Elastica/Search.php
@@ -451,7 +451,7 @@ class Search
 
         // Send scroll_id via raw HTTP body to handle cases of very large (> 4kb) ids.
         if ('_search/scroll' == $path) {
-            $data = $params[self::OPTION_SCROLL_ID];
+            $data = [ self::OPTION_SCROLL_ID => $params[self::OPTION_SCROLL_ID] ];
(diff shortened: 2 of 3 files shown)


bgaleotti:
https://github.com/elastic/elasticsearch-php/commit/59b3c08 looks the same

ruflin:
@nomoa LGTM. Could you add a changelog entry?

nomoa:
@ruflin done

ruflin:
@nomoa Merged. Thanks.

etudor:
Had the same problem with: ``` ruflin/elastica 5.2.1 ``` and ``` elasticsearch 5.4.0 ```
```
{
  "error": {
    "root_cause": [
      {
        "type": "illegal_argument_exception",
        "reason": "Failed to parse request body"
      }
    ],
    "type": "illegal_argument_exception",
    "reason": "Failed to parse request body",
    "caused_by": {
      "type": "json_parse_exception",
      "reason": "Unrecognized token 'DXF1ZXJ5QW5kRmV0Y2gBAAAAAAAHGKoWY3h2eGxBcHZUUGlINHRDa21USy0xUQ': was expecting ('true', 'false' or 'null')\n at [Source: org.elasticsearch.transport.netty4.ByteBufStreamInput@6d20e23; line: 1, column: 64]"
    }
  },
  "status": 400
}
``` 
Updating ``` ruflin/elastica ``` to ```dev-master``` solved the problem.

ruflin:
The 5.3 release went out yesterday which should include this bugfix.

PR #1333, opened 2017-07-07
update Travis to new Ubuntu Trusty images 

p365labs:
Hi All, after making a review on the problem we had yesterday on Travis I found that the solution look easier than what I thought :) so @ruflin this way we will start implementing new docker Ubuntu Images (trusty 14.04). I fix a minor error on php71 with the xdebug extension which looks updated on the base image.


ruflin:
This is great. Thanks a lot for taking care of this.

PR #1339, opened 2017-07-26
update Docker images to support ES6.0 alpha, removed groovy scripting

p365labs:
these amends will allow to start ES 6.0 properly (mapper-attachments plugin, scripts.inline, groovy scripts has been removed)

### DIFF
--- a/env/elasticsearch/Dockerfile
+++ b/env/elasticsearch/Dockerfile
@@ -1,7 +1,7 @@
-FROM docker.elastic.co/elasticsearch/elasticsearch:5.4.1
+FROM docker.elastic.co/elasticsearch/elasticsearch:6.0.0-alpha2
 MAINTAINER Nicolas Ruflin <spam@ruflin.com>
 
-RUN /usr/share/elasticsearch/bin/elasticsearch-plugin install mapper-attachments
+RUN /usr/share/elasticsearch/bin/elasticsearch-plugin install ingest-attachment
 
 # Copy config files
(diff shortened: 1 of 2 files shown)


p365labs:
@ruflin with these small amends on the Dockerfile and elasticsearch.yml we will be able to execute tests ... as expected the main effort should be pointed on the Types removal

env/elasticsearch/Dockerfile
```diff
@@ -1,8 +1,6 @@
-FROM docker.elastic.co/elasticsearch/elasticsearch:5.4.1
+FROM docker.elastic.co/elasticsearch/elasticsearch:6.0.0-alpha2
 MAINTAINER Nicolas Ruflin <spam@ruflin.com>
 
-RUN /usr/share/elasticsearch/bin/elasticsearch-plugin install mapper-attachments
```
ruflin: Should we replace this with `ingest-attachment`?

ruflin:
Nice, see my comment.

Unfortuantely travis did not build it as I told travis to only build master as it would otherwise build my branches always twice.

p365labs:
@ruflin I've seen ur comment, sonds good. What if I try to merge it on master? as we can start share errors on TYPES.. and take a decision on how to implement a strategy for the new ES6.0 Indices ?

ruflin:
@p365labs Open a PR directly against master so we see the errors.

p365labs:
ok!

PR #1340, opened 2017-07-27
update Docker images to support ES6.0 alpha, removed groovy scripting

p365labs:
This commit will allow to start ES 6.0 properly (mapper-attachments plugin, scripts.inline, groovy scripts has been removed, install ingest-mapper)


ruflin:
Seems like most breaking tests are type related as expected :-)

Makefile
```diff
@@ -60,7 +60,9 @@ tests:
 	make elastica-image
 	make start
 	mkdir -p build
-	docker run --network=elastica_esnet elastica_elastica  make phpunit
+	#docker run --name=nginx -e "ES_HOST=elasticsearch" --network=elastica_esnet elastica_nginx nginx
```
ruflin: Where is this coming from?

docker-compose.yml
```diff
@@ -34,19 +35,19 @@ services:
     volumes:
       - esdata1:/usr/share/elasticsearch/data
     ports:
-      - 9200:9200
+      - 19200:19200
```
ruflin: Anything wrong with using 9200?

ruflin:
Going to merge this as the two comments above can be still discussed / fixed in a follow up PR.

ruflin:
@p365labs Obviously thinking of this 10s after I merged, but we should update the CHANGELOG for this.

docker-compose.yml
p365labs: no my fault :) something that should be removed

PR #1347, opened 2017-08-11
ES6 use strict type on boolean  (true|false)

p365labs:
Updated failing tests failing on boolean type or on legacy values

- Removed the use of values different from TRUE or FALSE as [strict boolean type](https://github.com/elastic/elasticsearch/pull/22200) implemented in ES6.

- Removed 'analyzed' and not analyzed as index mapping now accepts only TRUE|FALSE

- Updated the use of store mapping with BOOLEAN

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,14 @@
 All notable changes to this project will be documented in this file based on the [Keep a Changelog](http://keepachangelog.com/) Standard. This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/5.3.0...master)
-
+ 
 ### Backward Compatibility Breaks
-- Numeric to and from parameters in [date_range aggregation](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/breaking_60_aggregations_changes.html#_numeric_literal_to_literal_and_literal_from_literal_parameters_in_literal_date_range_literal_aggregation_are_interpreted_according_to_literal_format_literal_now) are interpreted according to format of the target field 
+
+- Numeric to and from parameters in [date_range aggregation](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/breaking_60_aggregations_changes.html#_numeric_literal_to_literal_and_literal_from_literal_parameters_in_literal_date_range_literal_aggregation_are_interpreted_according_to_literal_format_literal_now) are interpreted according to format of the target field
+- In ES6 only [strict type boolean](https://github.com/elastic/elasticsearch/pull/22200) are accepted. [On ES6 docs](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/boolean.html)
+- removed analyzed/not_analyzed on [indices mapping](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/mapping-index.html)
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -2,9 +2,14 @@
 All notable changes to this project will be documented in this file based on the [Keep a Changelog](http://keepachangelog.com/) Standard. This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/5.3.0...master)
-
+ 
 ### Backward Compatibility Breaks
-- Numeric to and from parameters in [date_range aggregation](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/breaking_60_aggregations_changes.html#_numeric_literal_to_literal_and_literal_from_literal_parameters_in_literal_date_range_literal_aggregation_are_interpreted_according_to_literal_format_literal_now) are interpreted according to format of the target field 
+
+- Numeric to and from parameters in [date_range aggregation](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/breaking_60_aggregations_changes.html#_numeric_literal_to_literal_and_literal_from_literal_parameters_in_literal_date_range_literal_aggregation_are_interpreted_according_to_literal_format_literal_now) are interpreted according to format of the target field
```
ruflin: We should also link to the PR number. I'm aware this is normally tricky before opening the PR because it is not known yet :-)

test/Elastica/Query/BoostingTest.php
```diff
@@ -16,15 +16,15 @@ class BoostingTest extends BaseTest
         ['name' => 'Vital Match', 'price' => 2.1],
         ['name' => 'Mercury Vital', 'price' => 7.5],
         ['name' => 'Fist Mercury', 'price' => 3.8],
-        ['name' => 'Lama Vital 2nd', 'price' => 3.2],
+        ['name' => 'Lama Vital 2nd', 'price' => 1.2],
```
ruflin: Why did this change?

test/Elastica/Query/MoreLikeThisTest.php
```diff
@@ -28,8 +26,8 @@ public function testSearch()
 
         $type = new Type($index, 'helloworldmlt');
         $mapping = new Mapping($type, [
-            'email' => ['store' => 'true', 'type' => 'text', 'index' => 'analyzed'],
-            'content' => ['store' => 'true', 'type' => 'text',  'index' => 'analyzed'],
+            'email' => ['store' => 'true', 'type' => 'text', 'index' => 'true'],
```
ruflin: We could probably now use `true` instead of `'true'` here?

ruflin:
Decided to already merge this as all of the comments we can fix in follow up comments.

CHANGELOG.md
p365labs: ok I'll update it in the next PR

test/Elastica/Query/BoostingTest.php
p365labs: I was testing also the BoostingTest.php, and I forget to remove it during the creation of the PR.

test/Elastica/Query/MoreLikeThisTest.php
p365labs: dunno why :) revert in next PR, sorry.

PR #1348, opened 2017-08-15
Delete indices only by concrete name

p365labs:
It has been introduced a BC into ES 6 which force to delete an index *only* with [**concrete Index name**](https://github.com/elastic/elasticsearch/pull/25268). I've update a test, and created a new one to test the exception in case of index deletion by alias.


p365labs:
@ruflin done!

PR #1352, opened 2017-08-17
In QueryString cannot use [fields] parameter in conjunction with [default_field]

p365labs:
In QueryString cannot use anymore [fields] parameter in conjunction with [default_field]. There was a test failing because of an exception : **QueryValidationException**

After having a look a the docs on Es6 Breaking changes I did not find anything reporting this BC, the only way to check it is having a look at Elasticsearch 6.0+ source code.

Inside [QueryStringQueryBuilderTests.java](https://github.com/elastic/elasticsearch/blob/6.0/core/src/test/java/org/elasticsearch/index/query/QueryStringQueryBuilderTests.java) there a new test [testDefaultFieldsWithFields](https://github.com/elastic/elasticsearch/blob/6.0/core/src/test/java/org/elasticsearch/index/query/QueryStringQueryBuilderTests.java#L917) which checks exactly if both parameters are set and in case raise a QueryValidationException. 

I also implement a test to check for that Exception.


p365labs:
Hi @ruflin I changed the title in [WIP] as I have to inspect why some tests fails on travis. Wait on merging this one. TNX

p365labs:
The problem seems that the Exception return **query_validation_exception** not in the same position of the array. Instead of using assertEquals and validate the value of a specific position *$error['failed_shards'][0]['reason']['caused_by']['type']* I tried assertContains; like that seem to work

ruflin:
@p365labs CI is green, but the usual changelog conflicts.

PR #1354, opened 2017-08-21
Responses created|found on Index,Delete and Bulk are removed

p365labs:
The [created and found](https://github.com/elastic/elasticsearch/pull/25516) fields in index and delete responses became obsolete after the introduction of the result field in index, update and delete responses [#1354](https://github.com/ruflin/Elastica/pull/1354)


p365labs:
ok done, build on travis running :)

PR #1355, opened 2017-08-21
HTTP Transport already append / between path and host+port

p365labs:
There are many failing tests in Connection\Strategy classes that fails because the *_aliases* enpoint
are called with a slash that is already appened in HTTP Transport. So the final **GET** will be

*curl localhost:9200//_aliases*

This endpoint didn't reaised exception in ES 5.x, now it returns this error : 

```Incorrect HTTP method for uri [//_aliases] and method [GET], allowed: [PUT]","status":405```

With guzzle transport this doesn't happened 'cause guzzle solve this problem. I thought initially to add a test inside HttpTest.php, but I think that this is a responsibility of the developer.

@ruflin  What do you think ?

### DIFF
--- a/test/Elastica/Connection/Strategy/CallbackStrategyTest.php
+++ b/test/Elastica/Connection/Strategy/CallbackStrategyTest.php
@@ -75,8 +75,6 @@ class CallbackStrategyTest extends Base
      */
     public function testConnection()
     {
-        $this->markTestSkipped('ES6 update: Incorrect HTTP method for uri [//_aliases] and method [GET], allowed: [PUT]');
-
         $count = 0;
 
         $config = ['connectionStrategy' => function ($connections) use (&$count) {
@@ -86,7 +84,7 @@ class CallbackStrategyTest extends Base
        }];
 
         $client = $this->_getClient($config);
-        $response = $client->request('/_aliases');
+        $response = $client->request('_aliases');
 
         $this->assertEquals(1, $count);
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: CHANGELOG.md [hunk 1]
```diff
@@ -15,6 +15,7 @@ All notable changes to this project will be documented in this file based on the
 - In QueryString is not allowed to use fields parameters in conjunction with default_field parameter. This is not well documented, it's possibile to understand from [Elasticsearch tests :  QueryStringQueryBuilderTests.java](https://github.com/elastic/elasticsearch/blob/6.0/core/src/test/java/org/elasticsearch/index/query/QueryStringQueryBuilderTests.java#L917) [#1352](https://github.com/ruflin/Elastica/pull/1352)   
 - Index mapping field of type [*'string'*](https://www.elastic.co/guide/en/elasticsearch/reference/5.5/string.html) has been removed from Elasticsearch 6.0 codebase [#1353](https://github.com/ruflin/Elastica/pull/1353)
 - The [created and found](https://github.com/elastic/elasticsearch/pull/25516) fields in index and delete responses became obsolete after the introduction of the result field in index, update and delete responses [#1354](https://github.com/ruflin/Elastica/pull/1354) 
+- In HTTP Transport layer a '/' is already appended between host+port and path, since ES 6.0 if you make a request to _aliases with '/' it raises a 405 error
```
ruflin: I would not add this to the change log as we didn't change anything on our side. We only modified our tests to be "correct".

p365labs: @ruflin Another way to do it would be to add a sanitizer inside HTTP transport, and the make some tests against it. I prefer this solution, I mean leaving the responsibility on the shoulder of the developer. But we can discuss it if u prefer.

I'll remove the comment

ruflin:
I like it the way we have it and not fixing things for the dev. The more magic we do, the more likely it is to break :-)

PR #1356, opened 2017-08-22
Drop php5.x support, and remove old support classes

p365labs:
- Dropped support for **php5**
- update .travis.yml to execute tests only on 7.x php platforms
- update docker elastica configuration : removed Docker56 file
- remove Bool & Null classes (they worked only on 5.x php versions)
- set **dev-master** for elasticsearch-php until they released a stable version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,6 @@ branches:
     - master
 
 env:
-  - TARGET="56"
   - TARGET="70"
   - TARGET="71"
 
--- a/composer.json
+++ b/composer.json
@@ -12,9 +12,9 @@
         }
     ],
     "require": {
-        "php": ">=5.6.0",
+        "php": "^7.0",
         "psr/log": "~1.0",
-        "elasticsearch/elasticsearch": "5.3.0"
+        "elasticsearch/elasticsearch": "dev-master"
     },
     "require-dev": {
         "aws/aws-sdk-php": "~3.0",
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     "require": {
         "php": "^7.0",
         "psr/log": "~1.0",
-        "elasticsearch/elasticsearch": "dev-master"
+        "elasticsearch/elasticsearch": "6.0.0-beta1"
     },
     "require-dev": {
         "aws/aws-sdk-php": "~3.0",


.travis.yml
```diff
@@ -12,7 +12,6 @@ branches:
     - master
 
 env:
-  - TARGET="56"
```
ruflin: I assume this has to be dropped because tests were not passing with elasticsearch-php for 5.6?

The reason I'm asking is because I would like to keep the tests for 5.6 around as long as possible, even if we don't official support it. But if it becomes a hassle, lets remove it. I probably have to check if es-php is really not compatible with 5.6 or if it just removed it from the composer file.

p365labs: Ok got it, I tried executing tests with new deps and php5.6

first error : 
```
Problem 1
    - This package requires php ^7.0 but your PHP version (5.6.31) does not satisfy that requirement.
  Problem 2
    - Installation request for elasticsearch/elasticsearch dev-master -> satisfiable by elasticsearch/elasticsearch[dev-master].
    - elasticsearch/elasticsearch dev-master requires ext-json >=1.3.7 -> the requested PHP extension json has the wrong version (1.2.1) installed.
```

due to composer.json reqs in es-php, so I try using **--ignore-platform-reqs** in order to skip php platform requirements.

Then we have another error :

> Parse error: syntax error, unexpected ':', expecting ';' or '{' in /root/composer/vendor/doctrine/instantiator/src/Doctrine/Instantiator/Instantiator.php on line 95
> Call Stack:
>     0.0002     234056   1. {main}() /root/composer/vendor/phpunit/phpunit/phpunit:0
>     0.0184    1065728   2. PHPUnit_TextUI_Command::main() /root/composer/vendor/phpunit/phpunit/phpunit:52
>     0.0185    1066480   3. PHPUnit_TextUI_Command->run() /root/composer/vendor/phpunit/phpunit/src/TextUI/Command.php:116
>     0.9737   32252296   4. PHPUnit_TextUI_TestRunner->doRun() /root/composer/vendor/phpunit/phpunit/src/TextUI/Command.php:186
>     1.0118   32801632   5. PHPUnit_Framework_TestSuite->run() /root/composer/vendor/phpunit/phpunit/src/TextUI/TestRunner.php:517
>     2.2593   40272824   6. PHPUnit_Framework_TestSuite->run() /root/composer/vendor/phpunit/phpunit/src/Framework/TestSuite.php:722
>     2.2648   40276728   7. PHPUnit_Framework_TestSuite->run() /root/composer/vendor/phpunit/phpunit/src/Framework/TestSuite.php:722
>     2.2649   40281032   8. PHPUnit_Framework_TestCase->run() /root/composer/vendor/phpunit/phpunit/src/Framework/TestSuite.php:722
>     2.2649   40281656   9. PHPUnit_Framework_TestResult->run() /root/composer/vendor/phpunit/phpunit/src/Framework/TestCase.php:860
>     2.2663   40284560  10. PHPUnit_Framework_TestCase->runBare() /root/composer/vendor/phpunit/phpunit/src/Framework/TestResult.php:686
>     2.2680   40302592  11. PHPUnit_Framework_TestCase->runTest() /root/composer/vendor/phpunit/phpunit/src/Framework/TestCase.php:905
>     2.2690   40303744  12. ReflectionMethod->invokeArgs() /root/composer/vendor/phpunit/phpunit/src/Framework/TestCase.php:1054
>     2.2690   40303848  13. Elastica\Test\Bulk\ResponseSetTest->testIsOk() /root/composer/vendor/phpunit/phpunit/src/Framework/TestCase.php:1054
>     2.2690   40304024  14. Elastica\Test\Bulk\ResponseSetTest->_createResponseSet() /elastica/test/Elastica/Bulk/ResponseSetTest.php:21
>     2.2691   40304184  15. PHPUnit_Framework_TestCase->createMock() /elastica/test/Elastica/Bulk/ResponseSetTest.php:145
>     2.2725   40587416  16. PHPUnit_Framework_MockObject_MockBuilder->getMock() /root/composer/vendor/phpunit/phpunit/src/Framework/TestCase.php:1460
>     2.2726   40587792  17. PHPUnit_Framework_MockObject_Generator->getMock() /root/composer/vendor/phpunit/phpunit-mock-objects/src/Framework/MockObject/MockBuilder.php:118
>     2.3138   40756808  18. PHPUnit_Framework_MockObject_Generator->getObject() /root/composer/vendor/phpunit/phpunit-mock-objects/src/Framework/MockObject/Generator.php:275
>     2.3158   40954640  19. spl_autoload_call() /root/composer/vendor/phpunit/phpunit-mock-objects/src/Framework/MockObject/Generator.php:305
>     2.3160   40954880  20. Composer\Autoload\ClassLoader->loadClass() /root/composer/vendor/phpunit/phpunit-mock-objects/src/Framework/MockObject/Generator.php:305
>     2.3163   40955064  21. Composer\Autoload\includeFile() /root/composer/vendor/composer/ClassLoader.php:322



any idea anyone ? maybe it's an easier task not but I'm nop sure about it.
making a quick search it depends on [this issue](https://github.com/doctrine/instantiator/issues/30) 
@ruflin would you like from me to get a little more deeper ?

p365labs:
@ruflin I would like to understand why you're saying 
```
I would like to keep the tests for 5.6 around as long as possible, even if we don't official support it.
```

If we decide to follow the official client, they have already dropped support for PHP 5.x, it sounds a bit strange to me to maintain code to test it against 5.x when the client doesn't support it, even if the tests don't fail.
Can we talk about this ?
Maybe I can split the PR in order to move on to implement es-php-6.0-beta1 and the we can continue discussing on the php matter, what do u guess?

ruflin:
For the PHP 5.6 issue. I agree that it is kind of incosistent, but I already see some push backs happening. In general my approach is to remove compatibility if a code changes require it. If there are things in elasticsearch-php that do not run under 5.6, I'm ok with dropping it. But if it's just the composer file that says it's not working, I would keep our tests around. Officially I would still say we only support 7.0 but we still test with 5.6 because we get it kind of for free. Like this we also immidiately know when we break it can can leave an additional note in the Changelog.

There is also a difference between the official client which is commercially supported, so stating there it works with 5.6 would mean it must be commercially supported. That is one of the nice things of Elastica as it is kind of a grey area at the moment and we can do these tricks as there is no commercial support.

To move forward I'm also ok to drop travis build for now for 5.6 and do a follow up PR to figure out what goes wrong, in this PR I would only not remove the Dockerfile needed for 5.6

p365labs:
Ok It sounds ok to me, and I can agree and uderstand the approach. 
I will fix the PR and maybe create a new one trying to make it work on 56.

Thanks again

p365labs:
I've left the dockerfile for php56, and the Bool and Null classes (which are there for BC), removed the amends to test/Elastica/SnapshotTest.php which was something I forgot and commit to this branch. Now we removed php56 from travis and updated composer.json

PR #1360, opened 2017-08-23
update testNegativeBoosting

p365labs:
I spent a lot of time on this as *testNegativeBoosting* was failing due to a different order in sorting results and scoring. I did not found documentation but some help from @polyfractal and after a bit of discussion on ES forums and **many many** tests on different ES version I reach the fact that that test was wrong. In that test we were testing a specific result due to negative_boosting and due to scoring and sorting of the query. The problem here is that we are testing the implementation of scoring and sorting of Lucene. in fact [Elasticsearch tests on BoostingQuery](https://github.com/elastic/elasticsearch/blob/6.0/core/src/test/java/org/elasticsearch/index/query/BoostingQueryBuilderTests.java) did not test any result they only test if the json and the query is well formed.

I've found that in [Lucene 7.0  some feature of TF/IDF](https://issues.apache.org/jira/browse/LUCENE-7347) has been removed and this could lead to different scoring.
Talking with @polyfractal [here](https://discuss.elastic.co/t/elasticsearch-6-0-beta1-boosting/97049/13) we agree at some point to not test on result of BoostingQuery but as Elasticsearch do only testing the well formed Boosting query.

> In general we rely on Lucene’s test framework to work out issues that directly relate to lucene queries (e.g. we pass the boosting query directly to Lucene without any modification, so we rely on their tests).
> 
> Lucene occasionally adds things that change scoring, like 7347 or the earlier changes to TF/IDF vs BM25. As a downstream user of Lucene we just accept these changes and don’t try to keep scoring backwards compatible, since that would be a losing proposition.

and again

> ++ to not checking default scoring order. There are places in ES (and lang clients) where we validate the response to various degrees, but it usually involves something deterministic to make it work. E.g. you can check user-defined sorting on fields, or the values of aggregation results, etc. But things like implicit scoring can be tricky indeed, since it’s liable to change slightly.

So my suggestion as implemented in this PR is to completely remove that functional test, and I crated a new one which exactly copy the one from Elasticsearch Tests

### DIFF
--- a/test/Elastica/Query/BoostingTest.php
+++ b/test/Elastica/Query/BoostingTest.php
@@ -9,36 +9,6 @@ use Elastica\Test\Base as BaseTest;
 class BoostingTest extends BaseTest
 {
     /**
-     * @var array
-     */
-    protected $sampleData = [
-        ['name' => 'Vital Lama', 'price' => 5.2],
-        ['name' => 'Vital Match', 'price' => 2.1],
-        ['name' => 'Mercury Vital', 'price' => 7.5],
-        ['name' => 'Fist Mercury', 'price' => 3.8],
-        ['name' => 'Lama Vital 2nd', 'price' => 3.2],
-    ];
-
-    protected function _getTestIndex()
-    {
-        $index = $this->_createIndex();
-        $type = $index->getType('test');
-        $type->setMapping([
-            'name' => ['type' => 'text', 'index' => 'true'],
-            'price' => ['type' => 'float'],
-        ]);
-        $docs = [];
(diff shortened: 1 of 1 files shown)


ruflin:
Fully agree on the conclusion of this adventure that we should remove the detailed tests and let it to Lucene to make sure it works as expected.

It think it is a general problem we sometimes have that our tests are too detailed. On the one hand that is a very positive thing as it make sure things work exactly as they should and I often use for developing new features. But then if things change we need to update all the tests. Challenge is to find the right balance.

PR #1361, opened 2017-08-25
Querystring bool values should be a string

p365labs:
Due to ES 6.0 bool strict type, boolean values in as url parameters (querystring) should be passed as string (true|false). Http & Guzzle Transport use to pass bool as php do so **true** was converted to **1**. 

Due to strict type hint only "true" and "false" are allowed I updated AbtractTransport and implemented a simple sanityzer that convert boolean values to String, and implemented a functional test to check it.

### DIFF
--- a/lib/Elastica/Transport/AbstractTransport.php
+++ b/lib/Elastica/Transport/AbstractTransport.php
@@ -61,6 +61,23 @@ abstract class AbstractTransport extends Param
     abstract public function exec(Request $request, array $params);
 
     /**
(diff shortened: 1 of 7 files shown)


ruflin:
Nice fix. I'm kind of surprised that true and false are converted to just 1 and 0 by default.

p365labs:
it's PHP :) **http_build_query** this function will convert true and false in 1 and 0

ruflin:
:-(

PR #1362, opened 2017-08-28
Update tests with two mapping type and some other

p365labs:
- Updated lot of tests that used two mapping type 
- remove reference to `include_in_all ` in mapping as [removed](https://github.com/elastic/elasticsearch/pull/22970) in 6.0 
- update some tests wich created non-unique index name which created some troubles

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ All notable changes to this project will be documented in this file based on the
 - In QueryString is not allowed to use fields parameters in conjunction with default_field parameter. This is not well documented, it's possibile to understand from [Elasticsearch tests :  QueryStringQueryBuilderTests.java](https://github.com/elastic/elasticsearch/blob/6.0/core/src/test/java/org/elasticsearch/index/query/QueryStringQueryBuilderTests.java#L917) [#1352](https://github.com/ruflin/Elastica/pull/1352)   
 - Index mapping field of type [*'string'*](https://www.elastic.co/guide/en/elasticsearch/reference/5.5/string.html) has been removed from Elasticsearch 6.0 codebase [#1353](https://github.com/ruflin/Elastica/pull/1353)
 - The [created and found](https://github.com/elastic/elasticsearch/pull/25516) fields in index and delete responses became obsolete after the introduction of the result field in index, update and delete responses [#1354](https://github.com/ruflin/Elastica/pull/1354) 
- 
+
 ### Bugfixes
 - Enforce [Content-Type requirement on the layer Rest](https://github.com/elastic/elasticsearch/pull/23146), a [PR on Elastica #1301](https://github.com/ruflin/Elastica/issues/1301) solved it (it has been implemented only in the HTTP Transport), but it was not implemented in the Guzzle Transport. [#1349](https://github.com/ruflin/Elastica/pull/1349)
 - Scroll no longer does an extra iteration both on an empty result and on searches where the last page has a significantly smaller number of results than the pages before it.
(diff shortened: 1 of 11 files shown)


test/Elastica/IndexTest.php
```diff
@@ -356,167 +356,238 @@ public function testCount()
      */
     public function testDeleteByQueryWithQueryString()
     {
-        $this->markTestSkipped('ES6 update: the final mapping would have more than 1 type');
+//        $this->markTestSkipped('ES6 update: the final mapping would have more than 1 type');
```
ruflin: Leftover?

test/Elastica/Base.php
```diff
@@ -135,7 +135,7 @@ protected function _checkScriptInlineSetting()
         $nodes = $this->_getClient()->getCluster()->getNodes();
         $scriptInline = $nodes[0]->getInfo()->get('settings', 'script', 'inline');
 
-        if ($scriptInline != 'true' && $this->_getVersion() < 6) {
+        if ($this->_getVersion() < 6 && $scriptInline != 'true') {
```
ruflin: There is no logical change here, right?

test/Elastica/Query/IdsTest.php
```diff
@@ -176,27 +156,11 @@ public function testSetTypeSingleSearchArrayDocInOtherType()
...
-        $this->assertEquals(2, $resultSet->count());
-    }
-
-    /**
-     * @group functional
-     */
     public function testSetTypeArraySearchArray()
     {
-        $this->markTestSkipped('ES6 update: the final mapping would have more than 1 type');
+        if (!$this->_getVersion() < 6) {
```
ruflin: We should leave the code in, but should we leave the tests in?

CHANGELOG.md
```diff
@@ -15,7 +15,8 @@ All notable changes to this project will be documented in this file based on the
 - In QueryString is not allowed to use fields parameters in conjunction with default_field parameter. This is not well documented, it's possibile to understand from [Elasticsearch tests :  QueryStringQueryBuilderTests.java](https://github.com/elastic/elasticsearch/blob/6.0/core/src/test/java/org/elasticsearch/index/query/QueryStringQueryBuilderTests.java#L917) [#1352](https://github.com/ruflin/Elastica/pull/1352)   
 - Index mapping field of type [*'string'*](https://www.elastic.co/guide/en/elasticsearch/reference/5.5/string.html) has been removed from Elasticsearch 6.0 codebase [#1353](https://github.com/ruflin/Elastica/pull/1353)
 - The [created and found](https://github.com/elastic/elasticsearch/pull/25516) fields in index and delete responses became obsolete after the introduction of the result field in index, update and delete responses [#1354](https://github.com/ruflin/Elastica/pull/1354) 
- 
+- Disallow [include_in_all](https://github.com/elastic/elasticsearch/pull/22970) for 6.0+ indices [#1362](https://github.com/ruflin/Elastica/pull/1362)
```
ruflin: There is no change for this in Elastica as far as I can see.

test/Elastica/Base.php
p365labs: no to be honest I'll be evaluate in a next PR if we still need _checkScriptInlineSetting()
as ES6 has script inline and stored allowed by default. So no need to enable them by elasticsearch.yml

test/Elastica/Query/IdsTest.php
p365labs: I agree... better to leave the code but remove the tests

CHANGELOG.md
p365labs: right. I removed the entry in the changelog.

PR #1363, opened 2017-08-28
Fix extra iteration on Scroll

dmvdbrugge:
`Scroll` did an extra iteration both on an empty result and on searches where the last page has a significantly smaller number of results than the pages before it.

You can see in the deleted part of the test that it actually needed one foreach too many, now it doesn't anymore.


CHANGELOG.md
```diff
@@ -18,7 +18,8 @@ All notable changes to this project will be documented in this file based on the
  
 ### Bugfixes
 - Enforce [Content-Type requirement on the layer Rest](https://github.com/elastic/elasticsearch/pull/23146), a [PR on Elastica #1301](https://github.com/ruflin/Elastica/issues/1301) solved it (it has been implemented only in the HTTP Transport), but it was not implemented in the Guzzle Transport. [#1349](https://github.com/ruflin/Elastica/pull/1349)
-  
+- Scroll no longer does an extra iteration both on an empty result and on searches where the last page has a significantly smaller number of results than the pages before it.
```
ruflin: I would more see this as an enhancement then a bug as the code was executing as expected, just not very performant. In can still change that later.

dmvdbrugge: I disagree, the docs on `valid()` (seem to) say the empty result shouldn't be a valid one, and thus the last iteration that is now prevented was not only "performance fix" but also unexpected behaviour.

In fact that is why I found the "bug", I assumed (because of `valid()`) the resultset couldn't be empty.

ruflin: Got it. Thanks for the clarification. Was not fully clear for me from the changelog / PR description.

PR #1368, opened 2017-09-01
script types are all allowed by default

p365labs:
- By default all script types are allowed to be executed.
- By default all script contexts are allowed to be executed. 

I updated Base.php as now you by config you can only disable them


ruflin:
Seems like one test contains more then one type:

```
1) Elastica\Test\ClientTest::testEndpointParamsRequest
Elastica\Exception\ResponseException: Rejecting mapping update to [elastica_elasticatestclienttestabfd] as the final mapping would have more than 1 type: [foobar, test]
```

p365labs:
@ruflin done

PR #1369, opened 2017-09-02
Remove Search and DSL changes

p365labs:
- The deprecated minimum_number_should_match parameter in the bool query has been removed, use minimum_should_match instead.
- For geo_distance queries, sorting, and aggregations the sloppy_arc option has been removed from the distance_type parameter. 
- The geo_distance_range query, which was deprecated in 5.0, has been removed. 
- The optimize_bbox parameter has been removed from geo_distance queries. 
- The disable_coord parameter of the bool and common_terms queries has been removed. If provided, it will be ignored and issue a deprecation warning. 
- The percolate query’s document_type has been deprecated. From 6.0 and later it is no longer required to specify the document_type parameter.


ruflin:
Yay, more code removed :-)

PR #1370, opened 2017-09-02
Analyze removed support for request parameters

p365labs:
Analyze endpoint removed support for request parameters, we should use request body instead. I updated the Analyze endpoint in order to accept not only text, but an array of values, so even the explain functionality will work.

### DIFF
--- a/lib/Elastica/Index.php
+++ b/lib/Elastica/Index.php
@@ -548,24 +548,24 @@ class Index implements SearchableInterface
      *
      * Detailed arguments can be found here in the link
      *
-     * @param string $text String to be analyzed
+     * @param array $body String to be analyzed
      * @param array  $args OPTIONAL Additional arguments
      *
      * @return array Server response
      *
      * @link https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-analyze.html
      */
-    public function analyze($text, $args = [])
+    public function analyze(array $body, $args = [])
     {
         $endpoint = new Analyze();
-        $endpoint->setBody(['text' => $text]);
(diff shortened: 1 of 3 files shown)


ruflin:
Guest what ... ^

lib/Elastica/Index.php
```diff
@@ -548,24 +548,24 @@ public function requestEndpoint(AbstractEndpoint $endpoint)
...
      * @param array  $args OPTIONAL Additional arguments
      *
      * @return array Server response
      *
      * @link https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-analyze.html
      */
-    public function analyze($text, $args = [])
+    public function analyze($body, $args = [])
     {
         $endpoint = new Analyze();
```
ruflin: I wonder if we should return an error in case body is not a map to make it easier for app devs to detect the problem early.

p365labs:
sorry I make some mistakes between comments and review ... ;)

- I updated the docblock as $body should be and array
- I declared $body as array forcing the strict type
what do u think ? should we really raise an error in case of $body not a map ?

ruflin:
I like the strict typing of array. This will also show developers and error very quickly which hopefully pushes them to read the changelog :-)

p365labs:
I like strict type hinting too... maybe when this major upgrade has ended we could plan a major refactor in order to make Elastica more respectful of it, considering we are now upgraded to php7.
Just an idea :)

ruflin:
I like the idea as it will make devs detect problems much earlier. But it will definitively break all older versions (which is probably fine looking at http://php.net/supported-versions.php).

PR #1372, opened 2017-09-08
Implement Pipeline and Processors

p365labs:
This has been a long implementation, Pipelines need to be implemented and even processor if we want to use field attachments.

- Implementation of Pipeline
- Implementation of Processors (only 4 processors are missing : **Foreach, Grok, Gsub, Script**)

This is just a first implementation : I have some doubt on extending **Param** for Processors and for Pipeline Object. It's very useful, 'cause extending ArrayableInterface Object are quickly converted to Arrays.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ All notable changes to this project will be documented in this file based on the
 ### Added
 
 - Added `Query\SpanContaining`, `Query\SpanWithin` and `Query\SpanNot` [#1319](https://github.com/ruflin/Elastica/pull/1319)
+- Implemented [Pipeline](https://www.elastic.co/guide/en/elasticsearch/reference/current/pipeline.html) and [Processors](https://www.elastic.co/guide/en/elasticsearch/reference/current/ingest-processors.html). [#1373](https://github.com/ruflin/Elastica/pull/1373)
 
 ### Improvements
 
--- a/lib/Elastica/Client.php
+++ b/lib/Elastica/Client.php
@@ -306,12 +306,13 @@ class Client
      * @link https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html
      *
      * @param array|\Elastica\Document[] $docs Array of Elastica\Document
+     * @param array $requestParams
      *
      * @throws \Elastica\Exception\InvalidException If docs is empty
      *
      * @return \Elastica\Bulk\ResponseSet Response object
      */
-    public function updateDocuments(array $docs)
+    public function updateDocuments(array $docs, array $requestParams = [])
     {
         if (empty($docs)) {
             throw new InvalidException('Array has to consist of at least one element');
@@ -320,6 +321,9 @@ class Client
         $bulk = new Bulk($this);
 
         $bulk->addDocuments($docs, Action::OP_TYPE_UPDATE);
+        foreach ($requestParams as $key => $value) {
+            $bulk->setRequestParam($key, $value);
+        }
 
         return $bulk->send();
     }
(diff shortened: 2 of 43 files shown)


p365labs:
I've made some minors update and implemented some tests, in order to execute the pipeline tests against an Index.

lib/Elastica/Pipeline.php
```diff
@@ -0,0 +1,220 @@
...
+
+        return $this->requestEndpoint($endpoint);
+    }
+
+    /**
+     * @todo implement simulate API
+     */
+    public function simulate()
+    {
+
```
ruflin: Should we throw an exception for now?

lib/Elastica/Pipeline.php
```diff
@@ -0,0 +1,220 @@
...
+use Elastica\Exception\InvalidException;
+use Elastica\Processor\AbstractProcessor;
+use Elasticsearch\Endpoints\AbstractEndpoint;
+use Elasticsearch\Endpoints\Ingest\Pipeline\Delete;
+use Elasticsearch\Endpoints\Ingest\Pipeline\Get;
+use Elasticsearch\Endpoints\Ingest\Pipeline\Put;
+
+/**
+ * Elastica Processors object.
```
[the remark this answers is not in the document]
p365labs: my fault sorry.

lib/Elastica/Pipeline.php
p365labs: ok I raise a *NotImplementedException('simulate API on Pipeline not yet implemented.');*

p365labs:
As for now the community can create and update processor as they change in time, the only processor not yet implemented are **Foreach, Grok, Gsub, Script**.

Param is the right structure to use in order to have the body of the API call well formed, the only concern is that Param was clearly created for queries (at the top of the class --> * * This function can be used to handle params for queries, filter*), and maybe I could create a real similar class called Option or PipelineOption which do exactly the same thing as Param. Maybe is something we can think about in a future if something will change in Pipeline definition or implementation in ES. Creating now an *Option* class would mean only duplicate Param class. I preferred this approach, but I would like to discuss it if it sound ok for you and the community.

Processors does not have functional tests except KvTest.php, I will implement asap, I was waiting for a review first :)

ruflin:
Using Params as you did in this PR LGTM. I initially created it to reduce code duplication in Queries/Filters but I think we can apply it to almost any Class which needs to be transformed from PHP Object to json at one point.

Ping me when you think it's ready to be merged.

p365labs:
Hi @ruflin, before the functional tests I've made a small refactor on *tests*. I created a BasePipeline class (which extends Base), which create a pipeline and also implement a tearDown in order to delete all the pipelines created to make tests atomic. Every **Processor** should extends this class so it would be easy to create a Pipeline (and a functional test...) and also I have avoided to implement the same code to create a Pipeline in all the Processor tests.

p365labs:
Hi @Ruflin the PR for me is ready to be merged; let me know if u have more comments.

p365labs:
@ruflin any comment on the review :) ?

ruflin:
@p365labs Will have a look pretty soon. A bit under water these days ;-)

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ All notable changes to this project will be documented in this file based on the
 - Added `Query\SpanContaining`, `Query\SpanWithin` and `Query\SpanNot` [#1319](https://github.com/ruflin/Elastica/pull/1319)
 
 ### Improvements
+- Implemented [Pipeline](https://www.elastic.co/guide/en/elasticsearch/reference/current/pipeline.html) and [Processors](https://www.elastic.co/guide/en/elasticsearch/reference/current/ingest-processors.html). [#1373](https://github.com/ruflin/Elastica/pull/1373)
```
ruflin: This should probably go under `Added`

lib/Elastica/Client.php
```diff
@@ -349,6 +356,12 @@ public function addDocuments(array $docs)
 
         $bulk->addDocuments($docs);
 
+        if (!empty($requestParams)) {
```
ruflin: I wonder if the `!empty` is even needed as we already define it as array.

p365labs:
done!

PR #1375, opened 2017-09-14
Removed mapper-attachments plugin. Now use the ingest-attachment plugin

p365labs:
[Mapper Attachment plugin has been removed](https://github.com/elastic/elasticsearch/pull/20416) Use Ingest-attachment plugin and attachment processors with pipeline to ingest new documents.

the flow for ingesting a file into Elasticsearch changed a bit :

- u should create a Pipeline with an *Attachment Processor*
- add a file to the document
- ad the document to the index using a query string param of this format : **pipeline=name_of_the_pipeline**


ruflin:
Thanks for doing the switch. This was actually one of the very early lines of code in Elastica :-)

Issue #1377, opened 2017-09-18
Bug with PHP 7.2


kukulich:
`Warning: count(): Parameter must be an array or an object that implements Countable` in https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Query.php#L333

p365labs:
can you show us the code that generate this error ?

kukulich:
The bug is here: https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Query.php#L425

`Elastica\Query\AbstractQuery` doesn't implement `Countable`.

p365labs:
I confirm that @kukulich is right. 
PHP 7.2 change a bit [count() function](https://github.com/php/php-src/blob/php-7.2.0RC2/UPGRADING) (search for Countable).

```count() now raises a warning when an invalid parameter is passed. Only arrays and objects implementing the Countable interface should be passed.```

I think could it be useful having a PR ready to be merged when PHP 7.2 will be released, at the moment PHP 7.2 is a Release Candidate 2, so not ready for production use.

ruflin:
@kukulich Thanks for reporting. Always good to have people testing early with future versions.

I'm ok with getting a fix in before GA is out as I assume it will not be a breaking change?

@p365labs Thanks for starting the work on this.

p365labs:
hi all! I think we can close this one as this PR #1378 solves the problem :)

PR #1378, opened 2017-09-20
Implement Countable on Objects in PHP7.2

p365labs:
as reported #1377 in PHP 7.2 it's possibile to use the function count() only on Arrays and on \Countable objects.

This is a WIP PR, there are still things to do and some tests to check; more then this I will merge this PR when PHP 7.2 will be released (now it's a RC2).

### DIFF
--- a/lib/Elastica/Query.php
+++ b/lib/Elastica/Query.php
@@ -330,7 +330,7 @@ class Query extends Param
             $this->setQuery(new MatchAll());
         }
 
-        if (isset($this->_params['post_filter']) && 0 === count($this->_params['post_filter'])) {
+        if (isset($this->_params['post_filter']) && 0 === count(($this->_params['post_filter'])->toArray())) {
             unset($this->_params['post_filter']);
         }
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ branches:
 env:
   - TARGET="70"
   - TARGET="71"
+  - TARGET="72"
 
 before_install:
   # check running "docker engine" and "docker-compose" version on travis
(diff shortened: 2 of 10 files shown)


lib/Elastica/Response.php
```diff
@@ -11,7 +11,7 @@
  *
  * @author Nicolas Ruflin <spam@ruflin.com>
  */
-class Response
+class Response implements \Countable
```
ruflin: Is this change also needed?

p365labs:
@ruflin Now the PR# it's not in WIP anymore, tests now pass also with PHP 7.2-rc (which is now RC5). Today will be released php7.2-rc6 and for the end of the current month [PHP7.2 GA](https://wiki.php.net/todo/php72) will be released. Feel free to merge it or wait till PHP7.2 . GA.

lib/Elastica/Query.php
```diff
@@ -330,7 +330,8 @@ public function toArray()
             $this->setQuery(new MatchAll());
         }
 
-        if (isset($this->_params['post_filter']) && 0 === count($this->_params['post_filter'])) {
+        $a = $this->_params['post_filter'];
```
ruflin: Nit picking but could we name this variable differently, something more describtive? I'm curious what happens on the next line in case `toArray()` does not on the varialbe?

lib/Elastica/Response.php
```diff
@@ -361,4 +361,18 @@ public function getJsonBigintConversion()
     {
         return $this->_jsonBigintConversion;
     }
+
+//    /**
```
ruflin: We should remove this code if not needed anymore.

p365labs:
@ruflin the pr has been updated :) removed unused code and remove that weird var name :)

XWB:
Hi @ruflin 

Would it be possible to cherry-pick this commit to the 5.x branch?

ruflin:
@XWB Sound reasonable to me. Do you want to open a backport PR?

XWB:
@ruflin Done https://github.com/ruflin/Elastica/pull/1428

richardfullmer:
The change at https://github.com/ruflin/Elastica/pull/1378/files#diff-e421c6a76338876301c98a5a7e95ab87R333 broke my implementation today since `$this->_params['post_filter']` _was_ already an array.

Noticed exception 'Symfony\Component\Debug\Exception\FatalThrowableError' with message 'Call to a member function toArray() on array'

ruflin:
@richardfullmer Is this the same issue as in https://github.com/ruflin/Elastica/issues/1429 ?

richardfullmer:
Yes, looks the same to me.

PR #1380, opened 2017-09-22
prevent string submission to int value in min_doc_count. close #1379

### LINKED ISSUE #1379: Bug in min_doc_count
In [this lines of code](https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Aggregation/AbstractTermsAggregation.php#L16-L19) we can assign string value to `min_doc_count` and get elastic error. Should we convert value to int here?

osavchenko:
Description in #1379

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ All library issues should go to the [issue tracker from github](https://github.c
 
 Compatibility
 -------------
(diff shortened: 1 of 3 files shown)


giovannialbero1992:
@osavchenko now elastica is php7 compliant and has dropped support to php5. Maybe it can be useful start to use type hinting http://php.net/manual/en/migration70.new-features.php.

PR #1383, opened 2017-10-03
Implemented join datatype

p365labs:
Implemented [join datatype](https://www.elastic.co/guide/en/elasticsearch/reference/current/parent-join.html) and solved all failing tests : 

```The join datatype is a special field that creates parent/child relation within documents of the same index. The relations section defines a set of possible relations within the documents, each relation being a parent name and a child name. ```


ruflin:
Wow, that PR is quite an effort on cleaning / fixing the tests. Thanks a lot.

p365labs:
in fact all tests are fixed now ;) The only that are skipped are the ones that remains for php 5.x which are not working in php 7 (Null, Bool are reserved words in php 7).

PR #1384, opened 2017-10-04
Update wrong typing PHPDoc on the setRefreshInterval method

damienalexandre:
The `setRefreshInterval` method expect a string but the method documentation ask for integer.

Also the comment tells that this value is milliseconds but that's not the case anymore in Elasticsearch.

### DIFF
--- a/lib/Elastica/Index/Settings.php
+++ b/lib/Elastica/Index/Settings.php
@@ -263,7 +263,7 @@ class Settings
      * Sets the index refresh interval.
      *
      * Value can be for example 3s for 3 seconds or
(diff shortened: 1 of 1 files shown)


p365labs:
That seems good to me 👍  I will remove this part of the comment :

```-1 refreshing is disabled.```

which is not true anymore, have a look [here](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/indices-update-settings.html#bulk)

damienalexandre:
Yes that sentence was not clear, "-1" is to disable the refresh. PR updated, thanks!

ruflin:
@damienalexandre Thanks for the fix.

PR #1388, opened 2017-10-07
Fix some tests

p365labs:
It looks like when I updated some tests I left some code that should be remove. Removing type did not only mean use two indices in tests. Now it looks like tests are executing without flaky ones.

### DIFF
--- a/test/Elastica/Aggregation/ChildrenTest.php
+++ b/test/Elastica/Aggregation/ChildrenTest.php
@@ -13,8 +13,11 @@ class ChildrenTest extends BaseAggregationTest
     {
         $client = $this->_getClient();
         $index = $client->getIndex('testaggregationchildren');
-        $index->create([], true);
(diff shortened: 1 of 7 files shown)


test/Elastica/Aggregation/ChildrenTest.php
```diff
@@ -13,8 +13,13 @@ protected function _getIndexForTest()
     {
         $client = $this->_getClient();
         $index = $client->getIndex('testaggregationchildren');
-        $index->create([], true);
-        $type = $index->getType('test');
+        $index->create(['index' => ['number_of_shards' => 2, 'number_of_replicas' => 1]], true);
+
+//        $index = $this->_createIndex();
```
ruflin: Leftover?

p365labs: yes... sorry! removed now ;)

PR #1395, opened 2017-10-25
Clear search context after usage in scroll

mjanser:
Because open search contexts have a cost they should be explicitly
cleared as soon as the scroll is not being used anymore.

The method `clear()` is exposed in `Scroll` to also be able to clear the
search context manually from outside.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,11 @@ All notable changes to this project will be documented in this file based on the
 
 ### Added
 
+- Added clear() to `Scroll` for closing search context on ES manually
+
 ### Improvements
+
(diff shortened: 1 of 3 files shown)


mjanser:
I based this on 5.x because that's the currently stable version of ES and we'd like to use this change as soon as possible. Is that okay or should I open a pull request against master and you merge it to 5.x?

ruflin:
@mjanser Thanks for the PR. Any chance you could add a test for this?

I'm ok with having this first in 5.x but would really appreciate if you could also open a follow up PR with this for 6.0.

mjanser:
@ruflin I added a test and created a second pull request for master: #1399

PR #1398, opened 2017-11-02
update to es6-rc2 

p365labs:
This PR will 

- update the docker image to ES6 RC2
- removed xdebug 2.4 from the Dockerfile as it could lead to some errors on TRAVIS. Travis already has xdebug
- solves a BC break on innerHits Query

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file based on the
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/6.0.0-beta1...master)
 
 ### Backward Compatibility Breaks
+- Return the [_source of inner hit nested](https://github.com/elastic/elasticsearch/pull/26982) as is without wrapping it into its full path context [#1398](https://github.com/ruflin/Elastica/pull/1398)
 
 ### Bugfixes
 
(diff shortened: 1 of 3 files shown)


ruflin:
Thanks for the PR. Any chance you could move the innerHits change to an other PR?

For the xdebug part: How does it help that xdebug is already part of Travis as these tests run inside the docker containers?

p365labs:
The problem splitting the PR is that the BC has been introduced exactly by the RC2. So `testInnerHitsNested` will fail and so the build if I split the PR in two PR's. 

It's a bit weird they have introduced this BC in [ES-6-beta2](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/release-notes-6.0.0-beta2.html#breaking-6.0.0-beta2) and then introduced (in *RC2*) [the BC Break of a BC Break](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/release-notes-6.0.0-rc2.html)... 
It shows to Elastic is working hard in order to release the new ES6.ù

Xdebug, it's ok, it's tru what u said. It's that I noticed that sometimes tests fail due to this error :
```
Step 6/13 : RUN pecl install xdebug-2.4.0
 ---> Running in b33bc5554f34
Package "xdebug" Version "2.4.0" does not have REST xml available
install failed
```
and I still don't understand, I thought removing it will solve the problem. Maybe I need to inspect on that more ;)

ruflin:
PR merged. For the xdebug issue: As far as I remember xdebug is needed for the coverage report, so I think we still need it but we should fix the above issue. Perhaps it's just an old version? Haven't looked at xdebug in ages.

PR #1399, opened 2017-11-06
Clear search context after usage in scroll

mjanser:
Because open search contexts have a cost they should be explicitly
cleared as soon as the scroll is not being used anymore.

The method `clear()` is exposed in `Scroll` to also be able to clear the
search context manually from outside.

### DIFF
--- a/lib/Elastica/Scroll.php
+++ b/lib/Elastica/Scroll.php
@@ -83,9 +83,7 @@ class Scroll implements \Iterator
             $this->_revertOptions();
         } else {
             // If there are no pages left, we do not need to query ES.
-            // Reset scroll ID so valid() returns false.
-            $this->_nextScrollId = null;
-            $this->_currentResultSet = null;
+            $this->clear();
         }
     }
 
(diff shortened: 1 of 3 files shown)


lib/Elastica/Scroll.php
```diff
@@ -135,6 +133,24 @@ public function rewind()
     }
 
     /**
+     * Cleares the search context on ES and marks this Scroll instance as finished.
+     */
+    public function clear()
+    {
+        if (null !== $this->_nextScrollId) {
+            $this->_search->getClient()->request(
```
ruflin: I wonder if there are reasons we should make this optional. Are there use cases where users don't want to remove the search context?

mjanser: Some lines below this one the scrollId is reset (`$this->_nextScrollId = null;`) and without that id the search context can't be used anymore. There is also no getter for the scrollId in the `Scroll` class. Therefore I don't see how the search context could still be useful afterwards.

ruflin: Make sense 👍

ruflin:
@mjanser Merged. Thanks for the 2 PR's.

mjanser:
@ruflin Thanks for merging! Do you plan a new 5.x release soon?

ruflin:
@mjanser TBH so far I wasn't planning a release as I think that is the first PR agains the 5.x branch in quite some time :-) Do you need a release or could you also use 5.x directly?

mjanser:
@ruflin Sure, I can use 5.x directly for now :-)

PR #1408, opened 2017-11-14
Prepare 6.0.0 release

ruflin:
* Run linting
* Update Docker image to Elasticsearch 6.0.0
* Update elasticsearch-php lib dependency to 6.0.0


ruflin:
@p365labs Anything you can think of that we should get in before merging this one and do the release?

PR #1412, opened 2017-11-15
Update to PHPUnit 6

carusogabriel:
As we are on `PHP 7`, we can drop `PHPUnit 5` support and start to use to `PHPUnit 6`.

### DIFF
--- a/env/elastica/composer.json
+++ b/env/elastica/composer.json
@@ -19,8 +19,8 @@
         "pdepend/pdepend":"~2.2",
         "phploc/phploc":"~2.1",
         "phpmd/phpmd":"~2.3",
-        "phpunit/phpunit":"~5.6",
         "sebastian/phpcpd":"~2.0",
+        "phpunit/phpunit":"~6.5",
         "squizlabs/php_codesniffer":"~2.5"
     }
 }
--- a/test/Elastica/Base.php
+++ b/test/Elastica/Base.php
@@ -4,9 +4,11 @@ namespace Elastica\Test;
 use Elastica\Client;
 use Elastica\Connection;
 use Elastica\Index;
+use PHPUnit\Framework\TestCase;
+use PHPUnit\Util\Test;
 use Psr\Log\LoggerInterface;
 
-class Base extends \PHPUnit_Framework_TestCase
+class Base extends TestCase
 {
     protected static function hideDeprecated()
     {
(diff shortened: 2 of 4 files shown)


p365labs:
following this route, I think we could also remove the other stuff which is still related to php5.6 as here #1356

ruflin:
I'm really undecided on this one. I'm aware PHP 5.6 is EOL since quite some time, but that does not seem it's still heavy used (would be interesting to have some numbers on the usage). I really like all the improvements we can do with just supporting 7.x like type hinting or also move to more recent dependencies like in this PR. On the other hand I'm asking myself how many users do we leave behind? So far Elastica is not tested against 5.6 but I think it still works with 5.6. The moment we introduce the first type hinting it will not work anymore. 

How many users will have PHP 5.6 and want to use Elasticsearch 6.x? Is there a correlation between users with old PHP version and not using the most recent version of Elasticsearch?

From a sole development perspective I think we should do the step, remove the old methods, get new version of PHPUnit in and use type hinting and all the other shiny new things.

@p365labs @gabriel-caruso Any thoughts from your side?

gabriel-caruso:
@ruflin There are so many Frameworks, projects and many organizations that are on `PHP 7`. I guess we should too, or, we need to change our `composer.json` to match the tests in `Travis CI` with `PHP 5.x`.

ruflin:
In general I agree that we as library / framework devs can also slightly push users to upgrade to new versions. A quick check on symfony showed the `current` version is still on `5.5.9`: https://symfony.com/doc/current/reference/requirements.html Also there is a slight difference here between libraries which work on their own and libraries that have further external dependencies. In case of Symfony you can't use the most recent 4.0 features if you don't upgrade. In case of Elastica you can't use Elasticsearch 6.0.

Having said all that, upgrading to Elastica 6.0 requires to change your code in most cases, means deploying a new version, testing it and putting it then in production. In theory (which is never true) users should upgrade to PHP 7.0 in such cases as the old version only gets security fixes.

Especially as php-elasticsearch only supports 7.0 and we depend on it, +1 on going all in on 7.0. It helps us to improve the code base too which is good.

Any other thoughts?

gabriel-caruso:
I guess we have no reasons for not upgrading to `PHP 7`. Should I rewrite `Travis CI`? :innocent:

ruflin:
@gabriel-caruso I suggest we move forward was following

* Open a PR which changes Travis and removes the PHP 5.x stuff with a title similar to "Remove 5.6 PHP stuff and support" and mention in the description above. Reason to have in a PR is that there we will get all the message from people which are not happy with it and people that keep an eye on the repo can comment there. I prefer not to hide it in "PHPUnit update PR". 
* Then do follow up PR's to remove all the stuff

gabriel-caruso:
@ruflin Ok. Gonna work on that, get into code, start to refactoring to `PHP 7` and I submit the PR

gabriel-caruso:
@ruflin Actually, everything is already in `PHP 7`. I'm just removing all `PHP 5.x` references, and we should be ready to go.

p365labs:
just 2 things : 
- as @ruflin suggested move to another PR which has a more meaningful title
- remember to remove old deprecated Elastica\Query\Bool and Elastica\Transport\Null

PR #1415, opened 2017-11-21
Remove "<" and ">" as they cannot be escaped

martin-georgiev:
`<` and `>` cannot be escaped according to [the documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#_reserved_characters), that's why try to remove them.

### DIFF
--- a/lib/Elastica/Util.php
+++ b/lib/Elastica/Util.php
@@ -72,12 +72,12 @@ class Util
     }
 
     /**
-     * Replace the following reserved words: AND OR NOT
+     * Replace known reserved words (e.g. AND OR NOT)
      * and
-     * escapes the following terms: + - && || ! ( ) { } [ ] ^ " ~ * ? : \.
+     * escape known special characters (e.g. + - && || ! ( ) { } [ ] ^ " ~ * ? : etc.)
      *
-     * @link http://lucene.apache.org/java/2_4_0/queryparsersyntax.html#Boolean%20operators
-     * @link http://lucene.apache.org/java/2_4_0/queryparsersyntax.html#Escaping%20Special%20Characters
+     * @link https://www.elastic.co/guide/en/elasticsearch/reference/5.1/query-dsl-query-string-query.html#_boolean_operators
+     * @link https://www.elastic.co/guide/en/elasticsearch/reference/5.1/query-dsl-query-string-query.html#_reserved_characters
      *
      * @param string $term Query term to replace and escape
      *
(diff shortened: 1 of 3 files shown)


ruflin:
Thanks for the PR. I wonder if it had any side effects that we tried to escape it? All tests are still green ...

Could you update the changelog?

martin-georgiev:
Changelog updated.

As `<` and `>` were first reported to be inescapable in [v5.1 documentation](https://www.elastic.co/guide/en/elasticsearch/reference/5.1/query-dsl-query-string-query.html#_reserved_characters), it made sense [before that point](https://www.elastic.co/guide/en/elasticsearch/reference/5.0/query-dsl-query-string-query.html#_reserved_characters) in time to escape them. I can only blindly speculate that this somehow relates to the fact that `<` and `>` are Elastic specific and as such they are not inherited from [Lucene's special character](https://lucene.apache.org/core/7_1_0/queryparser/org/apache/lucene/queryparser/classic/package-summary.html#Escaping_Special_Characters) list.

lib/Elastica/Util.php
```diff
@@ -106,11 +106,19 @@ public static function escapeTerm($term)
...
-        foreach ($chars as $char) {
+        foreach ($escapableChars as $char) {
             $result = str_replace($char, '\\'.$char, $result);
         }
+        
+        // < and > cannot be escaped, so they should be removed
+        // @see https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#_reserved_characters
+        $nonEscapableChars = ['<', '>'];
+
+        foreach ($nonEscapableChars as $char) {
```
ruflin: I wonder if we should update line https://github.com/ruflin/Elastica/blob/e47dd0e4832b26127c1275fd7bbe1e3ca6138792/lib/Elastica/Util.php#L77 in some way? Interestingly it already doesn't list < and >.

martin-georgiev: I would drop the list of characters in the description of `replaceBooleanWordsAndEscapeTerm`. My reasoning is that it exposes the internal behaviour of the used submethods. The character list creates maintenance backlog as it is not something that the wrapper method `replaceBooleanWordsAndEscapeTerm` has control over at present.
Yet if you believe it adds value then it needs updating too. Are you looking at any particular wording?

ruflin: +1 on removing it. I'm good with most wordings :-)

martin-georgiev: The method description is updated. I believe we can move this PR forward now.

ruflin:
@martin-georgiev Thanks a lot for the contribution.

martin-georgiev:
@ruflin It was my pleasure.

Issue #1221, opened 2016-11-18
Response getError method returning array


bitgandtter:
Using version 2.3.1
Elasticsearch 5.x

got exception 

> Uncaught PHP Exception Symfony\Component\Debug\Exception\FatalThrowableError: "Wrong parameters for Elastica\Exception\ResponseException([string $message [, long $code [, Throwable $previous = NULL]]])"

reviewing the code i notice that getError method of Response class return an array instead of an string

p365labs:
release 2.3.1 of Elastica library has been released for Elasticsearch version 1.7.3.
Have a look here :
https://github.com/ruflin/Elastica/releases?after=3.0.0-beta1

If you want to use Elasticsearch 5.0.0 you should use Elastica 5.0.0-beta1. Beware that this is a beta version, use it  for testing purposes and not for production one, any feedback will be really 
appreciated.

bitgandtter:
Elastic 5.x its already GA, came out of beta and RC a month ago. This issue seems to be because the new version ways to handle errors. But since Elastic 5.x brings a lot of new enhancements speed among them a lot of people will start to use it. In my personal case i will for a new project, the one throw me this error and pretend to migrate other ones.

So it will be great to handle this new version

p365labs:
If you use Elasticsearch 5.0 you should use **[Elastica 5.0.0-beta1](https://github.com/ruflin/Elastica/releases)**
- getError now return a string :
https://github.com/ruflin/Elastica/blob/8fa05bb677ab49e62e2af3bdfa62f7d9b6de761c/lib/Elastica/Response.php#L82

- the only version of Elastica that was returning an array was from 3.0.0-beta1 to 3.1.0: https://github.com/ruflin/Elastica/commit/99d7cd147c469e20e0c8e550c431712edda4dd19#diff-b364251d02e70310f3c6e443372397d3R81

ruflin:
@bitgandtter As @p365labs state, for Elasticsearch 5.0 you need to use Elastica 5.0. Going to close this issue. In case you have issues with Elastica 5.0 and elasticsearch 5.0 please let us know. In case there are still some questions, feel free to reopen the issue.

PR #1418, opened 2017-12-06
Backport fix for exception messages.

jenkoian:
A fix was made in 3.x branch for exception messages to stop an un-useful
error `Error: Wrong parameters for Exception([string $exception [, long $code [, Exception $previous
  = NULL]]])` appearing. This commit backports that same fix into the
2.x branch.

Here's the commit in question: https://github.com/ruflin/Elastica/commit/99d7cd147c469e20e0c8e550c431712edda4dd19#diff-e1d9f3f0154639bb558fbe6fbe46d323

I appreciate 2.x is unlikely maintained anymore, but due to things outside of my control I can't easily upgrade to 3.x so would appreciate this merge and a new 2.x tag created if possible 🙏 

(I couldn't see a 2.x branch but noticed a `release-2.3` so am opening this against that branch, let me know if this isn't correct)

### DIFF
--- a/lib/Elastica/Response.php
+++ b/lib/Elastica/Response.php
@@ -78,14 +78,43 @@ class Response
      */
     public function getError()
     {
-        $message = '';
-        $response = $this->getData();
+        $error = $this->getFullError();
+        if (!$error) {
+            return '';
+        }
 
-        if (isset($response['error'])) {
-            $message = $response['error'];
+        if (is_string($error)) {
+            return $error;
+        }
+
+        if (isset($error['root_cause'][0])) {
+            $error = $error['root_cause'][0];
+        }
+
+        $message = $error['reason'];
+        if (isset($error['index'])) {
+            $message .= ' [index: '.$error['index'].']';
         }
 
         return $message;
+    }
+
+    /**
+     * A keyed array representing any errors that occured.
+     *
+     * In case of http://localhost:9200/_alias/test the error is a string
+     *
+     * @return array|string Error data
+     */
+    public function getFullError()
+    {
+        $response = $this->getData();
+
+        if (isset($response['error'])) {
+            return $response['error'];
+        }
+
+        return;
     }
 
     /**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,15 +4,10 @@ All notable changes to this project will be documented in this file based on the
 
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/2.3.1...HEAD)
 
-### Backward Compatibility Breaks
-
-### Bugfixes
-
-### Added
-
 ### Improvements
 
-### Deprecated
+- Introduce getFullError() for ResponseSet
+- Use the above to improve the output of getError() on Response
 
 
 ## [2.3.1](https://github.com/ruflin/Elastica/releases/tag/2.3.1) - 2015-10-17
--- a/lib/Elastica/Bulk/ResponseSet.php
+++ b/lib/Elastica/Bulk/ResponseSet.php
@@ -54,6 +54,25 @@ class ResponseSet extends BaseResponse implements \Iterator, \Countable
     }
 
     /**
+     * Returns first found error (full array).
+     *
+     * @return array|string
+     */
+    public function getFullError()
+    {
+        $error = '';
+
+        foreach ($this->getBulkResponses() as $bulkResponse) {
+            if ($bulkResponse->hasError()) {
+                $error = $bulkResponse->getFullError();
+                break;
+            }
+        }
+
+        return $error;
+    }
+
+    /**
      * @return bool
      */
     public function isOk()


ruflin:
What is the Elasticsearch version you are using? Looking at the PR I have the suspicion that you use incompatible versions and getting this in would break things with other versions. But need to have a closer look.

For the `release-2.3` branch it's probably not a good idea to use this one. If we allow such backports we should create a 2.x branch from the most recent 2.x tag we have. The `release-2.3` branch is more a left over from me. If we do a 2.x branch, would that be enough? If possible, I would prefer not to do a new release for such an old version as TBH I do not know what all the above could break.

jenkoian:
> What is the Elasticsearch version you are using?
2.4.6

> Looking at the PR I have the suspicion that you use incompatible versions and getting this in would break things with other versions.
Note, the change is copied verbatim from the 3.0.0 version of this library which claims to support ES versions 2.0.0 and above. Just noticed 2.3.1 claims support for 1.7.2 and above so perhaps there is some additional testing to be done here? Although looking at the change, it seems to be pretty backwards compatible? 

> The release-2.3 branch is more a left over from me. If we do a 2.x branch, would that be enough?
Yeah, only opened to this branch as wanted something which wasn't master and this deemed most appropriate, a 2.x branch would be ideal.

jenkoian:
I should mention it's to work around basically this issue https://github.com/ruflin/Elastica/issues/1221#issue-190403022

ruflin:
I remember we were going forth and back on this. What I think was the conclusion:

* 2.x of Elastica supports 1.x of ES
* 3.x of Elastica supports 2.x of ES

For the exact versions its important to check what is stated in the README it "should" be correct or is at least the one it was tested with. 

ES 1.x has a flat error string and ES 2.x has the nested object. So if we backport this to 2.x it MUST work with both and as `getError` calls here `getFullError` I think this is a breaking change. As it is not recommended/supported to use Elastica 2.x with ES 2.x I'm not sure we should make this change as other things will break. If we go down this route, we must get the for sure the CI running first and I think we should then only update the branch but not tag it.

jenkoian:
Yeah, fair enough, I'd probably take the same stance if in your position. One last attempt to sell it to you if I can, the change is non breaking, it will return early if the error response is a string as per ES 1.x.

jenkoian:
This would also help avoid people doing gross stuff like this: https://stackoverflow.com/a/37762277/380054

ruflin:
@jenkoian If we can make it BC compatible, I think I'm ok in getting it into a branch. But I'm hesitant to do a release. It would be the last release in a major and potentially be the one most used. So far it seems the current release works well so I don't think I should change it. Would be a branch version enough for you?

jenkoian:
@ruflin apologies for long delay in my response. Yeah a branch would work well if that's ok?

ruflin:
Works for me. So what I plan is taking the 2.3.1 tag and push a 2.x branch against which you can open the PR. ok?

jenkoian:
@ruflin sure, thanks!

ruflin:
@jenkoian There should be a 2.x release branch now. Can you open the PR against this one? I think Github recently introduced to edit the base branch of a PR, but never tried it ...

jenkoian:
@ruflin done!

jenkoian:
@ruflin it's added some commits, not sure this is desirable, so let me know if not and i'll rebase or something.

ruflin:
@jenkoian Better rebase and get rid of the commits. Not sure what exacty Github does in the background.

jenkoian:
@ruflin done!

lib/Elastica/Response.php
```diff
@@ -78,14 +78,43 @@ public function __construct($responseString, $responseStatus = null)
      */
     public function getError()
     {
-        $message = '';
+        $error = $this->getFullError();
+        if (!$error) {
```
ruflin: In which case will this line match or not match?

jenkoian: I copied this verbatim from this commit https://github.com/ruflin/Elastica/commit/99d7cd147c469e20e0c8e550c431712edda4dd19#diff-e1d9f3f0154639bb558fbe6fbe46d323 however it looks like this is to cover the case where the response has no error i.e. https://github.com/ruflin/Elastica/pull/1418/files#diff-b364251d02e70310f3c6e443372397d3R117

jenkoian:
@ruflin happy to add to the changelog, as it's not a tag just a branch how should I handle it? Titles as 'unreleased' or titled with the branch name or..?

ruflin:
It would be great if you could add a title "unreleased" as I also have in master.

jenkoian:
Changelog updated!

ruflin:
Merged, thanks for following through with this. Fingers crossed it does not break anything. My knowledge of the 2.x branch is a bit outdated ;-)

Koc:
we are using this backported fix in production, thank you @jenkoian for backport and @ruflin for merging

PR #1420, opened 2017-12-09
Remove 5.6 PHP stuff and support

carusogabriel:
As suggested by @ruflin in #1412


p365labs:
@carusogabriel remember to remove old deprecated 
- Elastica\Query\Bool 
- Elastica\Transport\Null

why did u decide not to update phpunit as u proposed in #1412  ?

carusogabriel:
@p365labs Gonna work on that!

carusogabriel:
There are some other classes deprecated, like:
- `Elastica\Exception\ElasticsearchException`;
- `Elastica\Query\GeohashCell`;

and methods:
- `Elastica\Client::optimizeAll`;
- `Elastica\Client::_log`;

Should we remove them as well?

ruflin:
I think there are 3 different things here:

* Remove of CI things which are required for 5.6
* Improve code because we don't have to support 5.6 anymore and can use 7.0 features
* Remove deprecated functions / methods

I suggest we do not mix the 3 and do separate PR's for each. In case tests exist for 5.6 that are not needed / run on any platform, they should be removed when 5.6 docker images etc. are removed.

lib/Elastica/Multi/ResultSet.php
```diff
@@ -144,7 +144,7 @@ public function offsetExists($offset)
      */
     public function offsetGet($offset)
     {
-        return isset($this->_resultSets[$offset]) ? $this->_resultSets[$offset] : null;
```
ruflin: I don't think this change is related to removing support for 5.6 but an improvement because we can use these features now. As in state in my other comments, lets do such changes in separate PR.

carusogabriel: Okay, gonna rollback this changes, sorry.

carusogabriel:
@ruflin For this PR: `Drop 5.6 Support`. Gonna open other 2 for `Drop deprecated stuff` and other for `Use PHP 7 Null Coalesce Operator`

ruflin:
@carusogabriel Thanks a lot for splitting it up. Really appreciate it.

Based on the comment

>  In case tests exist for 5.6 that are not needed / run on any platform, they should be removed when 5.6 docker images etc. are removed.

we should remove Null and Bool stuff in this PR instead of https://github.com/ruflin/Elastica/pull/1422.

carusogabriel:
Done!

ruflin:
@carusogabriel Thanks a lot. Merged. Could you rebase the other two PR's? Will probably only get to it next week.

ruflin:
@p365labs Feel free to have a look at the other 2 PR's and merge if looks good ;-)

p365labs:
@ruflin  👍

Issue #1374, opened 2017-09-13
BC breaks in minor releases


1ed:
If "This project adheres to Semantic Versioning." as it's said in the changelog, then it should not break BC in minor releases.

p365labs:
Hi @1ed can u explain better what do u mean and of what bc in minor are you talking about ? Tnx a lot!

p365labs:
not sure but I think we can close this.

1ed:
From the changelog:

```
5.3.0

Backward Compatibility Breaks

Removed Query\NumericRange, use Query\Range instead #1334

...

5.1.0

Backward Compatibility Breaks

\Elastica\Script\AbstractScript added the script language as constructor argument and sub-classes must implement getScriptTypeArray
...
3.2.0

Backward Compatibility Breaks

Method \Elastica\ResultSet::create and property \Elastica\ResultSet::$class were removed. To change the ResultSet class, implement your own ResultSet Builder. #1065
Properties on \Elastica\ResultSet _totalHits, _maxScore, _took and _timedOut that were originally set on object construction are now accessed by the getters on the ResultSet. #1065
...
```
If this project adheres to Semantic Versioning these BC breaks should only happen in major releases. If not, then the users of this library are prevented to use permissive version constraints and are in danger of version lock (or at least it is very hard to support multiple versions).

ruflin:
If we do BC breaks in Minors, there is normally a good reason for it. For example for #1334 this was deprecated since 0.90 in elasticsearch and removed in 5.0. As Elastica states it's only compatible with 5.0 it was more an oversight to have the function still in. So this only broke code if someone used Elastica with an unsupported version of Elasticsearch as with 5.x this would not work from the Elasticsearch side.

If I remember the BC break in 5.1.0 correctly it was needed to support some new features and only applied to users which used the AbstractScript type and extended it. As we tie Elastic release to the elasticsearch release, this was a trade off to way 12 month for the 6.0 release or do a small BC break. Did we break semantic versioning with this? Probably yes but I think here the benefits outweighted the potential downsides.

merk:
But why are you tying Elastica releases to ES versions? Theres no need for it? You'd simply have to publish your supported ES version range in README.md and when you decide to drop support for a specific ES version you can just bump a major.

You may have discussed this at some point recently and sorry if its already been debated to death :)

p365labs:
[...]

PR #1422, opened 2017-12-12
Remove deprecated stuff

carusogabriel:
Following #1420


carusogabriel:
Gonna search and fix it!

carusogabriel:
Some if failing, but I couldn't figure out what :disappointed:

p365labs:
yes 'cause there are tests on the this->_log method change it to this->_logger, don't remove it

ruflin:
@carusogabriel Could you rebase on top of master? I just merged https://github.com/ruflin/Elastica/pull/1431

lib/Elastica/Client.php
```diff
@@ -689,8 +689,6 @@ public function request($path, $method = Request::GET, $data = [], array $query
         } catch (ConnectionException $e) {
             $this->_connectionPool->onFail($connection, $e, $this);
 
-            $this->_log($e);
```
stof: shouldn't this be replace by a `$this->_logger->error()`  call ? Overriding the method was deprecated. But now, you removed the logging entirely.

carusogabriel: @ruflin What you say? I'll update this PR today, I forgot it :sweat_smile:

ruflin: +1 on the change suggested by @stof

ruflin:
@carusogabriel Just saw now that you pushed a new commit. Could you also add a changelog? And it seems to break travis somehow.

ruflin:
@carusogabriel Can you have a look at the comment here https://github.com/ruflin/Elastica/issues/1374#issuecomment-366554090 from @greg0ire to check into which "category" your changes fall?

ruflin:
@carusogabriel Can you share some background on the closing of the PR?

carusogabriel:
@ruflin Sure. I closed as I wasn't able to solve all issues and have green tests. I believe you know the library better than me to finish it :smile:

ruflin:
@carusogabriel I see. I was happy that you took that one on :-) If you find time, perhaps you can split up the PR into the parts that you got working so we can get at least these in?

PR #1431, opened 2017-12-21
Fix Travis build by removing php5-xsl dependency.

ruflin:


ruflin:
@p365labs I merged this one to get the builds back to green. I worry a bit about the side effects as I think the packages were need to build the docs. Perhaps they are now part of php7 as I couldn't find the dependency anymore. Will definitively figure it out with the next release in case the docs don't build anymore ...

PR #1432, opened 2017-12-21
Allow new features and bugfixes

greg0ire:
Not sure if pinning to a specific version like this was deliberate.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     "require": {
         "php": "^7.0",
         "psr/log": "~1.0",
-        "elasticsearch/elasticsearch": "6.0.0"
+        "elasticsearch/elasticsearch": "6.0.*"
     },
     "require-dev": {
         "aws/aws-sdk-php": "~3.0",


composer.json
```diff
@@ -14,7 +14,7 @@
     "require": {
         "php": "^7.0",
         "psr/log": "~1.0",
-        "elasticsearch/elasticsearch": "6.0.0"
+        "elasticsearch/elasticsearch": "^6.0"
```
ruflin: As I plan to keep Elastica and php elasticsearch versioning in sync I would suggest to go for `6.0.*` instead. Every time there is a minor on the php elasticsearch side, there should also be a minor of Elastica. So if someone wants new features he also needs to update Elastica. This should make bug tracking easier as we know almost the exact dependency.

@polyfractal Any thoughts here?

greg0ire: That implies you will never ever have a new minor in elastica w/o having a minor in elasticsearch first. Also, that means you can't introduce bc breaks willy nilly, and I think it means you do not follow semver (if there an es major implies an elastica major, that's a bit weird, especially if you introduce no BC-break in your API). Is that really what you want? I think following semver would outweight the benefit of knowing the exact dependency.

curry684: I agree with @greg0ire, as Elasticsearch itself also follows Semver there should never be an issue - even if I used Elastica 6.0.5 with Elasticsearch 6.5.9 Semver guarantees compatibility. So I would definitely only lock the major versions together, and that's what `^6.0` currently would imply.

Also I came here because right now I'm not getting 6.0.1 of the main API, released 29 days ago, because of this constraint   😉

ruflin: @curry684 I assume my suggestion above would already solve your issue that you didn't get the last bugfix release?

About semver, my take on it is here: https://github.com/ruflin/Elastica/issues/1374 But I don't think that really has an affect here.

@greg0ire I think you are right with your comment, that I don't expect Elastica to have any minor release without an elasticsearch minor release. There is quite a bit of history on linking release together. Initially it was, then we had our own versioning and synced it up again. I'm pretty happy so far that we synced it up again.

ruflin:
@greg0ire To move this forward, I suggest we first get the bugfix release in as I agree this is critical also to solve the issue that @curry684 has. Lets conitnue the minor / major discussion in a follow up?

composer.json
curry684: Yep it would   :) 

As for the Semver issue itself - imho you solved it wrong in https://github.com/ruflin/Elastica/issues/1374#issuecomment-335400478. The Semver way to solve the forgotten removed function would've been to mark the function as deprecated in the minor/patch, and replace its implementation with a proper exception to avoid the backend itself crashing, which would itself have been a bugfix. 

This would then automatically make you compliant with Semver and Elasticsearch versioning as it would still imply the current 5.x version (ie. the latest) behaves correctly with the 5.x backend, and make it irrelevant whether the first is 5.2.9 and the second is 5.1.1.

greg0ire:
@ruflin done, let's move the discussion to #1374

ruflin:
@greg0ire Finally merged. Thanks for this an the follow up discussion.

ruflin:
@greg0ire Now that it took me so long to publish 6.0.1 I think you have proven a point and I would like to get back to your initial proposal to allow also future minor releases of php-elasticsearch. Should this break one day, we can still revert it but it's better that people get the newest features / bug fixes from php-elasticsearch without requiring and Elastica release. WDYT?

greg0ire:
I think so too, let me draft something that will explain a new possible workflow in the form of a PR.

greg0ire:
See #1467

curry684:
👍 from me.

PR #1437, opened 2018-01-12
Allow dependency increase

XWB:
Dependency `elasticsearch/elasticsearch` is at version 5.3.2 currently but cannot installed right now.


ruflin:
I just realised travis is not running on this branch as we specified `master` only. Could you open a PR to change this?

XWB:
Sure: https://github.com/ruflin/Elastica/pull/1440

PR #1438, opened 2018-01-15
Rename `prefix_len ` to `prefix_length` which deprecated after v5.0.0

charleslxh:


lib/Elastica/Suggest/Term.php
```diff
@@ -126,4 +126,16 @@ public function setMaxTermFrequency($max)
     {
         return $this->setParam('max_term_freq', $max);
     }
+
+    /**
+     * Set the number of distance which implementation to use for comparing how similar suggested terms are.
```
ruflin: Could we move this into a separate PR?

ruflin:
@charleslxh Any update here? Would be nice to get this in.

deguif:
@charleslxh @ruflin would you mind if I take care of this?

ruflin:
@deguif Would be great if you could take this over.

lib/Elastica/Suggest/Term.php
deguif: And this already exists but method is named `setStringDistanceAlgorithm`, probably added in the meantime.

ruflin:
Closing this in favor of https://github.com/ruflin/Elastica/pull/1708

PR #1445, opened 2018-01-30
Document possible value type

greg0ire:
It seems that providing a single field name is ok.

### DIFF
--- a/lib/Elastica/Aggregation/TopHits.php
+++ b/lib/Elastica/Aggregation/TopHits.php
@@ -66,7 +66,7 @@ class TopHits extends AbstractAggregation
     /**
      * Allows to control how the _source field is returned with every hit.
      *
-     * @param array|bool $params Fields to be returned or false to disable source
+     * @param array|string|bool $params Fields to be returned or false to disable source
      *
      * @return $this
      */
--- a/test/Elastica/Aggregation/TopHitsTest.php
+++ b/test/Elastica/Aggregation/TopHitsTest.php
@@ -279,13 +279,22 @@ class TopHitsTest extends BaseAggregationTest
         $this->assertEquals([2, 4], $resultDocs);
     }
 
+    public function limitedSourceProvider()
+    {
+        return [
+            'string source' => ['title'],
+            'array source' => [['title']],
+        ];
+    }
+
     /**
      * @group functional
+     * @dataProvider limitedSourceProvider
      */
-    public function testAggregateWithLimitedSource()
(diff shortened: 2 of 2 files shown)


lib/Elastica/Aggregation/TopHits.php
```diff
@@ -66,7 +66,7 @@ public function setSort(array $sortArgs)
     /**
      * Allows to control how the _source field is returned with every hit.
      *
-     * @param array|bool $params Fields to be returned or false to disable source
+     * @param array|string|bool $params Fields to be returned or false to disable source
```
ruflin: Interesting. I assume you tested that this works? Could we add a test case?

greg0ire: I'll tried to, and yeah, I built my app that way but phpstan told me I couldn't

ruflin:
Any chance you could add a test like here https://github.com/ruflin/Elastica/blob/master/test/Elastica/Aggregation/TopHitsTest.php#L288 to the PR that tests setSource without an array?

greg0ire:
@ruflin sure, I was a bit overwhelmed lately but now I have the time

greg0ire:
Didn't manage to test locally, but Travis seems to be happy with this

test/Elastica/Aggregation/TopHitsTest.php
```diff
@@ -279,13 +279,22 @@ public function testAggregateTwoDocumentPerTagWithOffset()
         $this->assertEquals([2, 4], $resultDocs);
     }
 
+    public function limitedSourceProvider()
+    {
+        return [
+            'string source' => ['title'],
```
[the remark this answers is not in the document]
greg0ire: No it should not be: this is not the first argument, this is an array of all the arguments, and the first one is a string, not an array.

test/Elastica/Aggregation/TopHitsTest.php
```diff
@@ -279,13 +279,22 @@ public function testAggregateTwoDocumentPerTagWithOffset()
...
+            'array source' => [['title']],
```
greg0ire: @ruflin here is how you pass an array, see the difference?

test/Elastica/Aggregation/TopHitsTest.php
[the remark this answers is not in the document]
ruflin: Got it, reading https://phpunit.de/manual/current/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers also helped :-)

Issue #1446, opened 2018-02-01
Please rename `min_word_len` to `min_word_length` on 5.3.0


shihgangliu:
https://github.com/ruflin/Elastica/blob/5.x/lib/Elastica/Suggest/Term.php#L91-L94

I use `ruflin/elastica` on version `5.3.0`.
There is still deprecated parameter `min_word_len` at `lib/Elastica/Suggest/Term.php#93`.
I hope could fix it, thanks.

XWB:
https://github.com/ruflin/Elastica/pull/1447

shihgangliu:
@XWB thank you.

kubk:
It looks like the issue is fixed and can be closed.

shihgangliu:
Sorry I forgot, thanks help.

PR #1447, opened 2018-02-01
Rename deprecated `min_word_len` to `min_word_length`

XWB:
Fixes https://github.com/ruflin/Elastica/issues/1446


ruflin:
@XWB Thanks for the fix.

PR #1448, opened 2018-02-01
AvgBucket & SumBucket : fix phpdocs + add AggregationTest + Remove spaces in buckets_path

josselinh:


ruflin:
@josselinh Thanks for the change. Will that fix the change in master?

PR #1427, opened 2017-12-19
Request Params passed to Client through Index and Type

akadko:
Client's methods for updating and adding document have non-required array argument requestParams which can be used, for example, for specifying a pipeline. I added the same argument to the corresponding methods of Index and Type classes.

PR #1455, opened 2018-02-09
Add support for specifying a pipeline when indexing documents

### LINKED ISSUE #1453: Support for pipeline when indexing documents
Currently, the indexing methods in `Type`, `Index` and `Bulk` don't support the `pipeline` option.

AFAICT, the only way to trigger a pipeline when indexing is to instantiate the `Elasticsearch\Endpoints\Index` lower-level API myself, duplicating all the logic of `\Elastica\Type::addDocument`. Is there a better way ?

stof:
Closes #1453

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,15 @@
 All notable changes to this project will be documented in this file based on the [Keep a Changelog](http://keepachangelog.com/) Standard. This project adheres to [Semantic Versioning](http://semver.org/).
 
 
-## [Unreleased](https://github.com/ruflin/Elastica/compare/6.0.0...master)
+## [Unreleased](https://github.com/ruflin/Elastica/compare/6.0.1...master)
 
 ### Backward Compatibility Breaks
 
 ### Bugfixes
 
 ### Added
+
+* Added support for pipeline when indexing document. [#1455](https://github.com/ruflin/Elastica/pull/1455)
 
 ### Improvements
 
--- a/lib/Elastica/Document.php
+++ b/lib/Elastica/Document.php
@@ -276,6 +276,34 @@ class Document extends AbstractUpdateAction
     }
 
     /**
+     * Sets pipeline
+     *
+     * @param string $pipeline
+     *
+     * @return $this
+     */
+    public function setPipeline($pipeline)
+    {
+        return $this->setParam('_pipeline', $pipeline);
+    }
+
+    /**
+     * @return string
+     */
+    public function getPipeline()
+    {
+        return $this->getParam('_pipeline');
+    }
+
+    /**
+     * @return bool
+     */
+    public function hasPipeline()
+    {
+        return $this->hasParam('_pipeline');
+    }
+
+    /**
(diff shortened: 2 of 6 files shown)


stof:
@ruflin I added the `setPipeline` in the Document class. Should it be there or in the AbstractUpdateAction ? I don't know whether pipelines make sense for Script.

And I would like some guidance about the best way to cover this with tests.

stof:
@ruflin could you give me the necessary indications so that I can finish this PR ?

ruflin:
@stof Sorry for the late reply, only was partially online the last days.

I think having it in `Document` as you made it make sense as the pipeline is set for each Document.

The testing here is a bit trickier as you have to create a pipeline first in Elasticsearch. I would probably do that through Elastica by using php-elasticsearch as ingest does not really exist yet in Elastica. Best go for a simple processor like rename inside the pipeline which then should also make it easy to test if the document was actually processed as expected.

Cleaning up the pipeline after the test would be great, but I'm also ok if it sticks around if this gets too complicated.

stof:
@ruflin actually, for bulk requests, this goes in the bulk params, not in each document. Writing the test made me realize this mistake in my PR (I'm already using the pipeline when indexing documents with the lower-level API, but I was not using the bulk API).

and the update API does not support pipelines yet (this is discussed in https://github.com/elastic/elasticsearch/issues/17895)

So does it still make sense to have it in the Document object ?

stof:
Tests are written though.

They are currently leaking the rename pipeline, as ensuring that the cleanup is done even in case of test failure would involve wrapping the whole test in a `try/finally` or doing the cleanup in `tearDown` while most tests don't create it. But I use always the same pipeline name (overriding the pipeline defined by previous tests), so I won't leak multiple ones.

ruflin:
@stof Thanks for the details and changes. I will probably only have time to have a detailed look and think about it beginning next week. It's definitively on my todo list ;-)

ruflin:
@stof I'm positively surprised that using a pipeline with the bulk API already works as you showed in your tests. I wonder if that is enough for now or if we need a deeper integration?

For your PR I'm good to merge as is if your are ok with it and would add a Changelog.

stof:
@ruflin it already works, because the client exposes the request params in its API, and https://github.com/ruflin/Elastica/pull/1427 makes it much more usable by allowing to pass them along.

stof:
@ruflin I'm OK with merging this PR as is (and I added the changelog)

CHANGELOG.md
```diff
@@ -2,14 +2,16 @@
 All notable changes to this project will be documented in this file based on the [Keep a Changelog](http://keepachangelog.com/) Standard. This project adheres to [Semantic Versioning](http://semver.org/).
 
 
-## [Unreleased](https://github.com/ruflin/Elastica/compare/6.0.0...master)
+## [Unreleased](https://github.com/ruflin/Elastica/compare/6.0.1...master)
```
ruflin: Good catch

Issue #1461, opened 2018-02-16
Increase a coverage


dmitrijivanenko:
Hey, I would like to contribute to the project. Wanted to start from increasing the coverage, do you need a help with it, since now it is only 84%?

ruflin:
@dmitrijivanenko I'm more then happy about contributions to the test suite. What I would focus on is not pure coverage number but making sure we add tests for code that is complex and has potential to break. The part that is especially valuable are the functional tests as they tell us when something in Elasticsearch changed and we need to fix it in Elastica.

dmitrijivanenko:
@ruflin thank you, got it. I will look for such places.

PR #1462, opened 2018-02-16
#1461: unit tests

dmitrijivanenko:
https://github.com/ruflin/Elastica/issues/1461 Unit tests for `/lib/Elastica/Query/Common.php` and `/lib/Elastica/Query/QueryString.php`


ruflin:
@dmitrijivanenko Thanks for the tests. As mentioned in https://github.com/ruflin/Elastica/issues/1461#issuecomment-366549703 I would suggest to focus on functional tests and unit tests on parts with complex code which is likely to break.

PR #1467, opened 2018-02-18
Desynchronize from driver lib

greg0ire:
This has several benefits:

- Elastica can now make its own BC-breaks and release majors even if
there is no new elasticsearch/elasticsearch version.
- Elastica can benefit from minor features faster.
- Elastica does not have to make new major releases if there is no need
to.
- If Elastica starts supporting several versions of
elasticsearch/elasticsearch, users will have an easier time migrating.
- Elastica users now only have one compatibility matrix to look at, and
don't have to keep in mind that both libs are synchronized, and will let
Composer resolve dependencies like it should.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     "require": {
         "php": "^7.0",
         "psr/log": "~1.0",
-        "elasticsearch/elasticsearch": "6.0.*"
+        "elasticsearch/elasticsearch": "^6.0"
     },
     "require-dev": {
         "aws/aws-sdk-php": "~3.0",


ruflin:
@greg0ire Thanks a lot for taking a stab at this and sorry for the slow response.

I think there are 2 parts here:
* Desync minors / bufixes of the 2 clients
* Desync majors

I'm on board now with the first one (minor decoupling) but feel uncomfortable with decoupling majors. The reasons I feel uncomfortable is less about the version of elasticsearch-php but the elasticsearch version itself. With the minor decoupling we already get few of the benefits above. Ok to go only the first step for now?

For the installation I would hope that users that get started with Elastica do not have to think about elasticsearch-php, not even know that it's a dependency. In each Elastica version we would set the minimum required elasticsearch-php version. So if Elastica 6.1 requires as a minimum elasticsearch-php 6.1 we can say elasticsearch-php >=6.1 in the composer file. Someone updating Elastica will automatically update the elasticsearch-php version potentially even to 6.2 if it's already around.

greg0ire:
@ruflin ok, I changed to desynchronize only minors. That means no changes in the README.

ruflin:
@greg0ire Thanks, merged.

Issue #1429, opened 2017-12-20
Call to a member function toArray() on array


rafaelschmitt:
When i run a query then i get this error `Call to a member function toArray() on array`

My Setup:
----
- symfony/symfony v3.3.8
- friendsofsymfony/elastica-bundle dev-master 96c79eb
- elasticsearch/elasticsearch v6.0.0
- ruflin/elastica 6.0.0
- Elastic-Server Version: 6.0.0

Error:
----
```txt
Call to a member function toArray() on array
```

**Entry-point:** (use of FOSElasticaBundle)
```php
$adapter = $finder->createPaginatorAdapter($aqb->getQuery());
$aggData = $adapter->getAggregations();
```

**stack-trace:**
```txt
Symfony\Component\Debug\Exception\FatalThrowableError:
Call to a member function toArray() on array

  at vendor/ruflin/elastica/lib/Elastica/Query.php:333
  at Elastica\Query->toArray()
     (vendor/ruflin/elastica/lib/Elastica/Search.php:455)
  at Elastica\Search->search()
     (vendor/ruflin/elastica/lib/Elastica/Type.php:356)
  at Elastica\Type->search(object(Query))
     (vendor/friendsofsymfony/elastica-bundle/src/Paginator/RawPaginatorAdapter.php:107)
  at FOS\ElasticaBundle\Paginator\RawPaginatorAdapter->getAggregations()
     (src/ggm/ClassifiedAdBundle/Controller/AdvertController.php:455)
  at ggm\ClassifiedAdBundle\Controller\AdvertController->overviewAction(object(Request), 'literatur')
  at call_user_func_array(array(object(AdvertController), 'overviewAction'), array(object(Request), 'literatur'))
     (vendor/symfony/symfony/src/Symfony/Component/HttpKernel/HttpKernel.php:153)
  at Symfony\Component\HttpKernel\HttpKernel->handleRaw(object(Request), 1)
     (vendor/symfony/symfony/src/Symfony/Component/HttpKernel/HttpKernel.php:68)
  at Symfony\Component\HttpKernel\HttpKernel->handle(object(Request), 1, true)
     (vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php:171)
  at Symfony\Component\HttpKernel\Kernel->handle(object(Request))
     (web/app_dev.php:31)
  at require('/home/kokoloko/projects/php/portal-webapp-2/web/app_dev.php')
[...]

PR #1471, opened 2018-03-01
Do not call toArray method in Query::toArray

XWB:
Fixes https://github.com/ruflin/Elastica/issues/1429, introduced in https://github.com/ruflin/Elastica/pull/1378

Unless I'm missing something, we should not call `toArray` in Query::toArray. Either `post_filter` is already an array, or it is an instance of `AbstractQuery` which implements `Countable` through `Param` anyway.

### DIFF
--- a/lib/Elastica/Query.php
+++ b/lib/Elastica/Query.php
@@ -330,7 +330,7 @@ class Query extends Param
             $this->setQuery(new MatchAll());
         }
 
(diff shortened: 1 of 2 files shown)


ruflin:
@XWB Any chance you could add a test for this to make sure we do not reintroduce this?

deguif:
Closing as it should be fixed by https://github.com/ruflin/Elastica/pull/1950
Second fix is also fixed in master since https://github.com/ruflin/Elastica/pull/1696 was merged

PR #1472, opened 2018-03-01
Added username and password in connections def

mmoreram:
If you set your server configuration by using the connections array instead of the basic config mode, then you must explicitely tell, per each connection, the user and the password.


ruflin:
@mmoreram Based on your change I assume this was mainly a docs issue but already worked before.

Issue #1475, opened 2018-03-20
setRefresh


qqqqb:
I found you are accepting Bool only on refresh at [lib/Elastica/AbstractUpdateAction.php #L328](https://github.com/ruflin/Elastica/blob/master/lib/Elastica/AbstractUpdateAction.php#L328)

While it should be 
### Empty string or true
Refresh the relevant primary and replica shards (not the whole index) immediately after the operation occurs, so that the updated document appears in search results immediately. This should ONLY be done after careful thought and verification that it does not lead to poor performance, both from an indexing and a search standpoint.
### wait_for
Wait for the changes made by the request to be made visible by a refresh before replying. This doesn’t force an immediate refresh, rather, it waits for a refresh to happen. Elasticsearch automatically refreshes shards that have changed every index.refresh_interval which defaults to one second. That setting is dynamic. Calling the Refresh API or setting refresh to true on any of the APIs that support it will also cause a refresh, in turn causing already running requests with refresh=wait_for to return.
### false (the default)
Take no refresh related actions. The changes made by this request will be made visible at some point after the request returns.
[https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-refresh.html](url)

ruflin:
Definitively, we should support the above params. Do you want to open a PR with the change?

PR #1476, opened 2018-03-21
Support setRefresh as wait_for, true, false or empty

qqqqb:


qqqqb:
Ref #1475

ruflin:
Could you add some tests to this PR and entry to the CHANGELOG?

So far it seems this change breaks the existing tests.

qqqqb:
I'm not sure why this failed, What I did is allow to accept the following
'true' by (string) '', (integer) 1, (Boolean) True
'wait_for' by (string) 'wait_for'
'false' (integer) 0, (Boolean) False

And what I wrote is
`if (is_string($refresh) && strlen($refresh) == 0) {
          $refresh = True;
        }
        if ($refresh != 'wait_for') {
          $refresh = $refresh ? 'true' : 'false';
        }`

ruflin:
I haven't checked in detail yet. It could also be that we found an error in the test case that fails. 

In any case it would be helpful to add tests for this that will also make it easier to tell if the implementation is correct.

safwenbs:
Hello, it would be nice to merge this in order to handle the wait_for option as mentionned in the elastic doc

https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update-by-query.html

`refresh
(Optional, enum) If true, Elasticsearch refreshes the affected shards to make this operation visible to search, if wait_for then wait for a refresh to make this operation visible to search, if false do nothing with refreshes. Valid values: true, false, wait_for. Default: false.`

ruflin:
@safwenbs Interested to take the PR over?

deguif:
Closing as https://github.com/ruflin/Elastica/pull/1791 got merged.

Issue #1322, opened 2017-06-07
Reindex API allows more customizations than the Reindex class implements


mweibel:
#1315 implemented the new reindex API but did so in a way that it's impossible to pass all possible options to the underlying API.
Just a few examples of what the API allows but the Elastica implementation doesn't:
- it allows to set query parameters (timeout, requests_per_second, wait_for_completion etc.)
- it allows to adjust the batch size by setting a size on the source
- it allows routing on the destination index
- it allows importing from a remote source
- it allows to copy from multiple sources into one destination

etc.

As it currently is implemented, the reindex class will prevent those possibilities and if that is being released, BC probably needs to be preserved, which is bad. 

IMO the reindex class should just allow two parameters: `options` and `params`. The user of that class needs to specify the source and destination plus other options on it's own instead of relying on Elastica to assemble it.

giovannialbero1992:
For the moment i've deprecated CrossIndex and implemented Reindex with same functionalities of CrossIndex (and little bit more). Sure we can implement in a next time more functionality. 
if you want contribute with some PR you are welcome!

ruflin:
@mweibel Thanks for bring this up. I agree it's a bit bad that this will probably break BC. +1 on implementing it in a more "flexible way" and allowing `setParam` or `_setRawParam` to be used as for most other Elastica objects.

I'm ok with breaking this in a minor release as think the benefit will outweight the pain we will cause. But we will have to add a bold "remark" about it.

p365labs:
Hi @ruflin I think that now CrossIndex has been removed, the ReIndex class as @mweibel suggested should be fully in sync with ES 6.0 implementation, so it will be really useful implementing missing options.

ruflin:
@p365labs +1

ruflin:
@daviscaics Great to hear you are interested in contributing. For the examples best have a look at the Query objects: https://github.com/ruflin/Elastica/tree/master/lib/Elastica/Query They all indirectly extend `Param`: https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Param.php

PR #1494, opened 2018-05-02
Reindex more flexible

daviscaics:
#1322 
Upgraded Reindex class, still keeping it backward compatible.

The `$options` parameter of constructor is still there, but allows only the old set of limited options, it's just  there for BC.
Plus I added some ways to add some options via dedicated setters or setting raw data via `setParam`.

Not very happy with the result because of that cumbersome BC thing. 
Comments are very very welcome.

### DIFF
--- a/lib/Elastica/Reindex.php
+++ b/lib/Elastica/Reindex.php
@@ -3,7 +3,7 @@ namespace Elastica;
 
 use Elastica\Query\AbstractQuery;
 
-class Reindex
+class Reindex extends Param
 {
     const VERSION_TYPE = 'version_type';
     const VERSION_TYPE_INTERNAL = 'internal';
@@ -15,6 +15,8 @@ class Reindex
     const TYPE = 'type';
     const SIZE = 'size';
     const QUERY = 'query';
+    const WAIT_FOR_COMPLETION = 'wait_for_completion';
+    const WAIT_FOR_COMPLETION_FALSE = 'false';
 
     /**
      * @var Index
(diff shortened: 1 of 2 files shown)


lib/Elastica/Reindex.php
```diff
@@ -138,4 +162,62 @@ private function _setSourceType(array $sourceBody)
...
+            self::TIMEOUT => null,
+            self::SCROLL => null,
+            self::REQUESTS_PER_SECOND => null,
+        ]);
+    }
+
+    public function setWaitForCompletion($value)
+    {
+        is_bool($value) && $value = $value ? 'true' : 'false';
+        $this->setParam(self::WAIT_FOR_COMPLETION, $value);
```
ruflin: normally we return `$this` so setters can be chained. Here it would be `return $this->setParam(...` Same below for the other methods.

lib/Elastica/Reindex.php
```diff
@@ -31,53 +38,70 @@ class Reindex
...
     protected $_options;
 
+    /**
+     * @var Response|null
+     */
+    protected $_lastResponse;
+
+    /**
+     * @param array $options - deprecated because not compatible with complete Reindex API
+     */
```
ruflin: How would this look like if you would not keep it BC?

damienalexandre:
Hi @daviscaics, looks like you are not that far from being done with this PR, is there anything I can do to help?

ruflin:
@damienalexandre Feel free to take over the PR ;_)

ruflin:
Closing as https://github.com/ruflin/Elastica/pull/1637 was merged. Thanks @daviscaics for getting this started and @damienalexandre for pushing it over the line.

PR #1497, opened 2018-05-24
Don't use deprecated `inline` field

### LINKED ISSUE #1416: Deprecated field [inline] used, expected [source] instead
Elasticsearch 5.6 sends warning "Deprecated field [inline] used, expected [source] instead" in response header "Warning".

It's caused by https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Script/Script.php#L58

And here is second place where `'source'` should be supported: https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Script/AbstractScript.php#L66

andrejbaran:
Hey there,

this changes script to use `source` field instead of deprecated `inline`.

Fixes #1416

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ All notable changes to this project will be documented in this file based on the
 
 ### Improvements
 
+* Use `source` script field instead of deprecated (since ES 5.6) `inline` field. [#1497](https://github.com/ruflin/Elastica/pull/1497)
+
 ### Deprecated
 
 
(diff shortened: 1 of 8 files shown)


ruflin:
Thanks for the fix. Code LGTM. Could you add a line to the changelog?

andrejbaran:
Hey @ruflin, sure, should it go into `Bugfixes` or `Deprecated` ?

andrejbaran:
I have put it under `Improvements` since that imho fits most :)

ruflin:
@andrejbaran Thanks, LGTM.

ruflin:
Now that I got 2 requests in 1 day, I should probably do one now. Will try to find some time today.

ruflin:
Here you go: http://elastica.io/2018/05/29/release-6-dot-0-2/

PR #1506, opened 2018-06-19
Make ResultSet iterator current() not return a mixed value

Tobion:
Returning potentially false from the iterator current() means each result set item could be `false` which makes it impractical to safely use it.
When using phpstan, you will get errors like `Cannot call method getId() on bool|Elastica\Result.` when you just iterate over the result set.

Also `current()` should not need to check `valid()` again because that is done automatically by php when using foreach. When valid returns false, the loop is ended. So that is part of the Iterator spec. So using current on a non-existing item should not happen unless you misuse the Iterator.

### DIFF
--- a/lib/Elastica/ResultSet.php
+++ b/lib/Elastica/ResultSet.php
@@ -236,15 +236,11 @@ class ResultSet implements \Iterator, \Countable, \ArrayAccess
     /**
      * Returns the current object of the set.
      *
-     * @return \Elastica\Result|bool Set object or false if not valid (no more entries)
+     * @return \Elastica\Result Set object
      */
     public function current()
     {
-        if ($this->valid()) {
-            return $this->_results[$this->key()];
-        }
-
-        return false;
+        return $this->_results[$this->key()];
     }
 
     /**
@@ -253,8 +249,6 @@ class ResultSet implements \Iterator, \Countable, \ArrayAccess
     public function next()
     {
         ++$this->_position;
-
-        return $this->current();
     }
 
     /**
@@ -308,7 +302,7 @@ class ResultSet implements \Iterator, \Countable, \ArrayAccess
      *
      * @throws Exception\InvalidException If offset doesn't exist
      *
-     * @return Result|null
+     * @return Result
      */
     public function offsetGet($offset)
     {


lib/Elastica/ResultSet.php
```diff
@@ -253,8 +249,6 @@ public function current()
     public function next()
     {
         ++$this->_position;
-
-        return $this->current();
```
Tobion: next() is not meant to return something.

lib/Elastica/Bulk/ResponseSet.php
```diff
@@ -95,13 +95,11 @@ public function hasError()
...
     /**
-     * @return bool|\Elastica\Bulk\Response
+     * @return \Elastica\Bulk\Response
      */
     public function current()
     {
-        return $this->valid()
-            ? $this->_bulkResponses[$this->key()]
-            : false;
+        return $this->_bulkResponses[$this->key()];
```
ruflin: I wonder if we should treat this as a bug or breaking change ... Someone re

Tobion: I hope nobody would be affected by this potential bc break as it means people would use the Iterator in a wrong way. As an alternative we could leave the implementation and only change the phpdoc. But that would also be confusing.

ruflin: It reminds me of https://github.com/ruflin/Elastica/issues/1374 Let's ping people in there on what they think.

curry684: I'd take the pragmatic approach here - it could only break code that was already abusing a well-defined mechanism, and I can't think of a really practical example how it would *hurt* existing code, it would more probably *fix* it by accident. So I wouldn't call BC on this.

greg0ire: Looks indeed like a case of xkcd workflow xkcd workflow

ruflin: What would we do without xkcd?

Sounds like we are all agreement we should move forward with this.

@Tobion I suggest we still add an entry to the breaking changes for people to be aware of it.

lib/Elastica/ResultSet.php
Tobion: also it's not documented. and it would also not be really useful because it might return false or null or whatever when the iterator reaches the end which is what happens as it happens before valid is called.

lib/Elastica/Bulk/ResponseSet.php
Tobion: @ruflin done. This is ready from my point of view.

PR #1507, opened 2018-06-22
Fix missing phpdoc types in deleteByQuery

Tobion:
This fixes false phpstan reports like

> Parameter #1 $query of method Elastica\Type::deleteByQuery() expects Elastica\Query|string, Elastica\Query\Ids given.


ruflin:
@Tobion Thanks

PR #1510, opened 2018-06-29
Implement Countable on Objects in PHP7.2 - Backported from #1378

jmsche:
Hi,

As I'm currently stuck with ElasticSearch v1.7, I'd like to submit a PR to backport changes brought by initial PR #1378: Implement Countable on Objects in PHP7.2.

Thanks,

jmsche


ruflin:
@jmsche Thanks for the contribution. Haven't touched the 2.x branch in ages but happy to hear it still works :-)

jmsche:
Thanks for accepting the PR :)

Will you create a new tag/release, or should I target the 2.x branch?

ruflin:
Please use the 2.x branch. If somehow possible I would not like to do an other 2.x release.

jmsche:
I'm wondering if you can still create a release: the package I manage was updated to use "dev-2.x" of ruflin/elastica, but given this I'd have to update all projects to require as well "ruflin/elastica":"@dev", even if "dev-2.x" is mentioned in my package...

Thanks for considering my request.

ruflin:
If I do a new release, you would also have to update all packages? Or what is your line looking like at the moment?

The reason I'm very hesitant to do a release because I worry I might break things that work at the moment. It's a long time ago since I touched the 2.x code base so it's hard to think through the problems that could happen.

jmsche:
If you create a new release, I only have to revert the package I manage to use `^2.3` instead of `dev-2.x`, which allows me not to require `"ruflin/elastica":"@dev"` in all projects.

I understand why you hesitate to create a new release. But, if I'm not mistaken, only two new commits would be released - mine, and https://github.com/ruflin/Elastica/commit/d72afd0d16ab1ccaa0d7061696419f0d3da8c032 - see https://github.com/ruflin/Elastica/compare/2.3.1...2.x

From what I can see, my changes are ok, and after checking the other commit, they also seem legitimate. Both are backports and should be ok no matter the PHP version. The only problem being, I can't manage to run tests on my computer :/

ruflin:
The commit I'm much more worried is actually not yours but the other one as it adds error handling for I think Elasticsearch 5.x and I'm not sure about the side effects to 1.x installations here. I would appreciate if for now we could leave it as unreleased and reconsider if we get many request for it.

Please also keep in mind that 1.7 was EOL beginning of last year: https://www.elastic.co/support/eol

bdlabs:
> @jmsche Thanks for the contribution. Haven't touched the 2.x branch in ages but happy to hear it still works :-)

Sorry but not work this branch must be compatible with PHP 5.3.3 and now is it not.
look this https://github.com/ruflin/Elastica/issues/1559

ruflin:
@jmsche Could you comment on this one. That is what I was worried about doing a release :-( @bdlabs What you can always do is going back one bugfix release to the one you had before.

PR #1523, opened 2018-10-08
Set Content-Type header per Guzzle request

Persata:
Hi there!

I've run into a small issue when using the Guzzle transport on one of our projects.

We're performing a Bulk query (`replaceMany` / `updateDocuments`) followed by an `updateByQuery` request, however the second request passes an incorrect `Content-Type` header to Elasticsearch.

In this particular case, the initial Bulk request instantiates the `Guzzle\Client` with a `Content-Type` header of `application/x-ndjson`:

https://github.com/ruflin/Elastica/blob/07db285b4e89ff274b77df0a74f08fff527ae9a4/lib/Elastica/Transport/Guzzle.php#L159-L166

This header value is then re-used for the `updateByQuery` request, which causes Elasticsearch to return an error since it expects `application/json` for the `updateByQuery` request.

I've changed the logic in the Guzzle transport to set the `Content-Type` header per request, taking it from the `Elastica\Request` object.

I haven't submitted a test with this PR as I'm not sure how best to write one for this scenario (it's tricky to inspect the header internals of the Guzzle transport) - if you'd prefer to have one, I'll happily spend some more time on it.

Let me know if I can provide any more information.

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ All notable changes to this project will be documented in this file based on the
 
 ### Bugfixes
 
+* Guzzle transport now sets the `Content-Type` header per request, rather than reusing the potentially incorrect header from the first request made by the transport
+
 ### Added
 
 * [Field](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html#function-random) param for `Elastica\Query\FunctionScore::addRandomScoreFunction`
--- a/lib/Elastica/Transport/Guzzle.php
+++ b/lib/Elastica/Transport/Guzzle.php
@@ -53,7 +53,7 @@ class Guzzle extends AbstractTransport
     {
         $connection = $this->getConnection();
 
-        $client = $this->_getGuzzleClient($this->_getBaseUrl($connection), $connection->isPersistent(), $request);
+        $client = $this->_getGuzzleClient($this->_getBaseUrl($connection), $connection->isPersistent());
 
         $options = [
             'exceptions' => false, // 4xx and 5xx is expected and NOT an exceptions in this context
@@ -115,12 +115,14 @@ class Guzzle extends AbstractTransport
      */
     protected function _createPsr7Request(Request $request, Connection $connection)
     {
+        $headers = $connection->hasConfig('headers') && \is_array($connection->getConfig('headers'))
+            ? $connection->getConfig('headers')
+            : [];
+
         $req = new Psr7\Request(
             $request->getMethod(),
             $this->_getActionPath($request),
-            $connection->hasConfig('headers') && is_array($connection->getConfig('headers'))
-                ? $connection->getConfig('headers')
-                : []
+            \array_merge($headers, ['Content-Type' => $request->getContentType()])
         );
 
         $data = $request->getData();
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ All notable changes to this project will be documented in this file based on the
 
 ### Bugfixes
 
+* Guzzle transport now sets the `Content-Type` header per request, rather than reusing the potentially incorrect header from the first request made by the transport
```
ruflin: Could you link the PR number here?

lib/Elastica/Transport/Guzzle.php
```diff
@@ -115,12 +115,14 @@ public function exec(Request $request, array $params)
      */
     protected function _createPsr7Request(Request $request, Connection $connection)
     {
+        $headers = $connection->hasConfig('headers') && \is_array($connection->getConfig('headers'))
```
ruflin: What is the `\` doing here before is_array?

ruflin:
@Persata Thank a lot for the contribution with the detailed PR description 👍 

Overall the PR looks sane to me but as you mentioned a test would be nice. One way to tackle this could be using guzzle and do the two calls that returned an error before. So we would not check the headers directly but if the calls are executed correctly. There are quite a few examples for integration tests in the testsuite.

lib/Elastica/Transport/Guzzle.php
Persata: Just a micro-optimisation added by my linter, I'll remove it when I'm adding a test if you'd prefer.

Persata:
Great, thanks for looking this over. I'll have a go at adding an integration test this week. :+1:

lib/Elastica/Transport/Guzzle.php
ruflin: Can you share on how this optimises the code? Want to learn more here.

Persata: I first learned of it when I started using the [EA Inspections plugin](https://github.com/kalessil/phpinspectionsea) for PHPStorm as one of the things it suggests is to fully qualify global function calls.

There's a writeup and some benchmarks [here](https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/), and a few packages for automatically dealing with it - one from [Roave / ocramius](https://github.com/Roave/FunctionFQNReplacer) and one from [nilportugues](https://github.com/nilportugues/php-backslasher).

Granted it's a tiny tiny tiny optimisation but it's become part of my workflow.

ruflin: Interesting, was not aware of this optimisation. Thanks for the details, I would probably skip it in this PR and rather have a follow up PR that adds it in all parts of the code base to be consistent. It seems that is something that could be done with https://github.com/nilportugues/php-backslasher in an automated way?

killerwolf:
the current issue should be fixed with this commit https://github.com/ruflin/Elastica/commit/6feb970d3ffbc7235f0080cd5367e3657ea3f82d

ruflin:
@killerwolf Does this mean we should close this PR as the above was merged into master?

deguif:
Closing as there is no activity there since august 2019.
Feel free to reopen.

PR #1540, opened 2018-11-28
Use builtin functions from global namespace

massimilianobraglia:
Hey all, thank you for Elastica!

As stated here https://github.com/ruflin/Elastica/pull/1523#discussion_r225459352 (deeper informations can be found here: https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/), calling builtin functions from the global namespace can improve OP cache performance.

I think I've just covered all function calls (or I hope so 😄)

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -1,32 +1,30 @@
 <?php
 
-$finder = Symfony\CS\Finder\DefaultFinder::create()
-    ->in(['lib', 'test']);
+$finder = PhpCsFixer\Finder::create()
+    ->in([__DIR__])
+    ->exclude(['vendor', 'var'])
+    ->notPath('/cache/')
+;
 
-$config = Symfony\CS\Config\Config::create()
-    ->setUsingCache(true)
-    ->level(Symfony\CS\FixerInterface::SYMFONY_LEVEL)
-    ->fixers([
-        // [contrib] Multi-line whitespace before closing semicolon are prohibited.
-        'multiline_spaces_before_semicolon',
-        // [contrib] There should be no blank lines before a namespace declaration.
-        'no_blank_lines_before_namespace',
-        // [contrib] Ordering use statements.
-        'ordered_use',
-        // [contrib] Annotations should be ordered so that param annotations come first, then throws annotations, then return annotations.
-        'phpdoc_order',
-        // [contrib] Arrays should use the short syntax.
-        'short_array_syntax',
-        // [contrib] Ensure there is no code on the same line as the PHP open tag.
-        'newline_after_open_tag',
-        // [contrib] Use null coalescing operator ?? where possible
-        'ternary_to_null_coalescing',
-        // [contrib] There should not be useless else cases.
-        'no_useless_else',
-        // [contrib] Use dedicated PHPUnit assertions for better error messages.
-        '@PHPUnit60Migration:risky' => true,
-        //'php_unit_dedicate_assert' => ['target' => 'newest'],
+return PhpCsFixer\Config::create()
+    ->setFinder($finder)
+    ->setRules([
+        '@PSR2' => true,
+        '@Symfony' => true,
+        'psr0' => false,
+        'array_syntax' => ['syntax' => 'short'],
+        'concat_space' => ['spacing' => 'one'],
+        'blank_line_after_opening_tag' => false,
+        'lowercase_cast' => true,
+        'lowercase_constants' => true,
+        'lowercase_keywords' => true,
+        'no_trailing_comma_in_singleline_array' => true,
+        'no_unused_imports' => true,
+        'ordered_imports' => true,
+        'native_function_invocation' => true,
+        'php_unit_test_case_static_method_calls' => ['call_type' => 'self'],
+        'php_unit_dedicate_assert' => ['target' => 'newest'],
+        'ternary_to_null_coalescing' => true,
+        'phpdoc_order' => true,
     ])
-    ->finder($finder);
-
-return $config;
+;
--- a/env/elastica/composer.json
+++ b/env/elastica/composer.json
@@ -13,14 +13,14 @@
     ],
     "require": {
         "php": "^7.0",
-        "phpdocumentor/phpdocumentor":"~2.8",
-        "fabpot/php-cs-fixer":"1.11.*",
-        "mayflower/php-codebrowser":"~1.1",
-        "pdepend/pdepend":"~2.2",
-        "phploc/phploc":"~2.1",
-        "phpmd/phpmd":"~2.3",
-        "phpunit/phpunit":"~5.6",
-        "sebastian/phpcpd":"~2.0",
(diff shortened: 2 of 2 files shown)


p365labs:
Hi @massimilianobraglia tnx for the PR! 
This is a sux idea, I wonder if this could be a process to use during release instead of doing it on all classes. Considering tools like [php-backslasher](https://github.com/nilportugues/php-backslasher) or the one from [Ocramius](https://github.com/Roave/FunctionFQNReplacer) still exists I think it would be useful for us and the project to have this type of tools in pipeline during release. 
After linting the code we can also use this kind of tools. 

@ruflin & @Tobion ?

ruflin:
Would be nice if we could automate this. Also for linting, we should probably do it on each PR and not only on release time.

massimilianobraglia:
Hey @p365labs and @ruflin!

Regarding backlashing:
When using https://github.com/nilportugues/php-backslasher all builtin elements will be backlashed (even literals like `null` or `true`/`false`). 
As they are keywords, those should not be backslashed in my opinion.

I'll dig into the package to see if it is configurable some way.

The tool made by Ocramius is worthy as it only backslashes the function calls. But (as stated here https://github.com/Roave/FunctionFQNReplacer#usage) it will break the coding style.

For linting probably Elastica should adopt https://github.com/FriendsOfPHP/PHP-CS-Fixer 😄 

I agree with @ruflin regarding linting and backslashing every PR

ruflin:
We use `php-cs`: https://github.com/ruflin/Elastica/blob/master/.php_cs It's just not run on build time (but we should). 

Keep us posted on your findings.

massimilianobraglia:
Oh. Sorry about that, I must be blind!

Anyway, ok I'll keep you updated.

Il giorno mar 4 dic 2018, 08:42 Nicolas Ruflin <notifications@github.com>
ha scritto:

massimilianobraglia:
Unfortunately, https://github.com/nilportugues/php-backslasher is not configurable

massimilianobraglia:
Sorry about the delay in my answer. This week is a little busy 😞

I just found that with the last version of PHP-CS-Fixer it is possible to automate backslashing by using the `'native_function_invocation' => true` rule. 

Am I wrong or Elastica uses version 1 of PHP-CS-Fixer?
If yes, maybe we can define a new set of CS rules in this PR.

ruflin:
If we need a newer version of php-cs fixer this should not be an issue, we can just update it here: https://github.com/ruflin/Elastica/blob/master/env/elastica/composer.json#L17

massimilianobraglia:
ok then. I'm going to propose a new rule set for CS using the last PHP-CS-Fixer version

massimilianobraglia:
I have just updated the PR with a new rule set. @ruflin can you give it a look?

massimilianobraglia:
Ok, I'm gonna leave the comments. Anyway this is my proposal of CS trying to keep the Elastica CS and merge it with PSR2. Some rules from the version 1 and version 2 changed name or theirs value

p365labs:
@massimilianobraglia well done!
@ruflin this should be also solved the problem from #1535 

The ruleset config war from new Php-CS-Fixer version.

p365labs:
@massimilianobraglia looks like u have conflicts on composer.json versions. Remember if you want to check it out locally you can execute tests quickly on u're development machine:

```make tests```

Remember tests are executed against PHP 7.0, PHP 7.1, PHP 7.2, and PHP 7.3RC6 (stable PHP 7.3 waiting to be merged). Locally by default is executed against the lower version tested *PHP 7.0*.
If you want to execute tests against a specific version

```make tests TARGET=73```

and set the value for the specific PHP environmnet

**Comment by ruflin** [comment]
We should add to this PR running `make lint` to see the actual changes (it's not run by CI). What I worry is that the change will be massive.

It would be nice if we perhaps can split this up into 2 PR's:

* Migrate CS fixer to version 2 with as little changes as possible and run `make lint`
* Add new features to CS fixer 2 and run `make lint`

Like this we see what is actually new.

**Comment by massimilianobraglia** [comment]
Thank you @p365labs! I'll check it ASAP 😄

**Comment by ruflin** [comment]
If nothing really changed, why would then there be a massive PR? We can try with 1 and then see what happens :-)

**Comment by p365labs** [comment]
If I've understand well we should : 

- make lint before travis build
- make lint before merging every PR

am I right ? 

@ruflin & @massimilianobraglia ?

**Comment by ruflin** [comment]
@p365labs Moving forward, that would be great. The reason I have not enabled the check so far is because just running `make lint` requires docker to be around which is not the case on all dev machines and could be a burden for contributors. But with php-cs fixer as more and more a standard we could also offer a way outside docker to run it and get it green.

**Comment by massimilianobraglia** [comment]
@ruflin there will be a massive change because probably nearly all files will be touched with the new CS rules 😄 

@p365labs I agree with @ruflin :)

In my case, I have the fixer installed on my local machine (and I can run it outside docker). 
Should I change the `lint` make rule with the following?

Before:
```sh Makefile
lint:
	${RUN_ENV} php-cs-fixer fix
```

After:
```sh
	ifeq (, $(shell which php-cs-fixer))
		$(error "Could not find php-cs-fixer in $(PATH). Exiting...")
	endif
	php-cs-fixer fix --allow-risky=yes
```

p365labs:
@ruflin & @massimilianobraglia  mmm so let me understand :
The idea is to execute tests locally with docker but not the linter with it ?
The only way a developer has to do functional tests on our stack is to
have Docker setup, otherwise tests will fail. So I cannot see the need to execute
the linter outside docker. 

Also if you use this notation :
```bash
ifeq (, $(shell which php-cs-fixer))
		$(error "Could not find php-cs-fixer in $(PATH).")
		exit 1;
	endif
	php-cs-fixer fix --allow-risky=yes
```

linting will be always executed locally with an unknown version of the linter.
Linting inside the docker forces al the developers using the same Fixer 
'cause it's inside composer.json.

As for example the #1535 add some configs which fails on our LINT because 
That configs were for Php-CS-Fizer 2.x.

Not sure this is the most viable way, but I'm open to understand the best way to do
it.

massimilianobraglia:
@p365labs your is absolutely the right point IMO. 
The snippet of my rule lint I posted on my last comment was just an example and obviously editable.

+1 for having the linter in docker

ruflin:
Ok, lets keep the linter in docker only then. Like this we also ensure that everyone uses the exact same version and should make it easy to execute on CI. If people struggle with it we can still introduce `make lint-local` as a backup.

For the large PR: I'm ok if all files are touched in a PR but I hope we can make the move to version 2 as minimal as possible (mostly adjusting indentations etc) and then add more rules in separate PR's. The part I don't fully understand is why the move to php-cs fixer 2 means big changes.

Perhaps best open a PR with the new linting applied and we can discuss it there?

p365labs:
@ruflin yep agree

massimilianobraglia:
all right then. I'm going to open multiple PRs!

PR #1542, opened 2018-11-28
[WIP] Typehintings

massimilianobraglia:
Hey there, it's me again 😄
Since PHP 7.0 has been introduced, I think Elastica should use strict typehintings in method signatures.

This is a WIP PR.
For now I have type hinted `Elastica\Aggregation` and `Elastica\Bulk` namespaces.

I suggest also to use `declare(strict_types=1);` directive (http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration.strict)


ruflin:
In general I like the idea. I don't think it will be a problem for backports as we rarely do backports. Probably also a good time as security support ends for 7.0 in 3 days: http://php.net/supported-versions.php

What I wonder is how we do the full migration and don't have errors in :-) If we introduce a type but it was an other one or we even allowed 2 different ones (but 1 was documented) we break things.

massimilianobraglia:
In order to not do BC we can leave the `mixed` type whenever it is documented so. I have already used this strategy:
https://github.com/ruflin/Elastica/pull/1542/commits/f0f3f6d52abb0ef263a3fb8db4f370b583e3126d#diff-839167e54614a2676094c9f11220a039R109

```php
class TopHits
{
    // [...]

    /**
     * Set script fields.
     *
     * @param array|\Elastica\Script\ScriptFields $scriptFields
     *
     * @return $this
     */
    public function setScriptFields($scriptFields): self
    {
        if (is_array($scriptFields)) {
            $scriptFields = new ScriptFields($scriptFields);
        }

        return $this->setParam('script_fields', $scriptFields);
    }
 
   // [...]
}
```

`TopHits::setScriptFields` accepts array and ScriptFields.

In this case we should not BC the signature and leave the unspecified type IMO.

That said, as PHP 7.0 security support ends soon as stated by @ruflin, should Elastica consider an upgrade?

massimilianobraglia:
Obviously, using typehintings permits to remove no more useful variable type checks in method parameters (I'm mentioning this because the codebase sometime contains explicit type checks with exception and there are sometime also tests about it)

ruflin:
@massimilianobraglia I still wonder what our best path forward here is to make sure we keep having reviewable PR's but still can do batch changes. One idea: Create a meta issue to track the work and for example do PR's per package?

massimilianobraglia:
If you want me to create multiple PRs (one per namespace) for this one there are no problems :)
I probably need to leave Interfaces like `ArrayableInterface` at the end (since I cannot set the return type of `toArray` method as it is used by multiple classes across namespaces).

Did I understand correctly?

ruflin:
Yes, that would be great.

massimilianobraglia:
Ok! You can track the work in progress in this issue https://github.com/ruflin/Elastica/issues/1548 😄 
I'm closing this in favour of the others!

Issue #1541, opened 2018-11-28
Suggestions on multiple types/fields return results only for the last suggestion


ShytN1k:
Hello!
I've used your lib not for a long, but it helped me a lot. Thank you!
I have a question about suggestions on multiple types/fields.
I tried to get request with curl.
```
 curl -X GET localhost:9200/index/_search?pretty -d '
 {
     "suggest" : {
         "first_suggestion" : {
             "prefix" : "abe",
             "completion" : {
                 "field" : "first_suggestion"
             }
         },
         "second_suggestion" : {
             "prefix" : "abe",
             "completion" : {
                 "field" : "second_suggestion"
             }
         }
     }
 }'
```
and I got expected result with results for the both completions:
```
{
  "took" : 1,
  "timed_out" : false,
  "_shards" : {
    "total" : 5,
    "successful" : 5,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : 0,
    "max_score" : 0.0,
    "hits" : [ ]
  },
  "suggest" : {
    "second_suggestion" : [ ... ],
    "first_suggestion" : [ ... ]
  }
}
```
So now I try to do something similar to this in my code:
```
        $firstCompletion = new Completion('suggest', 'first_suggestion');
        $firstCompletion->setText('abe');

        $secondCompletion = new Completion('suggest', 'second_suggestion');
        $secondCompletion->setText('abe');

        $suggest = new Suggest();
        $suggest->addSuggestion($firstCompletion);
        $suggest->addSuggestion($secondCompletion);
        $query = Query::create($suggest);
        $resultSet = $index->search($suggest);
```
But I get only reasons for the last provided suggestion. Dump of the result:
```
array (size=1)
  'suggest' => 
    array (size=1)
      0 => 
        array (size=4)
          'text' => string 'abe' (length=3)
          'offset' => int 0
          'length' => int 3
          'options' => 
            array (size=1)
              0 => 
                array (size=5)
                  'text' => string 'abeda' (length=5)
                  '_index' => string 'index' (length=5)
                  '_type' => string 'second' (length=6)
                  '_id' => string '4' (length=1)
                  '_score' => float 1
```
I tried to switch lines `$suggest->addSuggestion($firstCompletion);` and `$suggest->addSuggestion($secondCompletion);` and I got results for the first completion. 
[...]

PR #1543, opened 2018-11-30
update Term Suggester, test multiple completion suggestions

p365labs:
In reference #1541 I made some more tests, I also addedd a method for choosing the distance algorithm


p365labs:
This PR in fact add a test in order to verify the #1541 to have a use case that cover that example.
I've also added a method 

```php
public function setStringDistanceAlgorithm($distanceAlgorithm)
```
and its test to include this [option](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-suggesters-term.html#_other_term_suggest_options) which was not in the class.

ruflin:
Code LGTM. Want to add a changelog entry?

p365labs:
@ruflin  done :)

PR #1545, opened 2018-12-04
Add php73 support and Xdebug

p365labs:
- Added support for php7.3rc6 (it will be released in dicember)
- Updated to XDebug 2.6.x in php 7.0,7.1 and 7.2
- add XDebug 2.7.1Beta into


p365labs:
The idea behind php 7.3 is to having it already running our test suite and check if there are problems, considering it will be released in December

ruflin:
Glad to see it's all green.

Issue #1548, opened 2018-12-09
PHP 7 Typehintings Tracker


massimilianobraglia:
In this issue I will update the work in progress.

Namespaces:
- [x] `Elastica\Aggregation` (https://github.com/ruflin/Elastica/pull/1549)
- [x] `Elastica\Bulk` (https://github.com/ruflin/Elastica/pull/1550)
- [x] `Elastica\Cluster` (https://github.com/ruflin/Elastica/pull/1551)
- [x] `Elastica\Connection` (https://github.com/ruflin/Elastica/pull/1560)
- [x] `Elastica\Exception` (https://github.com/ruflin/Elastica/pull/1568)
- [x] `Elastica\Index` (https://github.com/ruflin/Elastica/pull/1569)
- [x] `Elastica\Multi` (https://github.com/ruflin/Elastica/pull/1576)
- [x] `Elastica\Node` (https://github.com/ruflin/Elastica/pull/1577)
- [x] `Elastica\Processor` (https://github.com/ruflin/Elastica/pull/1579)
- [x] `Elastica\Query` (https://github.com/ruflin/Elastica/pull/1582)
- [x] `Elastica\QueryBuilder` (https://github.com/ruflin/Elastica/pull/1583)
- [x] `Elastica\Rescore` (https://github.com/ruflin/Elastica/pull/1566)
- [x] `Elastica\ResultSet` (https://github.com/ruflin/Elastica/pull/1565)
- [x] `Elastica\Script` (https://github.com/ruflin/Elastica/pull/1564)
- [x] `Elastica\Suggest` (https://github.com/ruflin/Elastica/pull/1562)
- [x] `Elastica\Transport` (https://github.com/ruflin/Elastica/pull/1558)
- [x] `Elastica\Type` (https://github.com/ruflin/Elastica/pull/1557)
- [ ] `Elastica` base namespace

Also tests will be updated.

osavchenko:
May I help with this feature?

p365labs:
@massimilianobraglia maybe you and @osavchenko could split the work in two and divide the effort ?
what do u think ?

massimilianobraglia:
That's a good idea @p365labs! Thank you, @osavchenko 

I was processing `Connection` namespace, so you can start with the others uncheck

osavchenko:
@massimilianobraglia I'll start from the end of the list and we will meet each other in the middle ;)

massimilianobraglia:
@osavchenko fine for me! 😄

[...]

PR #1558, opened 2018-12-13
php7 typehints for Elastica\Transport namespace

osavchenko:
See #1548

### DIFF
--- a/lib/Elastica/Transport/AwsAuthV4.php
+++ b/lib/Elastica/Transport/AwsAuthV4.php
@@ -15,7 +15,7 @@ use Psr\Http\Message\RequestInterface;
 
 class AwsAuthV4 extends Guzzle
 {
-    protected function _getGuzzleClient($baseUrl, $persistent = true, Request $request)
+    protected function _getGuzzleClient(string $baseUrl, bool $persistent = true, Request $request): Client
     {
         if (!$persistent || !self::$_guzzleClientConnection) {
             $stack = HandlerStack::create(GuzzleHttp\choose_handler());
@@ -33,7 +33,7 @@ class AwsAuthV4 extends Guzzle
         return self::$_guzzleClientConnection;
     }
 
-    protected function _getBaseUrl(Connection $connection)
+    protected function _getBaseUrl(Connection $connection): string
     {
         $this->initializePortAndScheme();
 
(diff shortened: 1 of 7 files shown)


lib/Elastica/Transport/AwsAuthV4.php
```diff
@@ -14,7 +14,7 @@
 
 class AwsAuthV4 extends Guzzle
 {
-    protected function _getGuzzleClient($baseUrl, $persistent = true, Request $request)
+    protected function _getGuzzleClient($baseUrl, $persistent = true, Request $request): Client
```
massimilianobraglia: missing `string` and `bool`

lib/Elastica/Transport/NullTransport.php
```diff
@@ -26,9 +27,7 @@ class NullTransport extends AbstractTransport
     /**
      * Set response object the transport returns.
      *
-     * @param \Elastica\Response $response
-     *
-     * @return $this
+     * @return Response
```
massimilianobraglia: Missing `: Response` return type

lib/Elastica/Transport/AwsAuthV4.php
osavchenko: With that typehints it's incompatible with parent class

lib/Elastica/Transport/NullTransport.php
osavchenko: By default, it trying to return null, but nullable typehints allowed only in php 7.1

massimilianobraglia: Can't you use the method `generateDefaultResponse` in order to get a Response object?

lib/Elastica/Transport/AwsAuthV4.php
massimilianobraglia: You should typehint the parent class, no? That class belongs to `Elastica\Transport` namespace

osavchenko: Ah yes, missed that. Thank you!

lib/Elastica/Transport/NullTransport.php
osavchenko: Sure

p365labs:
you just need to execute the linter : 
```make lint```

from our command line.

p365labs:
tnx @osavchenko !

PR #1564, opened 2018-12-16
php7 typehints for Elastica\Script namespace

osavchenko:
See #1548

### DIFF
--- a/lib/Elastica/Script/ScriptFields.php
+++ b/lib/Elastica/Script/ScriptFields.php
@@ -15,7 +15,7 @@ use Elastica\Param;
 class ScriptFields extends Param
 {
     /**
-     * @param \Elastica\Script\Script[]|array $scripts OPTIONAL
+     * @param Script[]|array $scripts OPTIONAL
      */
     public function __construct(array $scripts = [])
     {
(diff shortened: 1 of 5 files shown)


lib/Elastica/Script/ScriptFields.php
```diff
@@ -1,7 +1,11 @@
 <?php
+
+declare(strict_types=1);
```
massimilianobraglia: please, as stated here https://github.com/ruflin/Elastica/issues/1548#issuecomment-447821664, remove the strict_types declaration

osavchenko: In this case, we should remove one test because `int(0)` converts to `string("0")` silently

massimilianobraglia: Can you tell me where this happen? I can't find it :(

osavchenko: [Here](https://github.com/ruflin/Elastica/blob/82c640445ec754f4b4fcd04d3a787ef43b47e360/test/Elastica/Script/ScriptFieldsTest.php#L63) is that test.

p365labs: @osavchenko can you explain me better the problem ?

osavchenko: @p365labs as you can see from that test we are passing a plain array and it will have numeric indices (0, 1, 2, ...). Without typization we have check that key should be string and nut empty. With typization numeric array index will be converted silently to string ("0", "1", "2", ...) and that test will not pass.

PR #1566, opened 2018-12-16
php7 typehints for Elastica\Rescore namespace

osavchenko:
See #1548


lib/Elastica/Rescore/AbstractRescore.php
```diff
@@ -29,7 +29,7 @@ protected function _getBaseName()
      *
      * @return $this
      */
-    public function setWindowSize($size)
+    public function setWindowSize(int $size): AbstractRescore
```
p365labs: window_size default value ```10```

PR #1568, opened 2018-12-17
Typehinted exception namespace

massimilianobraglia:
see #1548


lib/Elastica/Exception/ConnectionException.php
```diff
@@ -12,23 +12,23 @@
...
 
     /**
      * Construct Exception.
      *
-     * @param string             $message  Message
-     * @param \Elastica\Request  $request
-     * @param \Elastica\Response $response
+     * @param string   $message
+     * @param Request  $request
+     * @param Response $response
```
p365labs: maybe ```null``` value as default too ?
- @param Request  $request
- @param Response $response

PR #1571, opened 2018-12-18
Implement Countable on Objects

mgallego:
Implements Countable interface in objects to run Elastica 3.x with PHP7.2 version.

Based on PR. https://github.com/ruflin/Elastica/pull/1378

**This is a WIP PR because the tests fails on local without changing code and I need to test through Travis**

The test errors are of kind "Elastica\Exception\ResponseException: all shards failed" and not alwais fails

PR #1573, opened 2018-12-18
Added 3.x branch to Travis execution

mgallego:
Added 3.x branch in Travis configuration.

Requested on https://github.com/ruflin/Elastica/pull/1571#issuecomment-448149818

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ services:
 branches:
   only:
     - master
+    - 3.x
 
 env:
   - TARGET="54"


mgallego:
The current code break the tests. This is the reason because I wanted to try with Travis, to discard local problems.

ruflin:
Interesting that it failed on 3 out of 4 PH versions :-(

ruflin:
@mgallego Any idea how to fix these?

mgallego:
> @mgallego Any idea how to fix these?

One of the problems is related with a library that does'nt exists in docker image, https://github.com/ruflin/Elastica/blob/3.x/env/elastica/Docker70#L12. Removing this dependency the execution continue, but tests broken in my local, I guess that in Travis too.

mgallego:
In 5.6 the same problem: `E: Unable to locate package php5-xsl`

In 5.4 looks like a memory problem: 
`Fatal error: Allowed memory size of 1073741824 bytes exhausted (tried to allocate 64 bytes) in phar:///usr/local/bin/composer/src/Composer/DependencyResolver/Rule.php on line 76`

ruflin:
@mgallego Would be great if you figure out the issue, if you could push it to this PR too. I wonder if we can still fix all of them as some of these PHP versions are EOL for quite some time so could be some libs are not around anymore or broken for these versions.

mgallego:
> @mgallego Would be great if you figure out the issue, if you could push it to this PR too. I wonder if we can still fix all of them as some of these PHP versions are EOL for quite some time so could be some libs are not around anymore or broken for these versions.

I'll try to remove the version errors and execute the tests again to check the unit tests.

mgallego:
Fixed!

PR #1579, opened 2018-12-19
Typehinted processor namespace

massimilianobraglia:
see #1548

### DIFF
--- a/lib/Elastica/Processor/Attachment.php
+++ b/lib/Elastica/Processor/Attachment.php
@@ -2,8 +2,19 @@
 
 namespace Elastica\Processor;
 
+/**
+ * Elastica Attachment Processor.
+ *
+ * @author Federico Panini <fpanini@gmail.com>
+ *
+ * @see https://www.elastic.co/guide/en/elasticsearch/plugins/current/ingest-attachment.html
+ */
 class Attachment extends AbstractProcessor
 {
+    const DEFAULT_TARGET_FIELD_VALUE = 'attachment';
+    const DEFAULT_INDEXED_CHARS_VALUE = 100000;
+    const DEFAULT_IGNORE_MISSING_VALUE = false;
+
     /**
      * Attachment constructor.
      *
@@ -21,7 +32,7 @@ class Attachment extends AbstractProcessor
      *
      * @return $this
      */
-    public function setField(string $field)
+    public function setField(string $field): self
     {
         return $this->setParam('field', $field);
     }
@@ -33,7 +44,7 @@ class Attachment extends AbstractProcessor
      *
      * @return $this
      */
-    public function setTargetField(string $targetField)
+    public function setTargetField(string $targetField): self
     {
         return $this->setParam('target_field', $targetField);
     }
@@ -45,7 +56,7 @@ class Attachment extends AbstractProcessor
      *
      * @return $this
      */
-    public function setIndexedChars(int $indexedChars)
+    public function setIndexedChars(int $indexedChars): self
     {
         return $this->setParam('indexed_chars', $indexedChars);
     }
@@ -57,19 +68,19 @@ class Attachment extends AbstractProcessor
      *
      * @return $this
      */
-    public function setProperties(array $properties)
+    public function setProperties(array $properties): self
     {
         return $this->setParam('properties', $properties);
     }
 
     /**
-     * Set ignore_missing. Default false.
+     * Set ignore_missing. Default value false.
      *
      * @param bool $ignoreMissing
      *
      * @return $this
      */
-    public function setIgnoreMissing(bool $ignoreMissing)
+    public function setIgnoreMissing(bool $ignoreMissing): self
     {
         return $this->setParam('ignore_missing', $ignoreMissing);
     }
--- a/lib/Elastica/Processor/Date.php
+++ b/lib/Elastica/Processor/Date.php
@@ -5,12 +5,16 @@ namespace Elastica\Processor;
 /**
  * Elastica Date Processor.
  *
- * @author   Federico Panini <fpanini@gmail.com>
+ * @author Federico Panini <fpanini@gmail.com>
  *
  * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/date-processor.html
  */
 class Date extends AbstractProcessor
 {
+    const DEFAULT_TARGET_FIELD_VALUE = '@timestamp';
+    const DEFAULT_TIMEZONE_VALUE = 'UTC';
+    const DEFAULT_LOCALE_VALUE = 'ENGLISH';
+
     /**
      * Date constructor.
      *
@@ -30,7 +34,7 @@ class Date extends AbstractProcessor
      *
      * @return $this
      */
-    public function setField(string $field)
+    public function setField(string $field): self
     {
         return $this->setParam('field', $field);
     }
@@ -42,7 +46,7 @@ class Date extends AbstractProcessor
      *
      * @return $this
      */
-    public function setFormats(array $formats)
+    public function setFormats(array $formats): self
     {
         return $this->setParam('formats', $formats);
     }
@@ -54,7 +58,7 @@ class Date extends AbstractProcessor
      *
      * @return $this
      */
-    public function setTargetField(string $targetField)
+    public function setTargetField(string $targetField): self
     {
         return $this->setParam('target_field', $targetField);
     }
@@ -66,7 +70,7 @@ class Date extends AbstractProcessor
      *
      * @return $this
      */
-    public function setTimezone(string $timezone)
+    public function setTimezone(string $timezone): self
     {
         return $this->setParam('timezone', $timezone);
     }
@@ -78,7 +82,7 @@ class Date extends AbstractProcessor
      *
      * @return $this
      */
-    public function setLocale(string $locale)
+    public function setLocale(string $locale): self
     {
         return $this->setParam('locale', $locale);
     }
--- a/lib/Elastica/Processor/DateIndexName.php
+++ b/lib/Elastica/Processor/DateIndexName.php
@@ -5,12 +5,17 @@ namespace Elastica\Processor;
 /**
  * Elastica DateIndexName Processor.
  *
- * @author   Federico Panini <fpanini@gmail.com>
+ * @author Federico Panini <fpanini@gmail.com>
  *
  * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/date-index-name-processor.html
  */
 class DateIndexName extends AbstractProcessor
 {
+    const DEFAULT_DATE_FORMATS_VALUE = ['ISO8601'];
+    const DEFAULT_INDEX_NAME_FORMAT_VALUE = 'yyyy-MM-dd';
+    const DEFAULT_TIMEZONE_VALUE = 'UTC';
+    const DEFAULT_LOCALE_VALUE = 'ENGLISH';
+
     /**
      * DateIndexName constructor.
      *
@@ -30,7 +35,7 @@ class DateIndexName extends AbstractProcessor
      *
      * @return $this
      */
-    public function setField(string $field)
+    public function setField(string $field): self
     {
         return $this->setParam('field', $field);
     }
@@ -42,7 +47,7 @@ class DateIndexName extends AbstractProcessor
      *
      * @return $this
      */
-    public function setDateRounding(string $dateRounding)
+    public function setDateRounding(string $dateRounding): self
     {
         return $this->setParam('date_rounding', $dateRounding);
     }
@@ -54,7 +59,7 @@ class DateIndexName extends AbstractProcessor
      *
      * @return $this
      */
-    public function setDateFormats(array $formats)
+    public function setDateFormats(array $formats): self
     {
         return $this->setParam('date_formats', $formats);
     }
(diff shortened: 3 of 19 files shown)


lib/Elastica/Processor/Attachment.php
```diff
@@ -44,7 +44,7 @@ public function setTargetField(string $targetField)
      *
      * @return $this
      */
-    public function setIndexedChars(int $indexedChars)
+    public function setIndexedChars(int $indexedChars = 100000): self
```
p365labs: @massimilianobraglia why did you set default value here, and not for example in ```setTargetField``` ?

massimilianobraglia: I'll check it later, but probably I just forgot that one :)

p365labs: No problem u're doing a huge *Monster* job!!!
It's really a lot! If you don't mind I'll have a review on default values on 
all classes :) maybe I can be of help

massimilianobraglia: Thank you, @p365labs!! 😄
Yep, obviously no problem from me about the review! I'm looking forward that 😄 

I just checked the code and yes, I forgot some default values! I have set those in the latest update!

lib/Elastica/Processor/Date.php
```diff
@@ -77,7 +77,7 @@ public function setTimezone(string $timezone)
      *
      * @return $this
      */
-    public function setLocale(string $locale)
+    public function setLocale(string $locale): self
```
p365labs: locale shouldn't be 'ENGLISH' as default ?

lib/Elastica/Processor/DateIndexName.php
```diff
@@ -53,7 +53,7 @@ public function setDateRounding(string $dateRounding)
      *
      * @return $this
      */
-    public function setDateFormats(array $formats)
+    public function setDateFormats(array $formats): self
```
p365labs: default value should be --> yyyy-MM-dd’T'HH:mm:ss.SSSZ

lib/Elastica/Processor/DateIndexName.php
```diff
@@ -77,7 +77,7 @@ public function setIndexNamePrefix(string $indexPrefixName)
      *
      * @return $this
      */
-    public function setIndexNameFormat(string $indexNameFormat)
+    public function setIndexNameFormat(string $indexNameFormat): self
```
p365labs: default value --> yyyy-MM-dd

lib/Elastica/Processor/DateIndexName.php
```diff
@@ -101,7 +101,7 @@ public function setTimezone(string $timezone)
      *
      * @return $this
      */
-    public function setLocale(string $locale)
+    public function setLocale(string $locale): self
```
p365labs: locale default value == 'ENGLISH'

lib/Elastica/Processor/Kv.php
```diff
@@ -67,7 +67,7 @@ public function setValueSplit(string $valueSplit)
      *
      * @return $this
      */
-    public function setTargetField(string $targetField)
+    public function setTargetField(string $targetField = '@timestamp'): self
```
p365labs: default value should be ```null```

lib/Elastica/Processor/DateIndexName.php
massimilianobraglia: Can it be directly `ISO8601`?

lib/Elastica/Processor/Kv.php
massimilianobraglia: Fixed also the phpdoc :)

massimilianobraglia:
Also here 😄

lib/Elastica/Processor/DateIndexName.php
p365labs: yep 👍

lib/Elastica/Processor/Attachment.php
ruflin: Are these the default values from Elasticsearch? One challenge when we set these is that ES might change them but Elastica does not because we miss it.

Do we need the default? Because I assume if someone calls this method, he does want to change the default?

massimilianobraglia: Yep they are the ES default values.

Anyway your's is a good point. From my PoV we have 2 ways:
- using everywhere the default values
- not using everywhere

In my travel through Elastica code, I found places where default values are used (obviously I am not talking about library default values but ES default values). My opinion is to keep them. A use could call that method just to restore the default.

Just a thought 😄

p365labs: @ruflin & @massimilianobraglia I agree with the idea of having default values in Elastica.

ruflin: The challenge I see here is: How do we keep up with defaults that are changing in ES?

From a maintainability perspective I would probably rather prefer not to have defaults in Elastica. I think the ones we have are rather old from the 0.x ages of Elasticsearch. Back then the API was not very extensive and it was still possible to keep up with changes. 

@massimilianobraglia Could you link some default examples that are already there?

massimilianobraglia: @ruflin yes no problem :)
I'll link those ASAP

massimilianobraglia: @ruflin sorry about the late answer. Anyway, I looked into the codebase of Processor namespace and I couldn't find the default values I was talkin' about. Probably those were in other namespaces (I think).

Should I remove these ones?

ruflin: @massimilianobraglia I would be for not having the defaults but lets' also see if @p365labs agrees :-)

p365labs: I understand the matter about maintainability perspective, and I agree partially. I mean maintaining defaults could be a bit annoying and without a lot of values for a library as Elastica. At the same time I'm wondering if these values could be useful not only in terms of documentation but also in order to restore default behaviours.

massimilianobraglia: Maybe we can just add class constants representing the default values, what do you think?
Something like `DEFAULT_* = value`. With these, we don't change any method signature and the user can restore the default value by using that constants

ruflin: @massimilianobraglia I like the middle ground you are proposing. This allows to also only introduce the defaults that are often needed instead of all.

massimilianobraglia: ok then! If also @p365labs agrees I will proceed 😄

p365labs: @ruflin @massimilianobraglia totally agree :) sorry for been so late in reply

massimilianobraglia: I have just updated this PR. Can you have a look, please?

massimilianobraglia:
Resolved! Let's see what Travis says :)

ruflin:
LGTM. 

@p365labs Did we want to branch out 6.x before all these PR's go in or afterwards?

p365labs:
@ruflin  that's a good point. We can take it on 6.x or not the problem I see not porting these PR's on 6.x is that if it will need to backport some PR's from master to 6.x we will have lot of code changes; otherwise if PR's will be merged before branching out 6.x it will be easier backporting PR's. How does it sound ?

ruflin:
Backporting is a good point. My biggest worry is that someone would see this change as a breaking change as some of their code might break. At the same time we just got more strict and code that was "wrong" before is now just shown. Is that statement correct?

p365labs:
mmm got the point, and I agree this approach could be better. It will be a funny challenge backporting stuff :) Anyway if we agree on this let's branch out right now 6.x and then starting merging on master these PR's.

ruflin:
I'm on the fence about these two options because I think it's also a nice thing that users have to clean up their code already in the 6.x cycle.

@massimilianobraglia Perhaps you can chime in here?

ruflin:
@thePanz If you have any feedback on this, would be great too.

ruflin:
@thePanz Thanks for the review. 

* Consts: We decided to add them but not use them, so it's not as critical to keep them up-to-date with any change on the ES side but still make it easy to use defaults.
* public const: Could you check which release this was? 
* For the php-cs-fixer: Are you looking for https://travis-ci.org/ruflin/Elastica/jobs/483331012 ? We run it now as part of each PR.
* 5.x: We plan to branch out 6.x before getting this into master. So if we event would want this into 6.x, we would need additional backports. We are not even sure yet if we should have it in 6.x as some users could see this as a breaking change.

massimilianobraglia:
Here I am, sorry about the late answer.
consts visibility have been introduced in PHP 7.1, so we cannot use it.

From my POV, if PHP 7 was introduced in Elastica in 6.x then the type hintings should be on that branch (as a new minor release, probably). Surely, this can be a BC break for someone, but if they're using the wrong types the library won't act as they expect (passing a string when the docs of a method requires explicitly an int is an error even without the type hintings).

Since we're not introducing type strictness, the BC breaks should be minimal.

I think that 6.x should be fully complain to PHP 7. In 7.x branch, PHP should be upgraded to the last officially supported PHP version with type strictness.

http://php.net/supported-versions.php

Obviously, this is just my opinion :)

ruflin:
I like the idea of a new minor release. We can do release now, then get in all the PR's and release a new minor, branch out 6.x. Thew new minor would mainly contain the PR's above and not really new features. Like this users who don't like it could stay on the old minor.

p365labs:
yes and no. I mean it depends how strict are the constraints in our users composer.json

ruflin:
@p365labs Agree, but assuming things break they have a way to fix their app by doing a more strict requirement, then fix what is broken and update the dependency afterwards.

Seems like a topic where it's hard to get to a conclusion. Should we vote?

My take so far: I'm good with getting it into a new minor release in 6.x. It has potential to break some apps but in general these apps were already broken but now we tell the users. I expect it to create some extra work on our end but overall I think our code base and the apps of the users are better of afterwards. 6.x is going to stick around for quite a long time in the future so getting users in 6.x already used to be a bit more strict should also be a good thing. It could simplify their upgrade to 7.x.

massimilianobraglia:
+1 for a new minor release with typehintings

p365labs:
Ok got it, and sounds reasonable, but doing bc's in minor release breaks a bit SemVer. 

+1 to apply typehintings on 7.x

ruflin:
😆 Seems like we are still in the same spot.

Let's discuss again the breaking change or not. I think that is the key question. Assuming some has an application `foo` and he uses some code in the wrong way. He thinks his app is working because it does not throw any error but in reality it doesn't because he uses wrong params. Now he goes from 6.1 to 6.2 and gets errors in his app. It actually tells him that his app was broken before and more or less how to fix it.

Two questions:

* Is this a breaking change?
* Is what I describe accurate?

massimilianobraglia:
First question: 
if a library change breaks my current code yes, it is a breaking change. This is the POV of the user of Elastica. This is a strict way of seeing this discussion.

I report another scenario like ours:
https://github.com/symfony/symfony/issues/29809
In one of the last Symfony updates the Form component's `TextType` was "fixed". Before it accepted any data (even an array) on it, now it accepts only strings. Which was the problem of this fix?
The problem was that when you add a field to a form with `null` as its type, Symfony silently sets the `TextType` as its type and it was not documented nor clear at all (from the user's POV again). 

Because of this, a lots of application were broken by updating that component.

In our use case, we have well documented APIs, so, as I stated in a previous comment, using erroneous  parameters and not throwing an error like an `\InvalidArgumentException` or a `\TypeError` results in a missing in our side, but it is documented (and so, it is (or it should be) clear to the user's POV). 

Also, in some classes, we have already some typehintings even without merging these PRs.

I hope I have explained better my view 😄

p365labs:
My point of view is clear, we are not leveraging an application but we're releasing a library. So if for historic reasons, bad habits, type hinting not supported, the library had bad behaviors it's not on the shoulder and responsibility of whom implement the library in its own application. We're not bug-fixing the library, we're giving a strict more approach due to the language version update. 

Continuing adding bc-breaks is totally against SemVer which at least should give the idea of more stability and a shared development model. *Patches and minor should not contain BC break*.

Again this is a never-ending discussion as this is going to be one of the new ["TwoHardThing in computer science"](https://martinfowler.com/bliki/TwoHardThings.html) ... 
sooner MF will add the new third statement :
_- have BC breaks in Minor and Patches version_

I would prefer not merging on 6.x immediately, and maybe merging it if our community really needs it and ask for TypeHintings in 6.x. So wait and see what happen instead of doing it immediately and then managing complains (right or wrong but they will come) from our users. 

these are just my 2 cents.
Again it could be a never ending story. This time I have a more conservative approach. every approach could right and every approch could be wrong. If we had a look on the net lots of things happens (thoudsands of issues :) ) on both approach.

ruflin:
Thanks everyone for his input, appreciate it. I will try to make a decision here then :-) 

* Let's branch out 6.x
* Then merge the PR's into master

As @p365labs that leaves us all the options open and only downside is potentially additional work.

@p365labs Do you want to take care of the branching of?

p365labs:
@ruflin I'll take care of it!

ruflin:
@p365labs Thank you, and after it let's get started with PR merging :-)

massimilianobraglia:
🎉

PR #1582, opened 2018-12-20
Typehinted query namespace

massimilianobraglia:
see #1548

### DIFF
--- a/lib/Elastica/Query/Common.php
+++ b/lib/Elastica/Query/Common.php
@@ -27,7 +27,7 @@ class Common extends AbstractQuery
      * @param string $query           the query string
      * @param float  $cutoffFrequency percentage in decimal form (.001 == 0.1%)
      */
-    public function __construct($field, $query, $cutoffFrequency)
+    public function __construct(string $field, string $query, float $cutoffFrequency)
     {
         $this->setField($field);
         $this->setQuery($query);
@@ -41,7 +41,7 @@ class Common extends AbstractQuery
      *
      * @return $this
      */
-    public function setField($field)
+    public function setField(string $field): self
     {
         $this->_field = $field;
 
@@ -55,7 +55,7 @@ class Common extends AbstractQuery
      *
      * @return $this
      */
-    public function setQuery($query)
+    public function setQuery(string $query): self
     {
         return $this->setQueryParam('query', $query);
     }
@@ -67,9 +67,9 @@ class Common extends AbstractQuery
      *
      * @return $this
      */
-    public function setCutoffFrequency($frequency)
+    public function setCutoffFrequency(float $frequency): self
     {
-        return $this->setQueryParam('cutoff_frequency', (float) $frequency);
+        return $this->setQueryParam('cutoff_frequency', $frequency);
     }
 
     /**
@@ -79,7 +79,7 @@ class Common extends AbstractQuery
      *
      * @return $this
      */
-    public function setLowFrequencyOperator($operator)
+    public function setLowFrequencyOperator(string $operator = self::OPERATOR_OR): self
     {
         return $this->setQueryParam('low_freq_operator', $operator);
     }
@@ -91,7 +91,7 @@ class Common extends AbstractQuery
      *
      * @return $this
      */
-    public function setHighFrequencyOperator($operator)
+    public function setHighFrequencyOperator(string $operator = self::OPERATOR_OR): self
     {
         return $this->setQueryParam('high_frequency_operator', $operator);
     }
@@ -105,7 +105,7 @@ class Common extends AbstractQuery
      *
      * @see Possible values for minimum_should_match https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-minimum-should-match.html
      */
-    public function setMinimumShouldMatch($minimum)
+    public function setMinimumShouldMatch($minimum): self
     {
         return $this->setQueryParam('minimum_should_match', $minimum);
     }
@@ -117,9 +117,9 @@ class Common extends AbstractQuery
      *
      * @return $this
      */
-    public function setBoost($boost)
+    public function setBoost(float $boost): self
     {
-        return $this->setQueryParam('boost', (float) $boost);
+        return $this->setQueryParam('boost', $boost);
     }
 
     /**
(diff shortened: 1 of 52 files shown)


massimilianobraglia:
The problem here is in the `InnerHits::toArray` method.
```php
class InnerHits
{
    [...]
    /**
     * @return array
     */
    public function toArray()
    {
        $array = parent::toArray();

        // if there are no params, it's ok, but ES will throw exception if json
        // will be like {"top_hits":[]} instead of {"top_hits":{}}
        if (empty($array['inner_hits'])) {
            $array['inner_hits'] = new \stdClass();
        }

        return $array['inner_hits'];
    }
    [...]
}
```

The problem is that I cannot set `array` as the return type (this one returns a `new \stdClass` because of the comment).

And this causes that I cannot set it to the base `ArrayableInterface`.

I hope I explained it better 😄

lib/Elastica/Query/Common.php
```diff
@@ -90,7 +90,7 @@ public function setLowFrequencyOperator($operator)
      *
      * @return $this
      */
-    public function setHighFrequencyOperator($operator)
+    public function setHighFrequencyOperator(string $operator): self
     {
         return $this->setQueryParam('high_frequency_operator', $operator);
     }
```
p365labs: - setLowFrequencyOperator
- setHighFrequencyOperator

```default value or```

File: lib/Elastica/Query/FunctionScore.php [hunk 2]
```diff
@@ -242,7 +260,7 @@ public function setMaxBoost($maxBoost)
      *
      * @return $this
      */
-    public function setBoostMode($mode)
+    public function setBoostMode(string $mode): self
```
p365labs: default value ```multiply ```

p365labs: - boost_mode
- score_mode

lib/Elastica/Query/GeoDistance.php
```diff
@@ -48,7 +48,7 @@ public function setDistance($distance)
      *
      * @return $this
      */
-    public function setDistanceType($distanceType)
+    public function setDistanceType(string $distanceType): self
```
p365labs: default value ```arc```

lib/Elastica/Query/InnerHits.php
```diff
@@ -49,9 +49,9 @@ public function setName($name)
      *
      * @return $this
      */
-    public function setSize($size)
+    public function setSize(int $size): self
```
p365labs: default vale for size

lib/Elastica/Query/Match.php
```diff
@@ -194,7 +194,7 @@ public function setFieldMaxExpansions($field, $maxExpansions)
      *
      * @return $this
      */
-    public function setFieldZeroTermsQuery($field, $zeroTermQuery = self::ZERO_TERM_NONE)
+    public function setFieldZeroTermsQuery(string $field, string $zeroTermQuery = self::ZERO_TERM_NONE): self
```
p365labs: default value ```zero_terms_query```

lib/Elastica/Query/MoreLikeThis.php
```diff
@@ -43,9 +41,9 @@ public function setLike($like)
      *
      * @return $this
      */
-    public function setBoost($boost)
+    public function setBoost(float $boost): self
```
p365labs: default value to ```1.0```

lib/Elastica/Query/MoreLikeThis.php
```diff
@@ -55,9 +53,9 @@ public function setBoost($boost)
      *
      * @return $this
      */
-    public function setMaxQueryTerms($maxQueryTerms)
+    public function setMaxQueryTerms(int $maxQueryTerms): self
```
p365labs: default value for ```max_query_terms```

lib/Elastica/Query/MoreLikeThis.php
```diff
@@ -115,9 +113,9 @@ public function setMinWordLength($minWordLength)
      *
      * @return $this
      */
-    public function setMaxWordLength($maxWordLength)
+    public function setMaxWordLength(int $maxWordLength): self
     {
-        return $this->setParam('max_word_length', (int) $maxWordLength);
+        return $this->setParam('max_word_length', $maxWordLength);
     }
```
p365labs: - min_term_freq
- min_doc_freq
- max_doc_freq
- min_word_length
etc.. 

have default values

p365labs: also 
- boost_terms

lib/Elastica/Query/Nested.php
```diff
@@ -41,7 +41,7 @@ public function setQuery(AbstractQuery $query)
      *
      * @return $this
      */
-    public function setScoreMode($scoreMode)
+    public function setScoreMode(string $scoreMode): self
```
p365labs: default to ```avg```

lib/Elastica/Query/ParentId.php
```diff
@@ -41,7 +41,7 @@ private function setId($id)
     /**
      * @param bool $ignoreUnmapped
      */
-    private function setIgnoreUnmapped($ignoreUnmapped)
+    private function setIgnoreUnmapped(bool $ignoreUnmapped)
```
p365labs: defaults to ```false```

lib/Elastica/Query/QueryString.php
```diff
@@ -71,7 +65,7 @@ public function setDefaultField($field)
      *
      * @return $this
      */
-    public function setDefaultOperator($operator)
+    public function setDefaultOperator(string $operator): self
```
p365labs: defaul ```or``` operator

lib/Elastica/Query/SimpleQueryString.php
```diff
@@ -52,7 +52,7 @@ public function setFields(array $fields)
      *
      * @return $this
      */
-    public function setDefaultOperator($operator)
+    public function setDefaultOperator(string $operator): self
```
p365labs: ```or``` is the default operator

massimilianobraglia:
@p365labs inserted the default values! Can you give it a look?

Issue #1581, opened 2018-12-20
DisMax with array of queries is broken


Destroy666x:
If you pass an array, which the function seemingly supports based on the exception/condition but not the function name, you get a `query malformed, must start with start_object` error. Moving the logic to separate `addQueries` function would be more logical

p365labs:
hi @Destroy666x 
- which version of Elastica r u using and which of Elasticsearch ?
- can you give us some example in order to have reproduce the same error ?

Destroy666x:
Latest, 6.3.x. But it'll happen with any ES version.

The example is that this https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Query/DisMax.php#L18 and this https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Query/DisMax.php#L27 line suggest that the parameter can be an array of objects extending `AbstractQuery`, but when you pass an array you simply get the mentioned error.

p365labs:
@Destroy666x got it, can u show me a simple array which raises the error ?

p365labs:
@Destroy666x #1584 PR with test proving the array works well. Probably you have tested it with a wrong array.

Destroy666x:
@p365labs Yes, you're right I had some confusion between BoolQuery and this since I used both next to eachother. My bad.

PR #1584, opened 2018-12-23
add a test for DisMax and check addQuery works with array type

p365labs:
Referencing #1581 I've added some tests in order to test addQuery behaviour.
New test added


ruflin:
@p365labs First one to fall into our new linting trap ;-) Can you rebase?

p365labs:
@ruflin it's already rebased

p365labs:
the problem seems like 
```make: *** [check-style] Error 8```

ruflin:
@p365labs Seems like you need to run the linting script then ;-)

ruflin:
@p365labs Based on your first experience with the linting requirement, should we improve the message on what needs to be done? Can we improve it?

p365labs:
yes I think we need both too 👍

- a better explanation about our linting process (also if it fails)
- update CONTRIBUTOR.md file

I'll open a PR on that.

PR #1586, opened 2019-01-04
Fix docblock for Index:analyze

thePanz:
The Analyze API call allows the `text` to be passed as an array of strings: https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-analyze.html#indices-analyze (second example)

### DIFF
--- a/lib/Elastica/Index.php
+++ b/lib/Elastica/Index.php
@@ -563,12 +563,12 @@ class Index implements SearchableInterface
     }
 
     /**
-     * Analyzes a string.
+     * Analyzes a string, or a list of strings.
      *
      * Detailed arguments can be found here in the link
      *
-     * @param string $text String to be analyzed
-     * @param array  $args OPTIONAL Additional arguments
+     * @param string[]|string $text String or list of strings to be analyzed
+     * @param array           $args OPTIONAL Additional arguments
      *
      * @return array Server response
      *
--- a/test/Elastica/IndexTest.php
+++ b/test/Elastica/IndexTest.php
@@ -893,24 +893,51 @@ class IndexTest extends BaseTest
         $this->assertEquals(0, $stats['_all']['primaries']['docs']['deleted']);
     }
 
+
+    public function analyzeDataProvider()
+    {
+        return [
+            'single-keyword' => [
+                'foo',
+                [[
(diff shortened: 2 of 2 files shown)


ruflin:
Would be great if we could first add this in master and then backport it to 5.x to not have fixes in 5.x which are not in master. Could you also add a small integration test with an array to confirm it works?

thePanz:
The code to invoke that API is different on `master`, where the parameter is already handled as an array, rather than a `string` like in v5.
I will add a test ASAP.

p365labs:
@ruflin yep on master the method accept only an array. 
The update should be applied only on 5.x branch

*5.x branch*
```php
    /**
     * Analyzes a string, or a list of strings.
     *
     * Detailed arguments can be found here in the link
     *
     * @param string[]|string $text String or list of strings to be analyzed
     * @param array           $args OPTIONAL Additional arguments
     *
     *
     * @link https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-analyze.html
     */
    public function analyze($text, $args = [])
```

*master branch*
```php
    /**
     * Analyzes a string.
     *
     * Detailed arguments can be found here in the link
     *
     * @param array $body String to be analyzed
     * @param array $args OPTIONAL Additional arguments
     *
     *
     * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-analyze.html
     */
    public function analyze(array $body, $args = [])
```

lib/Elastica/Index.php
```diff
@@ -563,12 +563,12 @@ public function requestEndpoint(AbstractEndpoint $endpoint)
...
 
     /**
-     * Analyzes a string.
+     * Analyzes a string, or a list of strings.
      *
      * Detailed arguments can be found here in the link
      *
-     * @param string $text String to be analyzed
-     * @param array  $args OPTIONAL Additional arguments
+     * @param string[]|string $text String or list of strings to be analyzed
```
p365labs: 👍 for this in 5.x branch

p365labs:
@thePanz +1 for integration test

p365labs:
@thePanz the test is fine
The error you got is a docker pull error, so nevermind, I re-started Travis and all is *green*.
So ok to merge for me!

PR #1587, opened 2019-01-06
DX: Upgrade PHP CS Fixer and it's config

keradus:
ref #1555


.php_cs.dist
```diff
@@ -10,12 +10,10 @@ $config = PhpCsFixer\Config::create()
     ->setRules([
         '@PSR2' => true,
         '@Symfony' => true,
-        'psr0' => false,
```
keradus: rule was not enabled, no need to extra disable it

.php_cs.dist
```diff
@@ -10,12 +10,10 @@ $config = PhpCsFixer\Config::create()
...
         'single_blank_line_before_namespace' => false,
         'ordered_imports' => true,
         'array_syntax' => ['syntax' => 'short'],
         'phpdoc_order' => true,
-        'blank_line_after_namespace' => true,
```
keradus: rule was enabled by enabling `@PSR2`, no need for enabling it again

.php_cs.dist
ruflin: I like to have also rules inside that we disable to be aware of it and in case a default changes we it does not change here.

keradus: do you really expect anyone to enable PSR-0 by default nowadays ?

env/elastica/Lint
```diff
@@ -33,7 +33,7 @@ ENV PATH=/root/composer/vendor/bin:$PATH
 COPY composer.json /root/composer/
 
 RUN composer global remove --no-update phpdocumentor/phpdocumentor -vvv
-RUN composer global require --no-update friendsofphp/php-cs-fixer:^2.0
+RUN composer global require --no-update friendsofphp/php-cs-fixer:^2.14
```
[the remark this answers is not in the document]
keradus: <2.12 is not maintained for looong time.
2.14 is first one supporting PHP 7.3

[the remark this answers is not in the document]
p365labs: @keradus clear that's sound reasonable :) 
and thank u for supporting PHP 7.3 we made some tricks in order to let php_cs_fixer working properly on Travis.

.php_cs.dist
ruflin: Just had to look up the current state of PSR-0 and as it's deprecated, agree that we can remove it.

p365labs:
@ruflin 👍 to merge it as with php_cs_fixer we get support for PHP 7.3

keradus:
@ruflin , `.dist` approach is the same as for `phpunit.xml` vs `phpunit.xml.dist`

`.dist` is distributed config that is shared via repository
non-`.dist` is local config (ignored by vsc) adjusted by user, if he wanted, for example he can change cache or sth just for his local machine, without pushing it to all users

ruflin:
@keradus Thanks for the details, appreciate it.

PR #1588, opened 2019-01-08
empty the raw response string after parsing

dbu:
when a query yields a large response, duplicating the data between string and parsed array duplicates the memory consumption.
once $this->_response is set, $this->_responseString is not accessed anywhere in the code anymore.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ All notable changes to this project will be documented in this file based on the
 
 ### Improvements
 
+* Reduced memory footprint of response by not keeping the raw JSON data when JSON after JSON has been parsed. [#1588](https://github.com/ruflin/Elastica/pull/1588)
+
 ### Deprecated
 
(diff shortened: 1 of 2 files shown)


p365labs:
@dbu  that's really interesting.. I've had a look at the code and _responseString is never accessed after the use in getData(). Does it really created a problem in memory consumption ? Glad that you discovered it, maybe this could save life to lot of people :)

dbu:
we have queries where the raw json is several megabytes, so yes, duplicating that is a cost.

another option would be to parse in the constructor and never even set the raw string. that would mean a performance penalty if the result is not used at all, not sure if that is likely to happen. might be for changing requests that the http status is enough to know it worked...

this change looked like the smallest impact where we can be sure there is no negative side effect in any circumstate.

p365labs:
I agree with you. I think we can proceed this way right now. @ruflin what do you think on this ?

ruflin:
👍 on the change. A changelog entry would be nice.

dbu:
added a changelog entry. ok like this?

PR #1592, opened 2019-01-13
Implement an ErrorTransport (#1529)

jandom:
This is backport of ErrorTransport onto the 5.x branch, all tests pass for me locally

Need this PR merged to close another one https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1465

Also requesting a version bump on 5.x otherwise I won't be able to configure composer.json correctly in FOSElasticaBundle

```
When using AwsAuthV4 transport, it is possible that the user provides incorrect credentials.

This means that the Guzzle client will get a 403 from AWS.

By default Guzzle silently ignores all exceptions and stores them in `$response->setTransferInfo()`

This feature allows a downstream client/library to handle them however.

Currently, FOSElastica does nothing to process 403 or other errors that come back from ruflin/Elastica (transferInfo is never used).

This causes a problem because 403 are silently ignored and the progress bar appears to be moving unhindered ;-)

This PR is a pre-requisite for unit tests that will accompany this fix on FOSElastica
# Conflicts:
#	test/Elastica/Transport/NullTransportTest.php
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,6 @@ All notable changes to this project will be documented in this file based on the
 ### Added
 
 * Added a transport class for mocking a HTTP 403 error codes, useful for testing response failures in inheriting clients
-* [Field](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html#function-random) param for `Elastica\Query\FunctionScore::addRandomScoreFunction`
 
 ### Improvements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ All notable changes to this project will be documented in this file based on the
 
 ### Added
 
+* Added a transport class for mocking a HTTP 403 error codes, useful for testing response failures in inheriting clients
+* [Field](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html#function-random) param for `Elastica\Query\FunctionScore::addRandomScoreFunction`
+
 ### Improvements
 
 - [Backported] Reduced memory footprint of response by not keeping the raw JSON data when JSON after JSON has been parsed. [#1588](https://github.com/ruflin/Elastica/pull/1588)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ All notable changes to this project will be documented in this file based on the
 ### Added
 
 * Added a transport class for mocking a HTTP 403 error codes, useful for testing response failures in inheriting clients
+[#1529](https://github.com/ruflin/Elastica/pull/1529)
+* [Backported] Added a transport class for mocking a HTTP 403 error codes, useful for testing response failures in inheriting clients
+[#1592](https://github.com/ruflin/Elastica/pull/1592)
 
 ### Improvements
 
--- a/lib/Elastica/Transport/Guzzle.php
+++ b/lib/Elastica/Transport/Guzzle.php
@@ -83,7 +83,8 @@ class Guzzle extends AbstractTransport
             throw new GuzzleException($ex, $request, new Response($ex->getMessage()));
         }
 
-        $response = new Response((string) $res->getBody(), $res->getStatusCode());
+        $responseBody = (string) $res->getBody();
+        $response = new Response($responseBody, $res->getStatusCode());
         $response->setQueryTime($end - $start);
         if ($connection->hasConfig('bigintConversion')) {
             $response->setJsonBigintConversion($connection->getConfig('bigintConversion'));
@@ -93,6 +94,7 @@ class Guzzle extends AbstractTransport
             [
                 'request_header' => $request->getMethod(),
                 'http_code' => $res->getStatusCode(),
+                'body' => $responseBody,
             ]
         );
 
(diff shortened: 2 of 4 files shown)


lib/Elastica/Transport/Guzzle.php
```diff
@@ -93,6 +94,7 @@ public function exec(Request $request, array $params)
             [
                 'request_header' => $request->getMethod(),
                 'http_code' => $res->getStatusCode(),
+                'body' => $responseBody,
```
ruflin: I wonder if this could have some side effects?

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ All notable changes to this project will be documented in this file based on the
 
 ### Added
 
+* Added a transport class for mocking a HTTP 403 error codes, useful for testing response failures in inheriting clients
+* [Field](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html#function-random) param for `Elastica\Query\FunctionScore::addRandomScoreFunction`
```
ruflin: Is this in this PR?

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ All notable changes to this project will be documented in this file based on the
 
 ### Added
 
+* Added a transport class for mocking a HTTP 403 error codes, useful for testing response failures in inheriting clients
```
ruflin: Could you add a link to the PR at the end of the line?

lib/Elastica/Transport/Guzzle.php
jandom: No idea – it's my first contribution to this package. Don't think anything in FOSElastica is trying to read this (if i remember correctly)

jandom:
This PR is currently blocking https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1465

lib/Elastica/Transport/Guzzle.php
p365labs: @ruflin it should break anything, i'm doing some tests on that and then I let u know.

ruflin: @p365labs Not sure I can follow. So this is good?

In general it's just a gut feeling I have that the above could impact other things. If we don't find something I'm good with moving forward.

jandom: Thanks for looking into it both – happy to make any changes when needed

ruflin:
@thePanz Perhaps you could also have a quick look at this PR as you were rather active on the 5.x branch recently. I just want to make sure we don't break anything here.

jandom:
@thePanz a quick look would save the day ;-)

ruflin:
@jandom Merged for now. 🤞 that it didn't cause any issues.

@thePanz Feedback still welcome ;-)

jandom:
Thanks for the merge @ruflin and for the review @thePanz 

One last quick request – can we get a new minor version published @ruflin ? Otherwise i'm unable  to close https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1465 - composer.json has to contain a stable version for maintainers to merge it in

CHANGELOG.md
jandom: In terms of chronology, first came the PR into master – then the PR into the backward compatibility branch. Happy to go either way here – this is super easy

lib/Elastica/Transport/Guzzle.php
jandom: Ha, you're right – i'm doing something completely stupid here!

The use-case was to pass the body to have useful error messages (in case the AwsAuth is passed incorrect credentials) in FOSElasticaBundle

ruflin: Does this mean we get a follow up PR?

ruflin:
I leave it to @thePanz to decide if we need a follow up / improvement PR or not. If not, @thePanz feel free to cut a release.

CHANGELOG.md
thePanz: Thus, we did not add 2 new features, just backported something. In this case I'd like to have the `[backport] ...` log message, pointing to this PR. WDYT @ruflin ?

lib/Elastica/Transport/Guzzle.php
thePanz: @jandom if I got it correctly, you're using the `body` to display the response received from the client. In here: https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1465/files/e5386eac9a4b1ee22d7d215ebe32afc5fec046af#diff-dbbc2aef2d5d1dd159ad264fb54364dcL51, why don't you keep the response data and use it in case of an exception?

thePanz: The ` $responseData = $response->getData();` can be used instead of storing twice the body contents.

thePanz:
@ruflin @jandom let's continue this discussion on https://github.com/ruflin/Elastica/pull/1595
I'd like to clean this up, before tagging a new release

ruflin:
@thePanz Ups, I think I was too fast here. Already merged #1595 :-(

ruflin:
++ on cleaning it up first.

CHANGELOG.md
ruflin: Both options work for me, whatever you prefer.

jandom: Thanks for cleaning this up gents, super-fast work!

PR #1606, opened 2019-02-08
php-cs-fixer native_function_invocation rule

massimilianobraglia:
Introduced php-cs-fixer native_function_invocation rule and applied linting

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file based on the
 ### Added
 
 ### Improvements
+* Added `native_function_invocation` CS rule [#1606](https://github.com/ruflin/Elastica/pull/1606)
 
 ### Deprecated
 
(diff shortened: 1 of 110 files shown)


p365labs:
@massimilianobraglia :
- why introducing native_function_invocation right now?
- is this process to be applied during linting? if so why no applying during a new release?

ruflin:
This change would also need a CHANGELOG entry. Are there cases where this could break things (certain PHP versions?).

massimilianobraglia:
@p365labs:
1) because we said in https://github.com/ruflin/Elastica/pull/1540 to open multiple PRs to introduce new CS rules
2) yes, it is. Unfortunately I had no time to work on this at the moment the release was created

@ruflin ok about adding a changelog entry (I'll do it right now). The case this could break things is where someone in his codebase mocked the built-in functions I think

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file based on the
 ### Added
 
 ### Improvements
+* Added `native_function_invocation` CS rule
```
ruflin: Can you add the PR number / link to it?

massimilianobraglia: Yep! Sorry about that 😄

massimilianobraglia:
> For the mocking of the built in functions I assume that would be a testing use case?

Yep, exactly

p365labs:
@massimilianobraglia & @ruflin make sense to me. 
- So let's use this on linting
- don't backport it to 6.x branch

PR #1607, opened 2019-02-08
Add Query type, exists

EliuFlorez:

PR #1608, opened 2019-02-08
Add Query type, exists - version

EliuFlorez:


p365labs:
@EliuFlorez please be a bit more declarative in PR's... give us some context in order to understand what u want to achieve.

(it looks like this PR is part of another one #1607 but not sure)

deguif:
@ruflin it seems this can be closed.
As `type` query was removed, and `exists` already exist in the `Version700` class

ruflin:
Agree, closing.

Issue #1611, opened 2019-02-22
Guzzle transport cannot connect to a second ES instance


NathanBaulch:
The static `_guzzleClientConnection` field in the `Guzzle` transport class prevents my application from creating two Elastica clients that both use Guzzle/AwsAuth4 transport but point to different ES hosts. The host of the first connection is captured in this field and used in all subsequent Elastica clients, even if they're trying to connect to a completely different ES instance. This is not a problem in the `Http` transport because the base URI is set for every request.

Some possible solutions:

1. Pass the base URI to every request instead of passing it to the constructor.
```php
$client = $this->_getGuzzleClient($connection->isPersistent());
$options = [
    'base_uri' => $this->_getBaseUrl($connection),
    'exceptions' => false, // 4xx and 5xx is expected and NOT an exceptions in this context
];
```
Does it make sense to share a connection between different ES hosts? Not sure.

2. Change the connection cache field into an array keyed by base URI.
```php
protected function _getGuzzleClient($baseUrl, $persistent = true)
{
    if (!$persistent || !isset(self::$_guzzleClientConnection[$baseUrl])) {
        self::$_guzzleClientConnection[$baseUrl] = new Client(['base_uri' => $baseUrl]);
    }
    return self::$_guzzleClientConnection[$baseUrl];
}
```

3. Change the connection cache field into a non-static field.

For now I'll have to disable `persistent` on my configured connections.

This is similar to #1523 which is also a problem with this connection cache.

ruflin:
I think we should make it as similar as possible to what we have in Http, so this would be option 1? 

Not sure if make sense to share the connection to different nodes TBH.

NathanBaulch:
I'd be happy to send a pull request for either option #1 or option #2.
For #2 I can easily apply the same base URI keyed array pattern to `Http` too, so connections aren't shared across hosts for either transport. Let me know.

ruflin:
For now I would start with mirroring what we have in Http as it seems to work well so far. We can still do a follow up then to improve both at the same time if needed.

NathanBaulch:
I've sent a PR for each approach for your consideration.

ruflin:
What a luxury to be able to pick between two PR's ❤️ What is your preferred approach? 

Short term I'm slightly in favour of https://github.com/ruflin/Elastica/pull/1618 as it only changes Guzzle. WDYT?

NathanBaulch:
[...]

PR #1619, opened 2019-03-10
No connection sharing between nodes (#1611)

NathanBaulch:


ruflin:
This PR is related to https://github.com/ruflin/Elastica/pull/1618 as both solve the same problem (this is for myself so the two PR's are linked ;-) ).

PR #1621, opened 2019-03-27
Properly handle underscores in update option names

nomoa:
Since 6.1.0 all metadata except _index, _type and _id are accepted without the
underscore. In 6.5 it produces a deprecation warning and in 7 they are no
longer accepted.

### DIFF
--- a/lib/Elastica/AbstractUpdateAction.php
+++ b/lib/Elastica/AbstractUpdateAction.php
@@ -113,7 +113,7 @@ class AbstractUpdateAction extends Param
      */
     public function setVersion($version)
     {
-        return $this->setParam('_version', (int) $version);
+        return $this->setParam('version', (int) $version);
     }
 
     /**
@@ -123,7 +123,7 @@ class AbstractUpdateAction extends Param
      */
     public function getVersion()
     {
-        return $this->getParam('_version');
+        return $this->getParam('version');
     }
 
     /**
@@ -131,7 +131,7 @@ class AbstractUpdateAction extends Param
      */
     public function hasVersion()
     {
-        return $this->hasParam('_version');
+        return $this->hasParam('version');
     }
 
     /**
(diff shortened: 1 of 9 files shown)


lib/Elastica/Bulk/Action/IndexDocument.php
```diff
@@ -39,5 +41,17 @@ protected function _getMetadata(AbstractUpdateAction $action): array
...
             'retry_on_conflict',
         ], true);
+
+        $newOptions = \array_intersect_key($options, [
+            '_index' => 1,
+            '_type' => 1,
+            '_id' => 1,
+        ]);
+
+        foreach (\array_diff_key($options, $newOptions) as $k => $v) {
```
ruflin: Couldn't we just rename the options as they should be in 7.0?

nomoa: ok, force pushed a second version of the patch. I had to make breaking changes.

lib/Elastica/AbstractUpdateAction.php
```diff
@@ -455,31 +455,16 @@ public function hasUpsert()
...
-            }
-        } else {
-            $data = $this->getParams();
-        }
-        if (!$withUnderscore) {
-            foreach ($data as $key => $value) {
-                $data[\ltrim($key, '_')] = $value;
-                unset($data[$key]);
-            }
+            return \array_filter(\array_intersect_key($this->getParams(), \array_flip($fields)));
```
ruflin: Nice cleanup.

nomoa:
Thanks for the merge, as for the initial patch I think it was still breaking 6.0 for users of retry_on_conflict with bulk updates.
We've patched our client code to hide the deprecation warnings so unless someone would like such feature in 6.x I'm going to leave this is as is.

ruflin:
@nomoa I'm pretty sure it will pop up again, but lets wait until it does ;-)

PR #1623, opened 2019-04-04
Improve error to include caused_by reason

bashilbers:
Instead of 

`[reason: failed to execute script]`

You will now get more details about the failure:

`[illegal_argument_exception:Variable [count] is not defined.] [reason: failed to execute script]`

### DIFF
--- a/lib/Elastica/Response.php
+++ b/lib/Elastica/Response.php
@@ -94,6 +94,7 @@ class Response
 
         if (isset($error['root_cause'][0])) {
             $rootError = $error['root_cause'][0];
+            $message = $rootError['reason'];
 
             if (isset($error['caused_by'])) {
                 $message .= sprintf(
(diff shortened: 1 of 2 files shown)


lib/Elastica/Response.php
```diff
@@ -90,11 +90,21 @@ public function getError()
         }
 
         $rootError = $error;
+        $message = $rootError['reason'];
```
ruflin: Do we need some checks if the fields actually exist (here and below)

ruflin (commented):
+1 on providing more details.

Could you also add a changelog entry?

bashilbers:
Done, can you check why the codestyle linter is failing for me?

ruflin:
You should be able to run `make lint` on your machine and it will fix the issue for you.

PR #1624, opened 2019-04-07
update global filter aggregation method

oguzhanT:
added filter to GlobalAggregation 

`"total": {
  "global": {},
  "aggs": {
    "all": {
      "filter": {
        "bool": {
          "must": [
            {
              "term": {
                "cityId": 1
              }
            }
          ]
        }
      }
    }
  }
}`

PR #1625, opened 2019-04-08
DGN-1497 update global aggregation filter

oguzhanT:


ruflin:
This looks like the same as https://github.com/ruflin/Elastica/pull/1624 ?

PR #1563, opened 2018-12-16
Update Elastica for Elasticsearch 7

p365labs:
This PR is to update the compatibility for Elastic to Elasticsearch 7.

The document followed to update Elastica is the one on [ES7 Breaking Changes](https://www.elastic.co/guide/en/elasticsearch/reference/master/breaking-changes-7.0.html).

PR #1628, opened 2019-04-09
[WIP] 7.x

giovannialbero1992:
I've changed docker-compose to works with 7.0.0-rc2. 
In my test there are several exceptoins and failure. 
Can it be useful start from these broken tests to develop a new release?


ruflin:
There is also work going on related to this here: https://github.com/ruflin/Elastica/pull/1563

PR #1421, opened 2017-12-12
Use Null Coalesce Operator

carusogabriel:
Following #1420

PR #1631, opened 2019-04-20
Use Null Coalesce Operator

carusogabriel:
Just a small refactoring following #1421 💅


ruflin:
@carusogabriel Thanks!

PR #1632, opened 2019-04-22
Add get document type hinting for type

livekn:
Fix getDocement() PHPDoc and add type hinting.


ruflin:
@livekn Thanks for the contribution.

Issue #1634, opened 2019-05-27
Float data type is lost in during json conversion at Action->toString()


pySilver:
Hi!

Given variable equals `(float) 1` \Elastica\Bulk\Action::toString converts it to JSON as `1` which obviously causes issues with Elasticsearch when mapping is set to double for the field.

Shouldn't we use `JSON_PRESERVE_ZERO_FRACTION` by default? 

See: https://php.net/manual/en/json.constants.php

```php
$n1 = 1.0;
$json = json_encode($n1, JSON_PRESERVE_ZERO_FRACTION); // '1.0'
$n2 = json_decode($json);

$t1 = gettype($n1); // 'double'
$t2 = gettype($n2); // 'double'

$d = $n1 === $n2; // true
```

vs

```php
$n1 = 1.0;
$json = json_encode($n1); // '1'
$n2 = json_decode($json);

$t1 = gettype($n1); // 'double'
$t2 = gettype($n2); // 'integer'

$d = $n1 === $n2; // false
```

pySilver:
Typical issue: `field mapping cannot be changed from type [float] to [long]` appears when floats are not correctly converted to json.

PR #1635, opened 2019-05-27
Fix JSON float formatting

pySilver:
See #1634 for details

### DIFF
--- a/lib/Elastica/JSON.php
+++ b/lib/Elastica/JSON.php
@@ -60,6 +60,9 @@ class JSON
         // extract arguments
         $args = \func_get_args();
 
+        // set defaults
+        isset($args[1]) ? $args[1] |= JSON_PRESERVE_ZERO_FRACTION : $args[1] = JSON_PRESERVE_ZERO_FRACTION;
+
         // run encode and output
         $string = \call_user_func_array('json_encode', $args);
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file based on the
 ### Bugfixes
 * Always set the Guzzle `base_uri` to support connecting to multiple ES hosts. [#1618](https://github.com/ruflin/Elastica/pull/1618)
 * Properly handle underscore prefixes in options and bulk request metadata ([cf upstream](https://github.com/elastic/elasticsearch/issues/26886). [#1621](https://github.com/ruflin/Elastica/pull/1621)
+* Preserve zeros while doing float serialization to JSON. [#1634] (https://github.com/ruflin/Elastica/pull/1635)
 
 ### Added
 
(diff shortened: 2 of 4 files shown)


ruflin:
Interesting addition, didn't even know this constant exists. I'm trying to think of any existing deployments we might break with this. In the case someone defined it as `integer` on the Elasticsearch side already and a float is sent, as far as I know it will just convert it back to integer and keep working. Any ideas on what it could break?

@ezimuel Any thoughts related to this?

@pySilver Could you also add a CHANGELOG entry and some tests to it?

pySilver:
@ruflin I can't see any way it can brake a valid setup here since on both sides data will be in a proper/expected type.

Btw, its a default JSON serialization flag in elasticsearch-php library for a long time already: https://github.com/elastic/elasticsearch-php/pull/481

I'll add changelog entry in a moment, but I'm unable to add tests (I've switched back from python not so long ago and not yet familiar with PHP tests yet)

ezimuel:
@ruflin, I agree to add `JSON_PRESERVE_ZERO_FRACTION` for json_encode, we used it in [elasticserach-php](https://github.com/elastic/elasticsearch-php/pull/481), as suggested by @pySilver.
I'm not sure about the fix proposed by @pySilver, since the `JSON::stringify` is also used in other Elastica classes. Instead of change all the stringify calls why don't add  `JSON_PRESERVE_ZERO_FRACTION` directly in `JSON::stringify` as default parameter?

ruflin:
Wasn't aware we also use it already in the official client. ++ on adding it then :-)

@pySilver WDYT about the proposal from @ezimuel ?

pySilver:
@ruflin I think its totally correct and should be a default behavior for all serialization tasks.

ruflin:
@pySilver Could you modify your PR to this approach?

pySilver:
@ruflin done!

ruflin:
Thanks. Could you try to run `make lint`. For some reason the linter in CI is happy. The strange thing is the file it complains about is not touched in this PR?!?

pySilver:
@ruflin yeah, same thing on localhost. The reason is here: https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Index/Settings.php#L207 introduced 4m ago at https://github.com/ruflin/Elastica/pull/1569

(Bool should be bool), changing it in my pr.

lib/Elastica/JSON.php
```diff
@@ -60,6 +60,9 @@ public static function stringify($args/* inherit from json_encode */)
         // extract arguments
         $args = \func_get_args();
 
+        // set defaults
+        isset($args[1]) ? $args[1] |= JSON_PRESERVE_ZERO_FRACTION : $args[1] = JSON_PRESERVE_ZERO_FRACTION;
```
ruflin: Had to read this line a few times but makes sense now :-)

ruflin:
@pySilver Thanks for the fix, not sure how that Bool thingy slipped through for so long.

lib/Elastica/JSON.php
pySilver: Yeah, kinda cryptic. I couldn't find easier way to append const if arg[1] is set/not set.

chx: `$args[1] = ($args[1] ?? 0) | JSON_PRESERVE_ZERO_FRACTION;`

ruflin: Thanks for chiming in @chx . Should we update the code?

chx: I think so :)

ruflin: @chx @pySilver Someone wants to take it? ;-)

PR #1637, opened 2019-06-04
Improvements on the Reindex API to support more options

damienalexandre:
This PR replace #1494 and add some missing options and tests. Thanks to @daviscaics for the initial contribution.

The Reindex API now supports all the options listed on the documentation https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-reindex.html - tests are added too for Script, remote, etc.

## BC Break

I chose to introduce some BC Breaks and I hope that's ok.

- the `run` method was also calling `refresh()` on the new Index. So it's impossible to reindex an Index and make sure it's not yet refreshed _for some reason, like doing other indexation in that index after the reindex_. It does not make sense to force a refresh on the users. If that's really an issue, I can use the REFRESH Param value with a default value to "true", to keep the current behavior, but still allow to disable it;
- the `run` method was returning the new Index object, that's not consistent with other API and that does not allow to read the response content (is there any conflict, how long does it took...). So I changed the returned value.

## Still not supported

- the ability to set multiple index as the source index

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ All notable changes to this project will be documented in this file based on the
 * The parameter [fields](https://www.elastic.co/guide/en/elasticsearch/reference/master/breaking-changes-7.0.html#_the_parameter_literal_fields_literal_deprecated_in_6_x_has_been_removed_from_bulk_request) deprecated in 6.x has been removed from Bulk requestedit and Update request.
 * The [_parent](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-parent-field.html) field has been removed in favour of the join field.
 * hits.total is now an object in the search response [hits.total](https://www.elastic.co/guide/en/elasticsearch/reference/master/breaking-changes-7.0.html#_literal_hits_total_literal_is_now_an_object_in_the_search_response) 
+* Elastica\Reindex does not return an Index anymore but a Response.
+* Elastica\Reindex->run() does not refresh the new Index after completion anymore. Use `$reindex->setParam(Reindex::REFRESH, 'wait_for')` instead.
 
 ### Bugfixes
 * Always set the Guzzle `base_uri` to support connecting to multiple ES hosts. [#1618](https://github.com/ruflin/Elastica/pull/1618)
@@ -21,6 +23,7 @@ All notable changes to this project will be documented in this file based on the
 ### Added
 
 * Added `ParentAggregation` [#1616](https://github.com/ruflin/Elastica/pull/1616)
+* Elastica\Reindex missing options (script, remote, wait_for_completion, scroll...)
 
 ### Improvements
 * Added `native_function_invocation` CS rule [#1606](https://github.com/ruflin/Elastica/pull/1606)
@@ -28,7 +31,7 @@ All notable changes to this project will be documented in this file based on the
 
 ### Deprecated
 
-## [Unreleased](https://github.com/ruflin/Elastica/compare/6.1.0...6.1.1)
+## [6.1.1](https://github.com/ruflin/Elastica/compare/6.1.0...6.1.1)
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -21,14 +23,15 @@ All notable changes to this project will be documented in this file based on the
...
+* Elastica\Reindex missing options (script, remote, wait_for_completion, scroll...)
 
 ### Improvements
 * Added `native_function_invocation` CS rule [#1606](https://github.com/ruflin/Elastica/pull/1606)
 * Elasticsearch test version changed from 6.5.2 to 6.6.1 [#1620](https://github.com/ruflin/Elastica/pull/1620)
 
 ### Deprecated
 
-## [Unreleased](https://github.com/ruflin/Elastica/compare/6.1.0...6.1.1)
+## [6.1.1](https://github.com/ruflin/Elastica/compare/6.1.0...6.1.1)
```
ruflin: Thanks for the fix.

lib/Elastica/Reindex.php
```diff
@@ -32,84 +46,94 @@ class Reindex
...
+    private function _resolveBodyOptions(array $params)
     {
-        return \array_intersect_key($options, [
+        return \array_intersect_key($params, [
             self::SIZE => null,
             self::CONFLICTS => null,
         ]);
     }
 
-    /**
```
ruflin: Why did you remove these comments?

lib/Elastica/Reindex.php
```diff
@@ -32,84 +46,94 @@ class Reindex
...
-        $body = $this->_getBody($this->_oldIndex, $this->_newIndex, $this->_options);
+        $body = $this->_getBody($this->_oldIndex, $this->_newIndex, $this->getParams());
 
         $reindexEndpoint = new \Elasticsearch\Endpoints\Reindex();
+        $params = \array_intersect_key($this->getParams(), \array_fill_keys($reindexEndpoint->getParamWhitelist(), null));
+        $reindexEndpoint->setParams($params);
         $reindexEndpoint->setBody($body);
 
-        $this->_oldIndex->getClient()->requestEndpoint($reindexEndpoint);
-        $this->_newIndex->refresh();
```
ruflin: Good idea to remove it as I agree we should not enforce the refresh.

lib/Elastica/Reindex.php
```diff
@@ -32,84 +46,94 @@ class Reindex
...
+        $this->_lastResponse = $this->_oldIndex->getClient()->requestEndpoint($reindexEndpoint);
 
-        return $this->_newIndex;
+        return $this->_lastResponse;
```
ruflin: +1 on returning the response. The _newIndex will still be accessible through the object but the response would be lost. Should we even store it in the object?

damienalexandre: Storing the Response is only useful for `getTaskId()`, which is not that important to me as it's easy to get now (thanks to the response being returned). I will remove this method.

lib/Elastica/Reindex.php
damienalexandre: To make the code more homogeneous and less noisy - lots of methods are self-documented via type-hint, this PHP DocBlock don't provide any useful information and is on a private method.

Should I put it back?

lib/Elastica/Reindex.php
ruflin: I missed that you use it for getTaskId. I must confess there could be some convenience for users that know less about reindex to have the getTaskId method available.

lib/Elastica/Reindex.php
ruflin: How will doc blocks generated by phpdoc look like without the docs? Will an IDE know the type that is returned? If the behaviour with or without stays the same, happy to remove it.

damienalexandre: Those are private methods so I don't think it changes anything for the end user.

Also behavior is really good on IDE like phpStorm too. I'm even going one step further by adding return type hint as Elastica is 7.0 :+1:

damienalexandre:
I think we can agree to keep the `_lastReponse` property, as it's convenient for the end-user.

Let me know if there is anything else :pray: :+1:

ruflin:
@damienalexandre Thanks for pushing this over the line.

Issue #1441, opened 2018-01-17
Please add the possibility to change the request method


stefpe:
https://github.com/ruflin/Elastica/blob/02bf2fbb8ad6b679869af145eed89436a1c50229/lib/Elastica/Search.php#L466

Currently the request method is hardcoded and I need to change it to be able to use a proxy. My current workaround is the following:
```php
        $search = $this->indexType->createSearch($query);
//request via client, otherwise the path string generation will be prefixed again.
        $response = $this->indexType->getIndex()->getClient()->request(
            $search->getPath(),
            Request::POST,
            $search->getQuery()->toArray(),
            $options
        );
        $result = $search->getResultSetBuilder()->buildResultSet($response, $search->getQuery());
```

But what I actually wanted to do would be something like this: 
```php
$result = $this->indexType->search($query, $options, Request::POST);
```

ruflin:
Can you share some details on why GET requests do not work with proxies? Because if that is the case, there are many other requests where this becomes an issue I would think.

stefpe:
Because in the HTTP Transport, you are setting 
`curl_setopt($conn, CURLOPT_POSTFIELDS, $content);`
For elasticsearch somehow a GET Request with filled post body works, but a proxy which doesnt read the POST Body on a GET Request, you run into issues.

ruflin:
Thanks for the details. I'm ok with adding this option. Can you open a PR?

In case the proxy is under your control, I think the proxy should be configured accordingly. But I can see the issue in case the proxy is not under your control.

I wonder what the proxy is that does not support it?

ruflin:
@stefpe I just stumbled over this one: https://github.com/elastic/elasticsearch/pull/28342 It seems the official high level client moved to POST, so I wonder if we should make this even the default?

alamirault:
PR is now merge, I think @ruflin you can close issue.

PR #1643, opened 2019-06-30
Add request method parameter for search and count

alamirault:
This PR allow to change request method for search and count as discussed in #1441.

### BC BREAK

I also moved default request method from `GET` to `POST` as suggested by @ruflin in [his comment](https://github.com/ruflin/Elastica/issues/1441#issuecomment-363285144).


ruflin:
@alamirault Thanks for taking this on and sorry for the late reply, was on paternity leave ;-) I restarted CI to see if it was just a temporary failure or not. I assume your commit it based on current master?

alamirault:
Congratulation ! Yes is based on current master.
Some tests failed because POST is set by default and they check dynamic changes. Should I delete them?

ruflin:
Thanks :-) Should the tests be deleted or just adapted? Can you share a link to the culprits?

alamirault:
https://github.com/ruflin/Elastica/blob/bbe10bc2d16bed6041d98add9d66c01294a7f92f/test/Elastica/Transport/GuzzleTest.php#L46

I think this test must be deleted. Or I should rework my PR in order to keep GET by default, change to POST only if search have data. And the `$method` arg (nullable by default) allow to override this behavior. I think is less predictive.

ruflin:
If POST is the expected default, I think we should fully switch to POST and not have as you said, a less predictive behaviour. So far I can't see any downsides?

+1 on deleting the test.

alamirault:
Code is rebased and I deleted the test.

PR #1650, opened 2019-08-14
update to ES 7.3

p365labs:
update Elasticsearch version to ES 7.3

### DIFF
--- a/env/elasticsearch/Dockerfile
+++ b/env/elasticsearch/Dockerfile
@@ -1,5 +1,5 @@
 #https://www.docker.elastic.co/
-FROM docker.elastic.co/elasticsearch/elasticsearch:7.2.0
+FROM docker.elastic.co/elasticsearch/elasticsearch:7.3.0
 MAINTAINER Nicolas Ruflin <spam@ruflin.com>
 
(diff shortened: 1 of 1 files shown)


ruflin:
Thanks. Now that I'm back we should really get Elastica 7.0 out ;-)

p365labs:
@ruflin yes I think we need to release a 7.x version

ruflin:
@p365labs What is your view on where we are stading to ship a 7.x (or a beta first?)

p365labs:
@ruflin I think we can release the new version skipping the beta version:
people who need/wanted to work on 7.x branch used dev-master since June 2019.
Bugs or problems should have been already notified by our user base.

The only doubts I have is that, maybe, we don't have implemented all the features of 7.x 
I will do a review to all the **BC changes/new features** in the Elastic documentation and I will come back to you tomorrow. After that we can have a clear view of what we have on master :)

ruflin:
SGTM :-) I think it's ok if we don't have all the features of 7.x in Elastica, we can always add them if needed and there is also the underlying official php client which can be used for non existing features. 

The part I'm more worried about: Are we done with the breaking change? This will be our last chance for quite a few months :-)

Issue #1640, opened 2019-06-14
Allow to pass DSN to the Elastica\Client constructor


renanbr:
Currently I have to handle config options as separated keys in each environment my application is deployed (local, ci, staging, prod...). Things become hard when an environment required an option the other don't...

It could be cool if  `Elastica\Client`'s constructor becomes able to parse a single string.

```php
new Elastica\Client('192.168.1.1:9200');
```

```php
new Elastica\Client('https://user:p4ss@foo.com:9200/elastic;persistent=false;retryOnConflict=2');
```

tacman:
Yes, this would be particular helpful on deployments like Heroku, which creates the DSN environment variable when the service is created, but configuring it requires manually breaking it apart into the individual components.

ruflin:
Do you see a way how we could introduce this without making it a breaking change? In general, +1 on having such a feature.

alamirault:
PR is now merge, I think @ruflin you can close issue.

PR #1652, opened 2019-08-17
Support DSN in Client constructor for config argument (#1640)

alamirault:
This PR allow to pass DSN to the Elastica\Client constructor as discussed in #1640 .

For example: `$client = new Client('https://user:p4ss@foo.com:9200?persistent=false&retryOnConflict=2');`

I have just a doubt about `url` config parameter. Should I build it from `transport`, `host`, `port` and `path` ? Or it works when these args are filled ?

Another question: Is there a doc to update ? Elastica.io appears to be outdated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ All notable changes to this project will be documented in this file based on the
 * Elastica\Reindex does not return an Index anymore but a Response.
 * Elastica\Reindex->run() does not refresh the new Index after completion anymore. Use `$reindex->setParam(Reindex::REFRESH, 'wait_for')` instead.
 * `Elastica\Search->search()` and `Elastica\Search->count()` use request method `POST` by default. Same for `Elastica\Index`, `Elastica\Type\AbstractType`, `Elastica\Type`.
+* Elastica\Client `$_config` field is now a `ClientConfiguration` instead of an array
 
 ### Bugfixes
 * Always set the Guzzle `base_uri` to support connecting to multiple ES hosts. [#1618](https://github.com/ruflin/Elastica/pull/1618)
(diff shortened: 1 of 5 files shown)


lib/Elastica/Dsn.php
```diff
@@ -0,0 +1,222 @@
...
+        $this->connections = $connections;
+        $this->roundRobin = $roundRobin;
+        $this->log = $log;
+        $this->retryOnConflict = $retryOnConflict;
+        $this->bigintConversion = $bigintConversion;
+        $this->username = $username;
+        $this->password = $password;
+    }
+
+    public static function fromString(string $dsn): self
```
ruflin: As all params are kind of hardcoded, how do we keep it up-to-date on changes?

alamirault: What changes ? If you talk about config array maybe we have to move Dsn object to a Configuration object which have methods `fromDsn` and `fromArray` and use it in the Client.

alamirault:
I reworked my PR. Now the client configuration is exported in new class. It avoid code duplication.
I also allow any extra option (removed hardcoded part)

ruflin:
I like the change to use a separate class for the config object. I'm trying to poke holes into this change to see what side effects it could have. So far the only thing I could think of is that something that inherits the client could mess around with `_config` but even this should be fine as it can still be an array.

It seems everything should continue to works as before?

alamirault:
I think is only side effect. Indeed everything works without any changes.

I also would like deprecate `setConfig`,`getConfig`,`setConfigValue`,`getConfigValue` and add a ClientConfiguration getter in Client. WDYT ?

ruflin:
@alamirault Your proposal SGTM. I already merged the PR as I think these additional changes / deprecations should go into a separate PR. I hope you are ok with that?

alamirault:
I agree. I will make another PR when I have time.

Issue #1392, opened 2017-10-18
Missing support for Field Collapsing?


StefanGla:
Hello,
I wonder if the elastica package already supports field collapsing?

https://www.elastic.co/guide/en/elasticsearch/reference/5.6/search-request-collapse.html#search-request-collapse

If not, how can I add these collapse-parameters to my query?

ruflin:
@StefanGla You should be able to use `setRawQuery` to set the params: http://elastica.io/api/latest/classes/Elastica.Query.html#method_setRawQuery But it would be pretty nice to add support for it with functions.

ruflin:
Field collapsing is supported — `Elastica\Collapse` plus `Query::setCollapse()`:

```php
$collapse = new \Elastica\Collapse();
$collapse->setFieldname('user_id');

$query = new \Elastica\Query(new \Elastica\Query\MatchAll());
$query->setCollapse($collapse);
```

`Elastica\Collapse` additionally supports `setInnerHits()` / `addInnerHits()` (taking an `Elastica\Collapse\InnerHits`) and `setMaxConcurrentGroupSearches()`. It is also reachable through the query builder as `Elastica\QueryBuilder\DSL\Collapse`.

Closing as implemented. If a specific collapse option is still missing, please open a new issue naming it.

PR #1653, opened 2019-08-20
Field Collapsing Support

wackerl91:
Added support for Field Collapsing as requested in #1392.
ES documentation: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html#request-body-search-collapse

Note: Inheriting `Query\InnerHits` in `Collapse\InnerHits` was something I've been debating with myself over and over again. As far as I'm concerned it makes sense to have different classes, especially as the collapse one has to have support for the second level collapse, but I didn't want to duplicate the entire code for InnerHits.
Downside is that right now there's no `AbstractCollapse` base-class as there is for the other concepts. Still feel like that's okay, because field collapsing is a quite simple concept right now.

### DIFF
--- a/lib/Elastica/Query.php
+++ b/lib/Elastica/Query.php
@@ -58,7 +58,7 @@ class Query extends Param
      *
      * @return self
      */
-    public static function create($query): self
+    public static function create($query)
     {
         switch (true) {
             case $query instanceof self:
@@ -91,7 +91,7 @@ class Query extends Param
      *
      * @return $this
      */
-    public function setRawQuery(array $query): self
+    public function setRawQuery(array $query)
     {
         $this->_params = $query;
 
@@ -105,7 +105,7 @@ class Query extends Param
      *
      * @return $this
      */
-    public function setQuery(AbstractQuery $query): self
+    public function setQuery(AbstractQuery $query)
     {
         return $this->setParam('query', $query);
     }
@@ -127,7 +127,7 @@ class Query extends Param
      *
      * @return $this
      */
-    public function setFrom($from): self
+    public function setFrom($from)
     {
         return $this->setParam('from', $from);
     }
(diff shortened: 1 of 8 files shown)


lib/Elastica/Query.php
```diff
@@ -100,15 +105,15 @@ public function setRawQuery(array $query)
...
+    public function setQuery(AbstractQuery $query): self
     {
         return $this->setParam('query', $query);
     }
 
     /**
      * Gets the query object.
      *
-     * @return \Elastica\Query\AbstractQuery
+     * @return array|\Elastica\Query\AbstractQuery
```
ruflin: Where is this change coming from?

lib/Elastica/Query.php
```diff
@@ -86,7 +91,7 @@ public static function create($query)
      *
      * @return $this
      */
-    public function setRawQuery(array $query)
+    public function setRawQuery(array $query): self
```
ruflin: Is this needed or some cleanup?

lib/Elastica/Query.php
wackerl91: Both `Query::create` and `Query::__construct` allow passing in an array as argument, which don't parse / convert this array into an ensemble of objects but keep it as-is, essentially leveraging the `setRawQuery` method.

Calling `getQuery` after taking the route above (or by just setting a raw query on the Query object) will return an array. I think that's something one should be aware of.

FWIW, I do like the option to set the raw query, however I'd rather have it in a way where it's not mingled with the usual query abstraction but has it's own representation to have better control over it during it's lifecycle.

lib/Elastica/Query.php
wackerl91: Cleanup. I'll remove them later today and re-add them in a new PR as you requested in your comment above.

wackerl91:
> I guess we can still change it later without breaking the API if we figure out the other path works better?

Yes, that should work. Thing is that this might lead to a lot of duplicated code between both versions of InnerHits, on the flip side offering better control if they ever start to drift apart in ES DSL. I'd personally sit that out and wait how ES continues to handle field collapsing in the future, as this decision likely also depends on how complex collapsing becomes.

PR #1654, opened 2019-08-22
Do not force `include_type_name`

thePanz:
Do not arbitrary alter the `requestEndpoint` for index creation or when pushing mappings, the `include_type_name` should be configured on the endpoint, not added by the client


ruflin:
@thePanz The `include_type_name` param will still exist in 7.0. You mention it should be configured by the endpoint, can you elaborate on how this would look like after this PR?

thePanz:
@ruflin I expect these two different ways to create an index:

The one with type name included;
```php
$create = new Create();
$create->setBody(['mappings' => ['type' => ['properties' => ['field1' => ['type' => 'text']]]]]);
$create->setParams(['include_type_name' => true]);
$index->requestEndpoint($create);
```

And without deprecation:

```php
$create = new Create();
$create->setBody(['mappings' => ['properties' => ['field1' => ['type' => 'text']]]]);
$index->requestEndpoint($create);
```

WDYT?

Looks like the the current `index->create()` shortcut does not allow to pass any create parameters (thus the `include_type_name`).

thePanz:
Tests are failing, as this would require more work: we currently push the mappings with the `type` (we even require the type to be set in `Mappings::toArray()`, where we throw an exception if the type is not set).

ruflin:
I like what you proposed above. It reduces the magic which is always good. 

That we require the type in Mapping sounds now more like a bug. Any chance you could tackle this or is this a rabbit hole?

thePanz:
@ruflin I will try to spend some time into this, no guarantees tho :)

thePanz:
Closing, as those changes are done in #1666

PR #1656, opened 2019-08-25
Update dev-master alias to 7.0.x-dev

predakanga:
Update in line with the "7.x" mentioned in README.md


ruflin:
LGTM. TBH I think I need to read up on `branch-alias` if this is the optimal value we want to set. I guess also the minor should be variable?

PR #1659, opened 2019-09-02
Add callable type hinting to Connection callback

franmomu:
The `$callback` parameter is only used in `ConnectionPool` which already has the type hint.


ruflin:
Code LGTM to me. Could you add a changelog entry just in case someone was using this in an unexpected way?

franmomu:
Sure!

ruflin:
@franmomu Thanks for the contribution.

PR #1660, opened 2019-09-05
Clear scroll context also when empty page was received

mjanser:


mjanser:
Added a changelog entry.

The scroll context was already cleared when a `foreach` loop over the scroll was finished (`currentPage >= $totalPages`). This change clears the context also when the page contains an empty result. This can happen if there aren't any hits and therefore the first page contains no results. This directly ends the `foreach` loop and the scroll context was kept open.

ruflin:
Thanks for the additional details. I must confess I have not used this part of the code enough myself to fully understand the potentially side effects but it looks good. So let's get it in, especially as it doesn't break any tests :-)

Thanks for the contribution.

PR #1662, opened 2019-09-19
Make classes with factory methods extensible

csizmazia:
# What this is changing
Use `static` instead of `self` in all factory methods in Elastica.

# What is my motivation/problem
I am working on a ES persistence layer library for my company.
My goal is to write it in a way that users can combine various query implementations with various ES clients (i.e. elasticsearch-php and elastica at the moment), Elastica queries being among the options. For this i wrote a thin wrapper for the `Elastica\Query` class and struggled with the factory method since it was using `self` instead of `static`. This is violating the open-closed principle in my opinion, hence i created this PR :)

### DIFF
--- a/lib/Elastica/Query.php
+++ b/lib/Elastica/Query.php
@@ -64,21 +64,16 @@ class Query extends Param
             case $query instanceof self:
                 return $query;
             case $query instanceof AbstractQuery:
+            case \is_array($query):
+            case $query instanceof Suggest:
+            case $query instanceof Collapse:
                 return new static($query);
             case empty($query):
                 return new static(new MatchAll());
-            case \is_array($query):
-                return new static($query);
             case \is_string($query):
                 return new static(new QueryString($query));
             case $query instanceof AbstractSuggest:
                 return new static(new Suggest($query));
-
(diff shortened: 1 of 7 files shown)


lib/Elastica/Query.php
```diff
@@ -64,21 +64,16 @@ public static function create($query)
...
-            case $query instanceof AbstractSuggest:
-                return new self(new Suggest($query));
-
             case $query instanceof Suggest:
-                return new self($query);
-
             case $query instanceof Collapse:
-                return new self($query);
+                return new static($query);
+            case empty($query):
```
ruflin: Is there a reason you reordered these statements?

csizmazia: to avoid code duplication. but you're right, this changes semantics as will. i will revert that change.

csizmazia:
Well, there are potential side effects for code that is already extending any of the changed classes. Such code would potentially behave differently obviously.

ruflin:
@csizmazia Merged it for now. As it went into master, it's also ok if some users would consider this breaking. Overall I would expect devs extending the lib to benefit from this. Thanks for the contribution.

Issue #1663, opened 2019-09-20
Support for search option track_total_hits


skylord123:
This should be super easy to add support for. I was able to patch it locally by just changing a couple lines. 

Here is the page on elasticsearch for this:
https://www.elastic.co/guide/en/elasticsearch/reference/7.0/search-request-track-total-hits.html

Thanks!

ruflin:
Closing as https://github.com/ruflin/Elastica/pull/1664 was merged.

PR #1664, opened 2019-09-22
Add method setTrackTotalHits to Elastica\Query (#1663)

alamirault:
This PR add support of track_total_hits https://github.com/ruflin/Elastica/issues/1663


ruflin:
@alamirault Thanks for the contribution. Already merged as I don't see a reason it should not work. Still would be interesting to have some functional tests to confirm it works?

alamirault:
I agree. I'll do it.

Issue #1651, opened 2019-08-15
Cache AWS CredentialProvider


L3tum:
Hi,

I've used this library a tad bit on AWS and noticed that it requests the credential provider every time it makes a request. 

``` php
        if ($connection->hasParam('aws_secret_access_key')) {
            return CredentialProvider::fromCredentials(new Credentials(
                $connection->getParam('aws_access_key_id'),
                $connection->getParam('aws_secret_access_key'),
                $connection->hasParam('aws_session_token')
                    ? $connection->getParam('aws_session_token')
                    : null
            ));
        }

        return CredentialProvider::defaultProvider();
```

One way to improve this would be by caching the credentialprovider, like this

``` php
$doctrineCacheAdapter = new DoctrineCacheAdapter(new ApcuCache());
$cachedCredentialsProvider = CredentialProvider::cache(CredentialProvider::defaultProvider(), $doctrineCacheAdapter);
```

Would you be willing to do something like this?

As background, we ran into some bottlenecks while load testing on AWS and noticed that the connection often times and the only difference between our local servers and AWS is the request signing portion.

ruflin:
Would this introduce any new dependency?

L3tum:
I think it can work without extra dependencies and cache providers by using the aws-sdk memoize function [here](https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/guide_credentials_provider.html)(scroll down to the end of the page). Could even be done with an extra Transport like CachedAwsAuthV4 to make it opt-in. 

Optionally there could be another Transport that accepts some kind of PSR/Cache or so object to cache the provider with, but that'd be some neat extra that isn't really necessary IMO.

ruflin:
Overall the plan SGTM. One thing I'm generally concerned about all the AWS magic: We can't test it the way we test all the other things. So if we change things, I don't know if it will work :-) I only know it works know, because users are using it :-D

L3tum:
Yeah agreed, testing AWS stuff in general can be a PITA. 

I would propose a separate Transport like `CachedAwsAuthV4` so it'd be opt-in and mark it as experimental.

We sadly don't have a testing environment for this right now but I'd be willing to test it as soon as we got it set up.

ruflin:
[...]

PR #1667, opened 2019-10-04
Support an aws_credential_provider connection parameter for more flexible credentials strategies

jeskew:
Intended to address #1651. This PR updates the AWSAuthV4 transport to support an additional configuration parameter (`aws_credential_provider`) to allow users to supply a callable credential provider. This would let users customize their provider to include caching, fetch credentials from a non-standard location, etc.


jeskew:
@ruflin Sorry for the radio silence on this PR! I added a changelog entry.

PR #1668, opened 2019-10-08
[ES-7] Make Client class Type free

renanbr:
Attempt to:

- Cleanup Client methods requiring Type
- Fix ClientTest
- Copy methods from Type to Index:
    - `addDocument()`
    - `createDocument()`

Test is green `make test TEST="test/Elastica/ClientTest.php"`:

- `ClientTest::testEndpointParamsRequest()` checks against "index_total" instead of "type"


test/Elastica/ClientTest.php
```diff
@@ -349,86 +308,13 @@ public function testUpdateDocuments()
...
-        $idxString = $index->getName();
-        $typeString = $type->getName();
-        $this->assertInternalType('string', $idxString);
-        $this->assertInternalType('string', $typeString);
+        $docData1 = $index->getDocument(1)->getData();
+        $docData2 = $index->getDocument(2)->getData();
 
-        // Try to delete doc with a routing value which hashes to
-        // a different shard then the id.
-        $resp = $index->getClient()->deleteIds($ids, $index, $type, 'second_routing');
```
thePanz: Looks like this test was checking the "routing" too, I guess we should keep it

ruflin:
Would  be good to get a changelog for this PR.

thePanz:
@ruflin we have `Removed `\Elastica\Type` class, `\Elastica\Index` class must be used instead`, it should cover those changes. wdyt?

ruflin:
SGTM. I'm happy with having an "overall" changelog entry in the end.

PR #1671, opened 2019-10-11
Always set Guzzle base_uri (#1611) (#1618)

### LINKED ISSUE #1644: Reuse of guzzle client in Elastica 6.x
Hello everyone,

When issuing a Search and and Msearch in the same script execution, the Guzzle client is reused for better performance. The problème is that the Content-Type of the second request is ignored, because the Content-Type is stored within the Guzzle Client

Here is the problème
https://github.com/ruflin/Elastica/blob/6.x/lib/Elastica/Transport/Guzzle.php#L159-L171

We should maybe backport the fix from the master branch
https://github.com/ruflin/Elastica/pull/1618/files

reedy:
Fixes #1644

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file based on the
 ### Backward Compatibility Breaks
 
 ### Bugfixes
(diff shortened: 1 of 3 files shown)


ruflin:
@reedy Thanks for the backport

PR #1673, opened 2019-10-11
Require ext-json in composer

thePanz:


ruflin:
Could you share some background on why we need this? Was it missing?

Quick note on branches: I would prefer if you could branches in your fork and open PR's from there so. Unfortunately I can't do that too as it's in my repo :-(

thePanz:
@ruflin the requirement is sort of "implicit" from the `elasticsearch/elasticaserach` dependency.

This PR makes it explicit, as we use `json_decode()` functions, too.

composer.json
```diff
@@ -13,6 +13,7 @@
     ],
     "require": {
         "php": "^7.1",
+        "ext-json": "*",
```
ruflin: Could we use a more specific version here?

thePanz: PHP extensions have no version, either they're enabled or not :)

PR #1674, opened 2019-10-11
[ES-7] Fix tests

renanbr:
Here are some tests fixes.

I pushed individual commits to make easier you cheery-pick those you like, juste in case.

and, a question: Is there a reason to keep `Version240` class? Strangely, it has some queries that aren't avaiable in 2.4, such `parent_id`


thePanz:
Thank you @renanbr , I will merge your changes shortly :)
Regarding the Version240 class: we should remove it in a separated PR

Issue #1677, opened 2019-10-17
Aggregation: Ability to add metadata


leflings:
Hi

First of all, thank you for a very nice library, which I've enjoyed using (with/through [FOSElasticaBundle](https://github.com/FriendsOfSymfony/FOSElasticaBundle))

ElasticSearch has the ability to add _[metadata](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/agg-metadata.html)_ to aggregations, a feature I've found to be wanting while making a faceted search solution.

Is it possible to emulate this with the current API (I'm using version 6.1.1)?

If not, could this be added to the Aggregation API? Hopefully in a way where it lands in a 6.X version (the bundle doesn't yet support versions 7.X)

ruflin:
Just learned a new feature in ES, thanks! My guess would be that `setParam` could be  used on each aggregation to add it. I didn't try it out but as all Aggregations inherit `Params`, this should work: https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Aggregation/AbstractAggregation.php#L9 

Happy to get native support for it, interested to contribute 😇

leflings:
I tried `setParam` but it unfortunately doesn't generate a valid aggregation.

The following code:
```
$termsAgg = new \Elastica\Aggregation\Terms('titles');
$termsAgg->setField('title');
$termsAgg->setParam('meta', ['color' => 'blue']);

$query = new \Elastica\Query();
$query->addAggregation($termsAgg);
```

Generates the following query:
```
{
    "aggs": {
        "titles": {
            "terms": {
                "field": "title",
                "meta": {
                    "color": "blue"
                }
            }
        }
    },
    "query": {
        "match_all": {}
    }
}
```

Which generates the following response:
```
{
  "error": {
    "root_cause": [
      {
        "type": "x_content_parse_exception",
        "reason": "[6:17] [terms] unknown field [meta], parser not found"
      }
    ],
    "type": "x_content_parse_exception",
    "reason": "[6:17] [terms] unknown field [meta], parser not found"
  },
  "status": 400
}
```

The `meta` object needs to be output "one level higher" / at the same level as the aggregation type. See [Structuring Aggregations](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/search-aggregations.html#_structuring_aggregations).

[...]

PR #1678, opened 2019-10-18
Allow metadata to be set on AbstractAggregation (#1677)

leflings:
ElasticSearch allows metadata to be set on an aggregation that will be
returned as-is with the aggregation result.

See https://www.elastic.co/guide/en/elasticsearch/reference/6.8/agg-metadata.html

Using `setParam` generates invalid query, hence these additions. See
https://github.com/ruflin/Elastica/issues/1677

### DIFF
--- a/lib/Elastica/Aggregation/AbstractAggregation.php
+++ b/lib/Elastica/Aggregation/AbstractAggregation.php
@@ -19,6 +19,11 @@ abstract class AbstractAggregation extends Param implements NameableInterface
     protected $_aggs = [];
 
     /**
+     * @var array|null Metadata belonging to this aggregation
+     */
+    protected $_meta;
+
+    /**
      * @param string $name the name of this aggregation
      */
     public function __construct($name)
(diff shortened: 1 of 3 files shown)


lib/Elastica/Aggregation/AbstractAggregation.php
```diff
@@ -18,6 +18,11 @@ abstract class AbstractAggregation extends Param implements NameableInterface
      */
     protected $_aggs = [];
 
+    /**
+     * @var array|null Metadata belonging to this aggregation
+     */
+    protected $_meta;
```
ruflin: Do we need a separate `_meta` key here or could we just use the internal `setParam`?

leflings: After going over the code in `AbstractAggregation::toArray()` I agree the problem could be solved by using `AbstractAggregation::_setRawParam($key, $value)`.

leflings:
Closing this and opening #1680 instead. Branched from `master` and using `_rawParams`

PR #1679, opened 2019-10-18
Cleanup Query documentation

thePanz:
Cleanup Query documentation for ES-7
- [x] add `): self` to fluent methods
- [x] remove un-needed `@return $this` annotations
- [x] Updated changelog for breaking changes

### DIFF
--- a/lib/Elastica/Query.php
+++ b/lib/Elastica/Query.php
@@ -23,16 +23,14 @@ use Elastica\Suggest\AbstractSuggest;
 class Query extends Param
 {
     /**
-     * Suggest query or not.
-     *
-     * @var int Suggest
+     * If the current query has a suggest in it.
      */
-    protected $_suggest = 0;
+    private $hasSuggest = false;
 
     /**
      * Creates a query object.
      *
-     * @param array|\Elastica\Query\AbstractQuery $query OPTIONAL Query object (default = null)
+     * @param array|AbstractQuery $query Query object (default = null)
      */
     public function __construct($query = null)
     {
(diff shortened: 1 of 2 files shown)


lib/Elastica/Query.php
```diff
@@ -23,16 +23,16 @@
...
+     * @var bool Suggest
      */
-    protected $_suggest = 0;
+    protected $hasSuggest = false;
 
     /**
      * Creates a query object.
      *
-     * @param array|\Elastica\Query\AbstractQuery $query OPTIONAL Query object (default = null)
+     * @param array|AbstractQuery $query Query object (default = null)
```
ruflin: I had in the past the "Optional" inside to make it more obvious that there is a default when not looking at the code. But it should be obvious enough?

lib/Elastica/Query.php
```diff
@@ -194,51 +171,45 @@ public function setHighlight(array $highlightArgs)
...
         return $this->addParam('highlight', $highlight);
     }
 
     /**
      * Sets maximum number of results for this query.
      *
-     * @param int $size OPTIONAL Maximal number of results for query (default = 10)
+     * @param int $size Maximal number of results for query (default = 10)
      *
-     * @return $this
```
ruflin: Why do you consider the `return $this` as not needed anymore?

lib/Elastica/Query.php
```diff
@@ -23,16 +23,16 @@
...
 {
     /**
-     * Suggest query or not.
+     * If the current query has a suggest in it.
      *
-     * @var int Suggest
+     * @var bool Suggest
      */
-    protected $_suggest = 0;
+    protected $hasSuggest = false;
```
ruflin: Is it time to get rid of the protected `_` convention everywhere? The simple answer is yes, but how do we migrate there? Do we just adjust the new code to be different or all at once?

lib/Elastica/Query.php
thePanz: Yes, it is obvious as the `$var = null` signature :)

lib/Elastica/Query.php
thePanz: The method signature is `method(): self`, the annotation is no superfluous.
I am planning to take all of those into consideration in a next PR (php-cs-fixer can help here ;-) )

lib/Elastica/Query.php
thePanz: Yes, I'd move all of those to proper camelCase and get rid of the `_` prefix for protected properties in a next PR

ruflin: I hear breaking change? Should we try to do it before 7.0 release?

ruflin:
Seems CI is not happy :-(

lib/Elastica/Query.php
thePanz: Done, changelog updated too

ruflin:
Thanks!

PR #1682, opened 2019-10-23
Restore request data in debug logging

neild3r:
After the merge of #1661 I've not been getting the request body in my debug logs. The issue seems to be due to the the fact that it's not being converted into an array like it used to be.

I've not added to the CHANGELOG.md as this is fixing an issue that is already in that change log.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file based on the
 ### Bugfixes
 
 * The `\Elastica\Query\GeoPolygon::count()` method now returns the count of points passed to the filter [#1696](https://github.com/ruflin/Elastica/pull/1696)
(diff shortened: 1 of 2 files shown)


ruflin:
I assume this went missing in the conversion and change LGTM. But it would be nice if @Tobion could comment if it was intentional.

neild3r:
Any thoughts @Tobion?

ruflin:
@neild3r As it seems @Tobion is currently not around, lets move this forward. As we did a release in the meantime, could you add a changelog?

thePanz:
@neild3r I agree with `as this is fixing an issue that is already in that change log.` ,but in this case this is a bugfix :)

neild3r:
@thePanz I only suggested this as when I made the original PR there hadn't been a release since the original change that caused the problem. 

@ruflin I've added the to the CHANGELOG.md so this hopefully is now ready to go

ruflin:
@neild3r @thePanz Thanks for getting this in.

PR #1686, opened 2019-10-31
Update Elasticsearch test version to 7.14.1

ruflin:


ruflin:
Forcemerge test fails, probably a change in ES we should look into.

thePanz:
Do you have any update about this? the same test fails on ES 7.5
Could it be a breaking change introduced in the newest ES releases?

ruflin:
@thePanz Didn't get a chance to investigate this. I assume it is related to a breaking change, we should check the changelogs.

p365labs:
*IndexTest::testForceMerge* fails the last assertion:

-         $this->assertEquals(0, $stats['_all']['primaries']['docs']['deleted']);

dumping the value of *$stats['_all']['primaries']['docs']['deleted']* with:

- ES 7.3 
```
array(2) {
  ["count"]=>
  int(1)
  ["deleted"]=>
  int(0)
}
```

- ES 7.4.1
```
array(2) {
  ["count"]=>
  int(1)
  ["deleted"]=>
  int(2)
}
```

- ES 7.5.x
```
array(2) {
  ["count"]=>
  int(1)
  ["deleted"]=>
  int(2)
}
```

looks like a bc or something in the delete endpoint.

.github/workflows/continuous-integration.yaml
```diff
@@ -36,11 +36,11 @@ jobs:
           - php: '7.2'
             elasticsearch: '7.2.1'
           - php: '7.3'
-            elasticsearch: '7.3.2'
+            elasticsearch: '7.10.2'
```
ruflin, line 39: @deguif Can you remind me how the older ES versions are related to older PHP versions?

tests/IndexTest.php
```diff
@@ -798,7 +798,7 @@ public function testForcemerge(): void
 
         $stats = $index->getStats()->getData();
         $this->assertEquals(1, $stats['_all']['primaries']['docs']['count']);
-        $this->assertEquals(0, $stats['_all']['primaries']['docs']['deleted']);
+        $this->assertEquals(2, $stats['_all']['primaries']['docs']['deleted']);
```
ruflin, line 801: I'm not entirely sure why before it was 0 and not 1, and now I'm not really sure why it is 2 and not still 1 😕  Anyone knows more here?

ruflin (commented):
I updated this PR to go to 7.10.2 directly. There is one additional small change in the extend ExtendedStatsBucket with the output that needed changes.

The part that confuses we around the delete is that I think before and after it should have been 1. So both 0 and 2 don't seem to be correct.

The diffs in tests mean it is tricky to run it across multiple versions of Elasticsearch. What I would suggest is that we only test against 7.10.2 and remove the older versions and in case there is no simple answer for the delete 2, I would still get the PR in as is to not block progress on other PRs.

@deguif @thePanz @p365labs Would be great if you could chime in.

thePanz:
@ruflin Ok for me (but the docs still mention that we test against 7.4.1 :) )

I would clarify that this library "tries" to have running tests against the latest releases of ES, as being able to cover and run the whole set of tests on all minor releases would be quite time consuming. WDYT?

tests/IndexTest.php
deguif: After a forcemerge deleted document count should probably be `0` I think as it frees up space retaining by deleted document?

.github/workflows/continuous-integration.yaml
deguif: What do you mean about `related to older PHP versions`?

tests/IndexTest.php
ruflin: I'm browsing through the force merge docs and there are quite a few mentions of `deletes`: https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-forcemerge.html But so far could not tie it to this issue we have here.

.github/workflows/continuous-integration.yaml
ruflin: Why do we test PHP 7.2 with ES 7.1.1 and ES 7.2.1 but PHP 7.3 only with ES 7.3.2?

ruflin:
@thePanz  I made an update on the readme, can you check if this is what you were looking for?

tests/IndexTest.php
```diff
@@ -795,10 +795,11 @@ public function testForcemerge(): void
         $this->assertEquals(1, $stats['_all']['primaries']['docs']['count']);
 
         $index->forcemerge(['max_num_segments' => 1]);
+        $index->refresh();
```
ruflin, line 798: I tried a refresh to maybe get a different result, no success.

README.md
```diff
@@ -17,8 +17,7 @@ All library issues should go to the [issue tracker from github](https://github.c
 
 Compatibility
 -------------
-This release is compatible with all Elasticsearch 7.0 releases and onwards.
-It was tested with version 7.3.0.
+This release is compatible with all Elasticsearch 7.0 releases and onwards. The testsuite is run against the most recent minor version of the Elastic Stack, currently 7.10.2.
```
thePanz, line 20: IIRC we only test it against the latest version of "Elasticsearch", not Elastic Stack :)
Please keep the two sentences on two different lines, will help to keep the diff short

.github/workflows/continuous-integration.yaml
thePanz: IIRC php 7.2 is the lowest PHP version supported by this library.
Maybe we could test it against:
- the lowest PHP version we support
- the latest version available of PHP 7.4 (and 8.0, if we support PHP 8.x)

WDYT?

deguif: @ruflin I think it was done to test several ES versions but I'm not sure, but each ES version is not tied to a PHP one, just the matrix was done this way.

We already test it with PHP 8.0 but elasticsearch official library has not yet published a stable version supporting PHP 8.0.

ruflin: I definitively like the idea of testing multiple PHP versions and I think we should keep that. The part I'm challenging is aorund how many ES versions we should test in parallel. My suggestion would be to test against the most recent version released by Elastic against the available / supported PHP version. We might add one more test against the most recent ES version which we things is fully compatible and does not require special testing code, so that would be currently 7.4 I think but might soon be 7.10 so not sure it is worth it.

The alternative is to test the first and last minor of a major, like 7.0.x and 7.10.2 today to make sure we stay compatible, but I think this creates overhead on the testing side and we should rely on Elastic to keep things mostly backward compatible during minor releases.

tests/IndexTest.php
deguif: Just make some tests to investigate this problem. I will propose a PR to fix and test with the current version.

.github/workflows/continuous-integration.yaml
ruflin: I suggest to do this in a follow up PR.

tests/IndexTest.php
deguif: @ruflin I will push the fix, it's about using `only_expunge_deletes`.

ruflin: Nice, thanks!

deguif: Done

ruflin:
@deguif It seems something I broke on the linting side but not sure why. I get

```
Files that were not fixed due to errors reported during linting before fixing:
   1) /Users/ruflin/Dev/Elastica/src/Query/Match.php
   2) /Users/ruflin/Dev/Elastica/tests/IndexTest.php
   3) /Users/ruflin/Dev/Elastica/tests/Multi/SearchTest.php
```

ruflin:
@deguif Tests are no green but not because I fixed the tests but skipped the broken ones :-( Having the skip in should also allow us to discuss each a bit on what the issue should be and fix it. If you know how to fix some, please don't hold back with commits ;-)

tests/Multi/SearchTest.php
```diff
@@ -292,6 +292,7 @@ public function testSearchWithError(): void
 
         $multiSearch->addSearch($searchBad);
 
+        $this->markTestSkipped('Elastica\Exception\ResponseException: [size] parameter cannot be negative, found [-2]');
```
ruflin, line 295: In this test it seems we cause an error on purpose by passing in -2 (same below). No ES does proper validation and returns an error already here. Easiest change would be to just adjust the test and check for the error here or even remove the test.

ruflin:
@deguif I currently don't really find the time to investigate the failures further at the same time I think it is worth to have this PR in to unblock others. It means we would "reenable / fix" the tests later. Thoughts?

deguif:
@ruflin I'm OK with that. I don't have free time to work on theses issues on my side too.

PR #1687, opened 2019-10-31
Update changelog: mention move of Mapping class

thePanz:


ruflin:
How is it possible that this broke a client tests?!?

ruflin:
Thanks, I don't like flaky tests :-)

Issue #1691, opened 2019-11-12
Bump QueryBuilder version


deguif:
I see that the latest query builder version is `Version240` and the type filter was removed in it.
Shouldn't we instead bump it to a `Version700` or something like that?

ruflin:
@deguif That should work I think. Want to take care of it?

deguif:
@ruflin it's done with many other things ;)
Do you think you will have time to review/merge soon?

ruflin:
@deguif Thank you for all your contributions. Will try to go through all of them (some I already did).

PR #1693, opened 2019-11-14
Bump query builder version to 7.0

deguif:
Follow-up of #1691


deguif:
That's because there was a new version for PHP-CS-FIXER ;)

deguif:
@ruflin once #1695 gets merged, I will rebase and the tests should be green

ruflin:
Hm, interesting. I would hope such changes would not break our CI. What we should probably do moving forward is have a fixed version of the php-cs-fixer linked here instead of `^` .. https://github.com/ruflin/Elastica/blob/master/env/elastica/Lint#L36 ?

deguif:
@ruflin rebased, and it's green now.

ruflin:
Thanks, LGTM. Merged.

PR #1695, opened 2019-11-14
Fix travis (coding style)

deguif:


Makefile
```diff
@@ -86,7 +86,7 @@ lint:
 .PHONY: check-style
 check-style:
 	docker build -t ruflin/elastica-dev-base-linter -f env/elastica/Lint env/elastica/
-	${RUN_ENV_LINT} php-cs-fixer fix --allow-risky=yes --dry-run
+	${RUN_ENV_LINT} php-cs-fixer fix --allow-risky=yes --dry-run --diff
```
deguif: Add the `--diff` option to better see why it fails in Travis.

ruflin:
@deguif Is this an automated clean up done by the linting? Do we need any changes to the linting config?

deguif:
The linting changed because of a new PHP-CS-FIXER version and new rules for @Symfony conventions ;)

deguif:
I think that can be merged unless we don't want to follow @Symfony conventions

lib/Elastica/Index/Settings.php
```diff
@@ -281,10 +259,7 @@ public function getRefreshInterval(): string
      *
      * To have this changes made the index has to be closed and reopened
      *
-     * @param string $key   Merge policy key (for ex. expunge_deletes_allowed)
```
ruflin: I like that it keeps the ones with actual documentation.

ruflin:
Thanks @deguif for the clean up. We now probably have to rebase all PR's.

@thePanz I think we should have a fixed version of php-cs-fixer in our CI build so these kind of things don't happen in the future: https://github.com/ruflin/Elastica/blob/master/env/elastica/Lint#L36 A CI build should be reproducible. WDYT?

thePanz:
Yes, I totally agree @ruflin 
There are 2 ways to accomplish it (IMHO):
- add php-cs-fixer as a dev dependency to composer.json
- use Phive and have it pulled in with a fixed version/release

lib/Elastica/Exception/Connection/GuzzleException.php
```diff
@@ -21,9 +21,7 @@ class GuzzleException extends ConnectionException
     protected $_guzzleException;
 
     /**
-     * @param TransferException $guzzleException
-     * @param Request           $request
-     * @param Response|null     $response
+     * @param Request $request
```
thePanz: Not sure why only `$request` is kept here :confused:

lib/Elastica/Multi/Search.php
```diff
@@ -78,8 +74,7 @@ public function clearSearches(): self
     }
 
     /**
-     * @param BaseSearch $search
-     * @param string     $key
+     * @param string $key
```
thePanz: This is not correct IMHO, as it should be `string|null`

lib/Elastica/Util.php
```diff
@@ -198,8 +198,7 @@ public static function convertDate($date)
      *
      * Converts it to the lucene format, including the appropriate TimeZone
      *
-     * @param \DateTime $dateTime
-     * @param bool      $includeTimezone
+     * @param bool $includeTimezone
```
thePanz: This should be added to the method signature, `bool $includeTimezone = true`

thePanz:
@deguif could you address the couple of comments I left? just minor changes :)

ruflin:
@thePanz Sorry, already merged. @deguif Any chance you could do a follow up PR with the above proposed changes?

Does adding php-cs-fixer to composer have any disadvantages? I assume there is a reason we don't have it here?

deguif:
@ruflin I will do the follow up PR ;)

lib/Elastica/Multi/Search.php
deguif: maybe we should add `?string` in the type-hint of the method and remove the php-doc for the param?
Even better we could also use the `nullable_type_declaration_for_default_null_value` option for PHP-CS-FIXER to enforce the rule, WDYT?

thePanz: In this case I believe it is an error on the docblock: we accept null, but the annotation not.
The annotation is more strict than the method signature.

I'd rather (1) remove the annotation (if we handle null values), or (2) check the implementation if null values are not accepted. If (2) the method signature is wrong, and then the annotation is superfluous :)

lib/Elastica/Exception/Connection/GuzzleException.php
deguif: For the same reason probably I explained below

lib/Elastica/Multi/Search.php
deguif: https://github.com/php/php-src/pull/3535

thePanz: Didn't know about `nullable_type_declaration_for_default_null_value` and the discussion. LGTM to use it :+1: 

Anyway: is the method really accepting null values?

deguif: Null values seem to be accepted, so I'm going to remove the annotation, as argument type-hint is sufficient

deguif:
@ruflin @thePanz minor changes asked here are now addressed in #1706

PR #1696, opened 2019-11-15
GeoPolygon - Fix wrong property passed to \count()

deguif:


ruflin:
Change makes sense to me. Could you add a changelog entry? And if possible, add a functional test? We should have had this in the first place to detect this problem :-(

deguif:
@ruflin I added test and changelog entry.

ruflin:
@deguif Thank you!

PR #1702, opened 2019-11-18
Update query builder query dsl

deguif:


deguif:
I just added the `CHANGELOG` entry.

deguif:
@ruflin do you think this can be merged?
I'm relying on the query builder a lot in my code base, so I would really love to get this merged.

ruflin:
@deguif yes

Interesting to hear that you are heavily relying on the query builder. It is one of these features I thought we should potentially remove but sounds like there are users out there.

Issue #1701, opened 2019-11-16
Bump PHP minimum version to 7.2


deguif:
As PHP 7.1 is going EOL very soon, should we bump its constraint in composer file to 7.2 for the upcoming version 7.0 of the library and adjust Travis configuration?

ruflin:
@deguif SGTM. But I would still keep the 7.0 / 7.1 tests around so we know when we actually really going to break things.

philkra:
I would suggest to rather make a strong cut to min `7.2` as php version or to stick around with `7.0`, as the new language constructs introduced in `7.1` and `7.2` are not compatible with `7.0`. Running tests with a `7.2` code will break in case in `7.0` and likely in `7.1`.

ruflin:
See my comment in https://github.com/ruflin/Elastica/pull/1703

PR #1703, opened 2019-11-18
Drop support for PHP 7.1

deguif:
Follow up of #1701


.travis.yml
```diff
@@ -16,7 +16,6 @@ env:
     - DOCKER_VERSION=17.09.0
     - DOCKER_COMPOSE_VERSION=1.17.1
   matrix:
-    - TARGET="71"
```
ruflin: Would this all work but keep this target + the Docker file in? Or will all the tests fail because of the composer requirement?

deguif: I think this won't work because of the composer requirement, at least if `--ignore-platform-reqs` is not used

ruflin:
A few general questions:

* How widely is 7.1 still used? 
* Is it easy for users to upgrade to 7.2 or later?
* What are the things you expect we introduce in Elastica which will not be compatible with 7.1?

I'm good with moving forward with this PR. In the past what I did was that what we test and what is "supported" was not necessarily the same. This meant as soon as we introduced a feature that actually broke a previous PHP version we knew and then removed the testing. Like this it was easy to communicate what broke it.

deguif:
@ruflin I'm not an expert on the subject but here some stats from composer: https://blog.packagist.com/php-versions-stats-2019-1-edition/

Upgrading to PHP 7.2 should be pretty straightforward.
And some big packages on composer already switched to PHP 7.2 as minimum requirement, see @symfony 5 for example.

I think it could be judicious to really remove PHP 7.1 support as you are about to publish a new major version. This also brings opportunity to boost PHP 7.2 adoption if users of packages need this new version, which is good for the PHP community.

I see two PHP 7.2 features that could be used: `object` type-hint  and basic contravariance.

ruflin:
@deguif Thanks for sharing the additional details. I especially like the point around boosting adoption of 7.2.

PR #1705, opened 2019-11-18
Unskip test

deguif:
Try to unskip test as it seems it was resolved. https://github.com/elastic/elasticsearch/issues/21312


ruflin:
Seems something is still up:

```
There was 1 failure:
1) Elastica\Test\Query\InnerHitsTest::testInnerHitsLimitedSource
Failed asserting that an array has the key 'name'.
/elastica/test/Elastica/Query/InnerHitsTest.php:370

```

deguif:
Ok, let's close it then. I dumped the query result and this is not working.

PR #1706, opened 2019-11-18
Fix CS bis

deguif:
See https://github.com/ruflin/Elastica/pull/1695#discussion_r347510004 for the new rule

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -19,6 +19,7 @@ $config = PhpCsFixer\Config::create()
         '@PHPUnit60Migration:risky' => true,
         'php_unit_dedicate_assert' => ['target' => 'newest'],
         'native_function_invocation' => true,
+        'nullable_type_declaration_for_default_null_value' => true,
     ])
 ;
 
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -19,6 +19,7 @@ $config = PhpCsFixer\Config::create()
         '@PHPUnit60Migration:risky' => true,
         'php_unit_dedicate_assert' => ['target' => 'newest'],
         'native_function_invocation' => true,
+        'no_alias_functions' => true,
         'nullable_type_declaration_for_default_null_value' => true,
     ])
 ;
--- a/lib/Elastica/Aggregation/Derivative.php
+++ b/lib/Elastica/Aggregation/Derivative.php
@@ -25,7 +25,7 @@ class Derivative extends AbstractAggregation
      *
      * @return $this
      */
-    public function setBucketsPath(string $bucketsPath)
+    public function setBucketsPath(string $bucketsPath): self
     {
         return $this->setParam('buckets_path', $bucketsPath);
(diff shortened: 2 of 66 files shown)


deguif:
Ok, let's close it then. I dumped the query result and this is not working.

lib/Elastica/Aggregation/Derivative.php
```diff
@@ -35,7 +35,7 @@ public function setBucketsPath(string $bucketsPath)
      *
      * @return $this
```
thePanz: The `return $this` should be removed too :)

lib/Elastica/Transport/HttpAdapter.php
```diff
@@ -32,7 +32,7 @@ class HttpAdapter extends AbstractTransport
      *
      * @param Connection $connection
      */
-    public function __construct(Connection $connection = null, HttpAdapterInterface $httpAdapter)
+    public function __construct(?Connection $connection = null, HttpAdapterInterface $httpAdapter)
```
deguif: There is a glitch here, optional argument before a non optional argument.

ruflin:
@deguif Any chance you could update the PR description to contain a bit more details on the change? I'm aware the comment is linked but if someone comes back to this and asking why this all changed, would be nice to have it in a single place summarised.

lib/Elastica/Aggregation/Derivative.php
ruflin: Lets separate PR's which are fully automated and manual changes ;-)

lib/Elastica/Transport/HttpAdapter.php
ruflin: If we don't make it optional, would things break?

.php_cs.dist
```diff
@@ -19,6 +19,8 @@ $config = PhpCsFixer\Config::create()
         '@PHPUnit60Migration:risky' => true,
         'php_unit_dedicate_assert' => ['target' => 'newest'],
         'native_function_invocation' => true,
+        'no_alias_functions' => true,
```
thePanz: What about adding `@PhpCsFixer` and `@PHP71Migration` too?

ruflin: No clue what the effect will be but I suggest we do this in a follow up PR to keep PRs small.

ruflin: MR = PR? So you are ok to merge and then follow up? :-)

thePanz:
@ruflin Ok to merge! I'd squash the changes, just to keep the git history clean. wdyt?

ruflin:
@thePanz Yes, I always squash. Commit message is now not optimal because I took just the PR one but well ... @deguif Could you still update the PR description?

deguif:
@ruflin @thePanz Yes, will do that

deguif:
@ruflin @thePanz description should now better reflect the changes introduced here.

PR #1708, opened 2019-11-19
Remove string check and use type-hinting

deguif:


ruflin:
Test is probably failing because now we get a different exceptions. In general code LGTM.

lib/Elastica/Aggregation/AdjacencyMatrix.php
```diff
@@ -10,16 +10,10 @@ class AdjacencyMatrix extends AbstractAggregation
     /**
      * Add a named filter.
      *
-     * @param string $name
-     *
      * @return $this
      */
-    public function addFilter(AbstractQuery $filter, $name): self
+    public function addFilter(AbstractQuery $filter, string $name): self
```
deguif: Hum interesting, now with the type hint, the arguments are casted to string, as we don't use `declare(strict_types=1);` it doesn't throw an exception. Only here as there was no type-hint before.

I'm not sure if the test is still relevant?

ruflin: Removing the test for me is probably ok. Need to see the code but should be fine.

test/Elastica/Aggregation/AdjacencyMatrixTest.php
```diff
@@ -68,21 +68,6 @@ public function testToArray()
         $this->assertEquals($expected, $agg->toArray());
     }
 
-    /**
-     * @group unit
-     */
-    public function testNonStringNamedFilters()
```
deguif: Here's the test in question ;)

lib/Elastica/Aggregation/AdjacencyMatrix.php
thePanz: I agree, when moving to ES-7 I also removed tests that did not make sense anymore when we had typehints

test/Elastica/Aggregation/AdjacencyMatrixTest.php
thePanz: yea, just remove the test. PHP is taking care of passing a string to the method and throwing exceptions otherwise

deguif:
It's green ;)

PR #1711, opened 2019-11-19
Fix elasticsearch/elasticsearch conflicting versions

deguif:


deguif:
@ruflin @thePanz I changed to avoid using `conflict` but rather use version constraints.

deguif:
Any change to get this merged now in order to fix tests.
I would do another PR once a solution is found.

thePanz:
@deguif merged! Thanks!

Issue #1710, opened 2019-11-19
Tests broken since release of elastic/elasticsearch-php 7.4.0


deguif:
I investigated the issue and it seems some classes are now auto-generated, and their name was changed.

For example, `Elasticsearch\Endpoints\Ingest\Pipeline\Delete` was renamed to `Elasticsearch\Endpoints\Ingest\DeletePipeline`

deguif:
@ruflin @thePanz a quick solution would be to update composer.json to avoid installing this version.

But I'm not sure what will be the best way to support 7.4 version and all 7.x versions.
Maybe we should create an issue on their repository.

deguif:
https://github.com/ruflin/Elastica/pull/1711 should fix the tests, by marking the `elasticsearch` package as conflicting since version `7.4.0` in the `composer.json` file.

thePanz:
A better solution would be to set the version range to something like `^7.0 || < 7.4.0`
So that we could work on the issue with more time. wdyt?

deguif:
@thePanz I'm not sure the `||` will do what is expected here.
As `7.4.0` should match `^7.0`, but we should probably use `^7.0 < 7.4.0` as it would be an `and` in this case.
I was thinking the `conflict` was more appropriate, but it's up to you ;)

deguif:
I confirm `^7.0 || <7.4.0` is not working while `^7.0 <7.4.0` is working as expected.

deguif:
I also created an issue on the official `elastichsearch-php` repository, let's see what will come out.

ezimuel:
I just commented about this issue [here](https://github.com/elastic/elasticsearch-php/issues/967).

ruflin:
For having consitent testing, +1 on fixing the dependency version (already done). I would hope somewhere in the future, releases of the two libraries are even in more sync.

At the same time I'm kind of glad we found this now os testing with "latest" is not only bad. I suggest we make the change on our end before we ship 7.0 GA. As far as I understand it mainly changes some internal implementation but nothing external?

ezimuel:
@ruflin exactly, I'm doing my best to avoid BC breaks but sometime it's quite hard to keep all previous inconsistency in place. I'm working on `elasticsearch-php 7.4.1` to reintroduce (some) old namespaces as class aliases.
Now that we have a code generation from the REST API specification we'll be more aligned with Elasticsearch version.

ruflin:
[...]

PR #1715, opened 2019-11-25
Fix BC break in elasticsearch-php by excluding broken version

deguif:
Related to #1710 and elastic/elasticsearch-php#967
Should be merged once elastic/elasticsearch-php#968 is merged and version 7.4.1 of elastic/elasticsearch-php is released.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
         "php": "^7.2",
         "ext-json": "*",
         "psr/log": "~1.0",
-        "elasticsearch/elasticsearch": "^7.0 <7.4.0"
+        "elasticsearch/elasticsearch": "^7.0 !=7.4.0"
     },
     "require-dev": {
         "aws/aws-sdk-php": "~3.0",


ruflin:
I'm good with this change for now. But I think we should come up with a strategy on how we release Elastica and elasticsearch-php together or close together to align on the version numbers. So if you have Elastica 7.4, you will have elasticsearch-php 7.4 or similar.

philkra:
FYI, elasticsearch-php follows the stack release up to the minors. Usually, the release of the client is attempted on the same day or very close.

ezimuel:
I just released [7.4.1](https://github.com/elastic/elasticsearch-php/releases/tag/v7.4.1) that solves the BC break related to the endpoint class naming and it solves also the `trigger_error()` issue https://github.com/babenkoivan/scout-elasticsearch-driver/issues/297 using the suppress error `@` operator.

PR #1718, opened 2019-11-30
Update PHP-CS-FIXER composer version constraint

deguif:


env/elastica/Lint
```diff
@@ -33,7 +33,7 @@ ENV PATH=/root/composer/vendor/bin:$PATH
 COPY composer.json /root/composer/
 
 RUN composer global remove --no-update phpdocumentor/phpdocumentor -vvv
-RUN composer global require --no-update friendsofphp/php-cs-fixer:^2.14
+RUN composer global require --no-update friendsofphp/php-cs-fixer:^2.16
```
deguif: The version constraint can be changed to `~2.16.0` to avoid code style kreaking test on new update, but not sure if we should do this as this would require to update manually for newer minor version.

ruflin: +1 on having `~2.16.0` instead. I rather update it manually from time to time then to break CI. Also I "hope" the main reason we normally update this is because something in the formatting changed, so we need to do additional changes anways.

deguif: Done

ruflin: Thanks!

PR #1719, opened 2019-11-30
Add type-hints on ResultSet class

deguif:


lib/Elastica/ResultSet.php
```diff
@@ -354,7 +324,7 @@ public function offsetSet($offset, $value)
      *
      * @param int $offset
      */
-    public function offsetUnset($offset)
```
ruflin: Doc was not even correct 🤦‍♂

PR #1725, opened 2019-12-04
Changed default test target to PHP 7.2

wackerl91:
Just a minor change: following up on dropping support for PHP 7.1, the default test target for local tests (using the `make tests` target) should be set to 7.2.


ruflin:
Thanks for the change. LGTM. Somehow CI on 7.3 complained but restarted it. lets see.

deguif:
@ruflin that's not the first time this test is failing ;)

PR #1726, opened 2019-12-05
Remove duplicated "timeout" comment.

Brah0um:
I think the `timeout` term is duplicated. I just removed one form comment.


ruflin:
@Brah0um Thanks for the fix! Merged.

PR #1733, opened 2019-12-11
Update PHP Docs

thePanz:


lib/Elastica/Index.php
```diff
@@ -682,24 +642,19 @@ public function setSettings(array $data)
...
      * @param string       $path   Path to call
      * @param string       $method Rest method to use (GET, POST, DELETE, PUT)
-     * @param array|string $data   OPTIONAL Arguments as array or encoded string
-     * @param array        $query  OPTIONAL Query params
-     *
-     * @return Response Response object
+     * @param array|string $data   Arguments as array or encoded string
      */
-    public function request($path, $method, $data = [], array $query = [])
+    public function request($path, string $method, $data = [], array $queryParameters = []): Response
```
ruflin: I guess the first param is also a string?

lib/Elastica/Search.php
```diff
@@ -40,48 +40,41 @@ class Search
...
-    protected $_indices = [];
-
-    /**
-     * Array of types.
-     *
-     * @var array
+     * @var string[]
      */
-    protected $_types = [];
+    protected $indices = [];
```
ruflin: Be aware this part of the renaming could be seen as a breaking change as it is protected.

lib/Elastica/Search.php
```diff
@@ -163,73 +144,53 @@ public function setOptions(array $options)
...
     /**
-     * @param string $key
-     *
-     * @throws InvalidException
-     *
-     * @return bool
+     * @throws InvalidException If the given key is not a valid option
      */
-    protected function _validateOption($key)
+    private function validateOption(string $key): void
```
ruflin: change to private on purpose? I think it's fine just want to make sure it is on purpose?

lib/Elastica/SearchableInterface.php
```diff
@@ -26,30 +26,25 @@ interface SearchableInterface
      * }
      *
      * @param string|array|\Elastica\Query $query   Array with all query data inside or a Elastica\Query object
-     * @param null                         $options
-     * @param string                       $method  OPTIONAL Request method (use const's) (default = Request::POST)
-     *
-     * @return \Elastica\ResultSet with all results inside
+     * @param arra\                        $options
```
ruflin: arra\ ?

lib/Elastica/Index.php
thePanz: Fixed!

lib/Elastica/Search.php
thePanz: as v7.0 is not yet "released" this could be a good time for it :)

lib/Elastica/Search.php
thePanz: Moved back to "protected". I don't see why we would allow to extend from this class, tho

thePanz:
@ruflin yes, there were too many changes here. I reverted the `_` changes and the method's visibility.

lib/Elastica/SearchableInterface.php
```diff
@@ -26,30 +26,25 @@ interface SearchableInterface
...
+     * @param string                       $method Request method, see Request's constants
      *
      * @return int number of documents matching the query
      */
-    public function count($query = '', $method = Request::POST);
+    public function count($query = '', string $method = Request::POST);
 
     /**
      * @param \Elastica\Query|string $query
-     * @param array                  $options
```
deguif: int|array|null instead maybe as it is the same than for `search`?

lib/Elastica/SearchableInterface.php
```diff
@@ -26,30 +26,25 @@ interface SearchableInterface
      * }
      *
      * @param string|array|\Elastica\Query $query   Array with all query data inside or a Elastica\Query object
-     * @param null                         $options
-     * @param string                       $method  OPTIONAL Request method (use const's) (default = Request::POST)
-     *
-     * @return \Elastica\ResultSet with all results inside
+     * @param int|array                    $options Limit or associative array of options (option=>value)
```
deguif:
```suggestion
     * @param int|array|null                $options Limit or associative array of options (option=>value)
```

Not sure if we document when it can be null?

lib/Elastica/Search.php
```diff
@@ -40,26 +40,19 @@ class Search
     public const OPTION_SEARCH_IGNORE_UNAVAILABLE = 'ignore_unavailable';
 
     /**
-     * @var BuilderInterface
+     * @var BuilderInterface|null
      */
-    private $_builder;
+    private $builder;
```
deguif: What the strategy for this kind of renaming. Are we allowed to rename all private properties or as it's a new major version can we apply it to protected and public ones?

thePanz: @ruflin asked to have a new PR with the changes related to public/protected methods :)
Private ones should be fine, with the protected ones (and public too) I would change them before releasing the new major 7.0 release. wdyt?

lib/Elastica/SearchableInterface.php
thePanz: Hum, not really.
btw: I don't really like such mix of different types to be accepted by those methods... any inputs?

lib/Elastica/Search.php
deguif: Yes it would be nice to be consistent.

I have a big PR on https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1569 for bringing support for elastica 7.x.
This Symfony bundle relies on some protected property like the `_name` of the Index class, so we have a room for changing that before publishing the GA release?

lib/Elastica/SearchableInterface.php
deguif: Allowing only array is possible, and removing support for the int?
But I'm not confident with the impact

lib/Elastica/Search.php
ruflin: Yes, timing before 7.0 GA seems to be good. Can we do this change potentially in an automated manner with the linter?

thePanz: IIRC the linter (php-cs-fixer) does not rename properties...

lib/Elastica/SearchableInterface.php
thePanz: on the same line: what do you think about the `string|array|AbstractQuery` on some methods? Should we be consistent and *only* allow object of AbstractQuery to be passed in?

ruflin:
I merged this one to have as many small changes as possible at the time. Please keep your above discussion going.

PR #1734, opened 2019-12-13
Fix handling of Search::OPTION_SEARCH_IGNORE_UNAVAILABLE inside Scroll

rattuscz:
I've encountered issue when using scroll combined with ignore_unavailable search option.

Scroll object sends all search options with secondary requests via scroll API, but those secondary don't allow all options.

This change simply adds another option that is managed by the Scroll object stores its state and removes it on the Scroll api requests except from initial one

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ All notable changes to this project will be documented in this file based on the
 
 ### Bugfixes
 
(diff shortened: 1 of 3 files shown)


ruflin:
@rattuscz LGTM. Thanks for adding the functional tests 🎉

PR #1742, opened 2020-01-09
PHPDoc and tests cleanup

thePanz:


lib/Elastica/Cluster.php
```diff
@@ -138,8 +127,6 @@ public function getClient()
      * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-nodes-info.html
      *
      * @param array $args Additional arguments
-     *
-     * @throws \Elastica\Exception\NotImplementedException
```
ruflin: Isn't that still needed? Or is this replaced by some "magic"?

lib/Elastica/Node.php
```diff
@@ -34,55 +34,43 @@ class Node
...
-     *
-     * @return $this Refreshed object
-     */
-    public function setId($id)
+    public function setId(string $id): self
     {
         $this->_id = $id;
+        $this->refresh();
 
-        return $this->refresh();
```
ruflin: Nit: Does it have to be 2 lines?

test/Elastica/Base.php
```diff
@@ -22,41 +22,6 @@ protected static function showDeprecated()
         \error_reporting(\error_reporting() | E_USER_DEPRECATED);
     }
 
-    protected function assertFileDeprecated($file, $deprecationMessage)
```
ruflin: Quick reminder for me: We don't need this anymore because ...?

test/Elastica/Base.php
```diff
@@ -104,20 +69,21 @@ protected function _getProxyUrl403(): string
...
-        if (null === $name) {
-            $name = \preg_replace('/[^a-z]/i', '', \strtolower(\get_called_class()).\uniqid());
-        }
+        $name = $name ?: static::buildUniqueId();
 
         $client = $this->_getClient();
-        $index = $client->getIndex('elastica_'.$name);
+        $index = $client->getIndex($name);
 
-        if ('elasticsearch' === \getenv('ES_HOST')) {
```
ruflin: This is not needed anymore?

test/Elastica/ClusterTest.php
```diff
@@ -3,74 +3,63 @@
 namespace Elastica\Test;
 
 use Elastica\Cluster;
-use Elastica\Cluster\Health;
 use Elastica\Node;
 use Elastica\Test\Base as BaseTest;
 
+/**
+ * @group functional
```
ruflin: Nice, didn't know you can set it on the class level

test/Elastica/Processor/AttachmentTest.php
```diff
@@ -66,7 +66,7 @@ public function testAttachmentAddPdf()
         $bulk->setIndex($index);
 
         $doc1 = new Document(null);
-        $doc1->addFile('data', BASE_PATH.'/data/test.pdf');
+        $doc1->addFile('data', __DIR__.'/../../data/test.pdf');
```
ruflin: Much nicer.

test/Elastica/ResponseFunctionalTest.php
```diff
@@ -0,0 +1,92 @@
...
+use Elastica\Mapping;
+use Elastica\Query;
+use Elastica\Query\MatchAll;
+use Elastica\Request;
+use Elastica\Test\Base as BaseTest;
+
+/**
+ * @group functional
+ */
+class ResponseFunctionalTest extends BaseTest
```
ruflin: Always happy to have more tests. Could you share some background if this is "in addition" or replacing something else?

test/Elastica/ResponseTest.php
```diff
@@ -2,82 +2,15 @@
...
 
+/**
+ * @group unit
+ */
 class ResponseTest extends BaseTest
 {
-    /**
-     * @group functional
-     */
-    public function testResponse()
```
ruflin: Ah, the other one is replacing this part?

test/Elastica/Processor/AttachmentTest.php
thePanz: :)

test/Elastica/ResponseTest.php
thePanz: yes, exactly. :)

test/Elastica/ResponseFunctionalTest.php
thePanz: Those are tests extracted from the ResponseTest

test/Elastica/Base.php
thePanz: This method was not used

lib/Elastica/Cluster.php
thePanz: This method is just a placeholder, as it is only throwing that exception.
I was in doubt to remove the method completely ... wdyt?

p365labs:
@thePanz is the PHPDOC a cleanup due to the ```NoSuperfluousPhpdocTags``` introduction inside phpcsfixer (v.2.16.0)?

lib/Elastica/Cluster.php
ruflin: I think my "idea" here was that if users see an method and get an exception with "not implemented" they are likely to contribute it. But seems like it never happened. I would say either we keep the docs entry or remove the method completly. I'm good with removing it completely. I guess there are 2-3 other similar methods.

thePanz:
@p365labs yes, that's what I a aiming to. I am manually cleaning up the code, and discovering some interesting cases (methods where the PHP docs is `return $this`, but the method does not return anything).

I will open another PR with the `NoSuperfluousPhpdocTags` flag enabled :)

lib/Elastica/Cluster.php
thePanz: Removed the "not-implemented" code!

lib/Elastica/Node.php
ruflin: I see, need to learn the rules :-D

thePanz:
@ruflin CI is now green, should we merge this? :)

thePanz:
@ruflin do you mean adding a changelog about never-implemented functions?

I believe it is bringing more confusion to the users reading it, and no more clarity

PR #1749, opened 2020-01-16
Fix Scroll:current() annotation

thePanz:
When analyzing a code using the iterator in a `foreach($scroll as $resultSet)` with phpstan, the following error is reported:

```
Argument of an invalid type Elastica\ResultSet|null supplied for foreach, only iterables are supported.
```

There is quite a debate about the return type of an `Iterator::current()`, as it should always return the item in the collection.
Invoking the `current()` when "out of bounds" of the collection is expected not to happen, as the `Iterator::valid()` should be called first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Backward Compatibility Breaks
 ### Added
 ### Changed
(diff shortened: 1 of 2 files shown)


thePanz:
@ruflin Changelog added :)

Tobion:
Ref. #1506

thePanz:
Didn't know about that PR @Tobion . The whole library should get a cleanup on the Iterator implementations :)

ruflin:
@thePanz I would not say no 😇

PR #1752, opened 2020-01-24
Improvements on the Reindex API

renanbr:
Reindex improvements:

- Add pipeline support
    - Add `Reindex::PIPELINE`
    - Add `$reindex->setPipeline(Pipeline $pipeline)`
- Improve refresh support
    - Add `$reindex->setRefresh(bool|string $value)`
- Improve query support
    - Add `$reindex->setQuery(AbstractQuery $query)`
- Add slices support
    - Add `Reindex::SLICES`
    - Add `Reindex::SLICES_AUTO`

... and also:

- Add `$pipeline->getId()`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/7.0.0-beta.3...master)
 ### Backward Compatibility Breaks
 ### Added
+* Added `Elastica\Reindex->setPipeline(Elastica\Pipeline $pipeline): void`. The link between the reindex and the pipeline is solved when `run()` is called, and thus the pipeline given doesn't need to be created before calling `setPipeline()`
+* Added `Elastica\Reindex->setRefresh(bool|string $value): void`. It accepts boolean and `REFRESH_*` constants from its class
+* Added `Elastica\Reindex->setQuery(Elastica\Query\AbstractQuery $query): void`
+* Added constants `PIPELINE`, `REFRESH_TRUE`, `REFRESH_FALSE`, `REFRESH_WAIT_FOR`, `SLICES` and `SLICES_AUTO` to `Elastica\Reindex`
+* Added `Elastica\Pipeline->getId(): ?string`
+
 ### Changed
 - Require elastica-php library >= v7.1.1, fixes an issue on Ingestion/Put() type-hinting
 - Require guzzle >= v6.3 as development library: fixes issues on PHP >= 7.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/7.0.0-beta.3...master)
 ### Backward Compatibility Breaks
 ### Added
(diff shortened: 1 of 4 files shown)


lib/Elastica/Reindex.php
```diff
@@ -184,6 +194,23 @@ public function setScript(Script $script)
...
+    {
+        $this->setParam(self::QUERY, $query);
+    }
+
+    public function setPipeline(Pipeline $pipeline): void
+    {
+        $this->setParam(self::PIPELINE, $pipeline);
+    }
+
+    public function setRefresh($value): void
```
thePanz: Should we force the value type to be `bool $value`?

renanbr: The valid values are: `true`, `false` and `wait_for`.
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-reindex.html#docs-reindex-api-query-params

We could also create some class constants for them.

thePanz: I like your solution, it would be cleaner IMHO and would allow us to validate the parameter... or should it be something for the lower-level client library? :thinking: 

@ruflin , what do you think?

ruflin: In general I like the idea of validation at the same time it can put a burden on "us". Class constant would definitively be helpful but i would not validate that only these are used. If a new param is introduced in ES, we need to add it to support it which I don't think it optimal. Happy to also only have the most common / used ones as constants (or none to get started).

thePanz: I quite agree with the "burden", I just wanted to have a more strict typing here.
IIRC Elasticsearch accepts `"true"` (as string) and `true` (as boolean) as equals. if that's the case: we can allow `setRefresh(string $value)`  and define 3 public constants as "true" "false" and "wait_for".

renanbr: I added some constants via fb3e933ec6d068c0a711f9321ae587593a439ab8

renanbr:
> Could you add a changelog entry?

done via 06ffaa7ebcc71d23ddf12d7fd595c1dc3e2a1570

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/7.0.0-beta.3...master)
 ### Backward Compatibility Breaks
 ### Added
+* Added `Elastica\Reindex->setPipeline(Elastica\Pipeline $pipeline): void`. The link between the reindex and the pipeline is solved when `run()` is called, and thus the pipeline given doesn't need to be created before calling `setPipeline()`
```
ruflin: Nit: Could you add the links to the PR's? I'm just realising we miss it in some entries below too :-(

renanbr: done aa5d5f5b83eb0b38ce6a4ac1921bf881b77ac139 :-)

lib/Elastica/Reindex.php
thePanz: Thank you for the new fixes, what about only allowing `string` as parameter here?

thePanz:
@ruflin LGTM, just one comment about the "setRefresh()" parameters

ruflin:
I merged this PR to get it in. I see @renanbr 👍 your comment so I assume there will be a follow up PR? :-)

renanbr:
> I merged this PR to get it in. I see @renanbr  your comment so I assume there will be a follow up PR? :-)

yes, I'll do it

lib/Elastica/Reindex.php
renanbr: see #1758

PR #1754, opened 2020-02-11
Phpdoc cleanup

thePanz:


ruflin:
Seems like CI does not like the change.

PR #1755, opened 2020-02-11
Restructure library folders

thePanz:
Checklist:
- [x] Move `lib/Elastica/` under `src/`
- [x] Move `test/Elastica/` under `tests/`
- [x] Adjust composer.json accordingly


test/phpunit.xhprof.xml
```diff
@@ -1,39 +0,0 @@
-<phpunit backupGlobals="false"
```
ruflin: We don't need it anymore?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Backward Compatibility Breaks
 ### Added
 ### Changed
+- Refactored code structure: use `src/` and `tests/` folders [#1755](https://github.com/ruflin/Elastica/pull/1755)
```
ruflin: I wonder if we should put this under BC for users which don't use composer?

test/phpunit.xhprof.xml
thePanz: bench-marking is disabled since years, and the tests also mention "remove this".
If you check the XML file: the path is even pointing to a not-existing file (`lib/Elastica/Tests` does not exist)

CHANGELOG.md
thePanz: The current branch is 7.x, where we have no releases.
Imho there is no braking in moving to a different structure.
And my guess is that everyone is using this library with composer, it's 2020! :-)

ruflin: I know at least in the past this was not the case and I'm pretty sure there is old code around that will break ;-) Will not hurt us to put it under breaking changes.

Not sure I can follow the argument around no release? Releases or not, it still makes it a breaking change? Lets put it there just for awareness.

thePanz:
@ruflin I will rebase this and add the BC notice to the changelog :)

PR #1757, opened 2020-02-12
Cleanup .git* files

thePanz:


.editorconfig
```diff
@@ -16,10 +16,6 @@ indent_style = space
 indent_size = 2
 indent_style = space
 
-[Vagrantfile]
```
ruflin: I definitively don't miss the old Vagrant file :-)

thePanz: So, let's merge this? :)

PR #1758, opened 2020-02-12
Allow only string as argument in Reindex->setRefresh()

renanbr:
Follow up https://github.com/ruflin/Elastica/pull/1752#discussion_r378137629


thePanz:
@renanbr can you rebase this? We moved the code from  `lib/Elastica/` to `src/, and the tests fom `test/Elastica/` to `tests/`

thePanz:
Thanks @renanbr , merged! :)

PR #1760, opened 2020-02-13
Update PHP-cs-fixer settings and fix errors

thePanz:
This is a first step towards the PSR-12
No additional changes have been made apart from running the `make fix-phpcs` command

### DIFF
--- a/src/Aggregation/BucketScript.php
+++ b/src/Aggregation/BucketScript.php
@@ -71,7 +71,8 @@ class BucketScript extends AbstractAggregation
     {
         if (!$this->hasParam('buckets_path')) {
             throw new InvalidException('Buckets path is required');
-        } elseif (!$this->hasParam('script')) {
+        }
+        if (!$this->hasParam('script')) {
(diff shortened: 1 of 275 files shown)


src/Aggregation/BucketScript.php
```diff
@@ -71,7 +71,8 @@ public function toArray(): array
     {
         if (!$this->hasParam('buckets_path')) {
             throw new InvalidException('Buckets path is required');
-        } elseif (!$this->hasParam('script')) {
```
ruflin: Was this converted automatically? I think both options will work but they are not 100% identical I would argue.

thePanz: yes, all the changes were done by the cs-fixer

thePanz: well, it is identical. as the if branch is throwing an exception :)

ruflin: damn, true :-D `thePanz 1 : 0 ruflin`

PR #1761, opened 2020-02-21
Cleanup Wildcard query code

thePanz:


src/Query/Wildcard.php
```diff
@@ -5,33 +5,56 @@
 /**
  * Wildcard query.
  *
- * @author Nicolas Ruflin <spam@ruflin.com>
```
thePanz: Removed by accident, will restore it

src/Query/Wildcard.php
```diff
@@ -5,33 +5,56 @@
...
+     * Rewrite methods: @see https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-multi-term-rewrite.html.
      */
-    public function __construct(string $key = '', ?string $value = null, float $boost = 1.0)
+    public const REWRITE_CONSTANT_SCORE = 'constant_score';
+    public const REWRITE_CONSTANT_SCORE_BOOLEAN = 'constant_score_boolean';
+    public const REWRITE_SCORING_BOOLEAN = 'scoring_boolean';
+
+    private $field;
+
+    public function __construct(string $field, string $value, float $boost = 1.0)
```
ruflin: I don't specifically remember it here but I know we had in the past reasons to not necesarily require all params in the constructur. The reasons was something like:

* The object is instantiated
* Passed to something else that sets the param and returned
* Then toArray is called

In summary, not all info is available on creation time. If this is the case here, not sure.

src/Query/Wildcard.php
ruflin: Does not matter TBH. I think it is nice to give "credit" here to contributors, but my name should probably be remove in most places.

src/Query/Wildcard.php
thePanz: It is quite weird that those constructors allows the creation of an "invalid" object, it leads to some weird anti-patterns IMHO.
What I'd like is to have queries (Terms, WildCard,..) being able to be constructed with the required parameters, and throw exceptions is case the required `field`is empty as an example.

In the scenario you described: it would be better to have the "something else" to be responsible to the creation of the object, with the right parameters. wdyt?

ruflin: Probably :-) I'm good with trying out what you suggest and if we get push back, we have someone that we can ask for the detailed use case!

thePanz: great! I'll continue working on this PR then :)
I'll open other PR for other classes, just to keep the diffs small

src/Query/Wildcard.php
thePanz: About contributors: what about having a link to https://github.com/ruflin/Elastica/graphs/contributors ?
But this is a topic for another PR :)

PR #1766, opened 2020-03-11
Allow Terms aggregations to define exact excludes and includes

### LINKED ISSUE #1764: Terms aggregation with exact values in include
After finally upgrading to Elasticsearch 7.x and Elastica 7.0.0-beta.3 I discovered that `Elastica\Aggregation\Terms::setInclude` no longer works due to type-hinting in #1549 . Looking at the docs, it makes sense but it ignores filtering by exact values [](https://www.elastic.co/guide/en/elasticsearch/reference/7.x/search-aggregations-bucket-terms-aggregation.html#_filtering_values_with_exact_values_2) which expects an array instead of string. 

It seems that a similar issue was brought up in #1514 regarding incorrect docblock which has most likely caused the automated type-hinting to fail.

thePanz:
Allow Terms aggregations to define:
- exact excludes
- exact includes
- include-partitions

Removed the `$flags` from the `setInclude()` as not supported by ES anymore

Closes #1764

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * The `Wildcard` Query's constructor now requires the `name` and `value` properties.
 
 ### Added
+* Added `AbstractTermsAggregation::setIncludeAsExactMatch()` [#1766](https://github.com/ruflin/Elastica/pull/1766) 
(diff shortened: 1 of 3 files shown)


ruflin:
The only part I'm worried is that we need to add more functions if ES decides to add something new here. But at the same time, this is kind of how Elastica works ...

thePanz:
@ruflin the functions here are helping to not memorize the ES endpoints. if something new comes up: , we can add a new method, but in the meanwhile the `->setParam()` will work fine as a temporary solution :)

Issue #1779, opened 2020-05-15
Version 6.x - Http Transport: Use forceMerge() fails because of HTTP request missing its content type


romainneutron:
The issue happens on Amazon Elasticsearch 6.8 with the latest version 6 of ruflin/Elastica

I'm using`foremerge()` to force-merge an index:
```
$result = $this->client->getIndex($index)->forcemerge([
    'max_num_segments' => 1,
])
```

Unfortunately, this fails, here's the `var_dump` of `$result`:

```
object(Elastica\Response)#606 (6) {
  ["_queryTime":protected]=>
  float(0.0024068355560303)
  ["_responseString":protected]=>
  string(0) ""
  ["_transferInfo":protected]=>
  array(37) {
    ["url"]=>
    string(162) "https://mydomain.amazonaws.com:443/my_index/_forcemerge?max_num_segments=1"
    ["content_type"]=>
    string(16) "application/json"
    ["http_code"]=>
    int(403)
    ["header_size"]=>
    int(193)
    ["request_size"]=>
    int(287)
    ["filetime"]=>
    int(-1)
    ["ssl_verify_result"]=>
    int(0)
    ["redirect_count"]=>
    int(0)
    ["total_time"]=>
    float(0.002392)
    ["namelookup_time"]=>
    float(2.8E-5)
    ["connect_time"]=>
    float(2.8E-5)
    ["pretransfer_time"]=>
    float(0.000111)
    ["size_upload"]=>
    float(0)
    ["size_download"]=>
    float(213)
    ["speed_download"]=>
    float(106500)
    ["speed_upload"]=>
    float(0)
    ["download_content_length"]=>
    float(213)
    ["upload_content_length"]=>
    float(0)
    ["starttransfer_time"]=>
    float(0.000114)
    ["redirect_time"]=>
    float(0)
    ["redirect_url"]=>
    string(0) ""
    ["primary_ip"]=>
    string(9) "10.0.8.52"
    ["certinfo"]=>
    array(0) {
    }
    ["primary_port"]=>
    int(443)
    ["local_ip"]=>
    string(12) "192.168.64.2"
    ["local_port"]=>
    int(40512)
    ["http_version"]=>
    int(3)
    ["protocol"]=>
    int(2)
    ["ssl_verifyresult"]=>
    int(0)
    ["scheme"]=>
    string(5) "HTTPS"
    ["appconnect_time_us"]=>
    int(28)
    ["connect_time_us"]=>
    int(28)
[...]

PR #1780, opened 2020-05-15
Use the HTTP content type as specified, whatever the content of the body

romainneutron:
This solve the issue I reported here at #1779


ruflin:
I would like to better understand why this is needed as I would think we should not have a different content type then the body.

deguif:
@romainneutron would you be able to rebase and add a changelog?

deguif:
@ruflin should be the last one planned for `7.1.1`

ruflin (approved):

PR #1783, opened 2020-05-22
Fixing bug in usage of ltrim

viable-hartman:
Pretty sure the intention of this was to remove the '/', not lose the path, which this would do. So fixing it. Looks like this has been in the code going back quite a while, so might want to port it back.


ruflin:
This indeed looks odd :-( I wonder why this didn't break other things.

@viable-hartman 

* Could you add a changelog entry?
* Do you see a way we could test this change?

ruflin:
@viable-hartman Thank you! Hoping you get some spare time soon ;-)

deguif:
@viable-hartman would you be able to add a changelog?

deguif:
@ruflin I just added a changelog here.
If you could a look as I planned it for `7.1.1` too ;)

ruflin (approved):

PR #1791, opened 2020-08-06
Allow string argument in AbstractUpdateAction::setRefresh

knsarsaba:
Ref #1475 

Currently, `AbstractUpdateAction::setRefresh` only accepts boolean values: 
- `true` - to perform an immediate refresh
- `false` - to not refresh at all

The [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-refresh.html), however, strongly suggests the usage of `wait_for` option to ensure a good cluster performance while doing the refresh operation. But the problem is that `wait_for` is of type `string`; thus, we cannot pass it as an argument to the method.

This modifies `AbstractUpdateAction::setRefresh` so that arguments of type `string` such as `wait_for` can also be passed to the method. Additionally, the return type of its getter counterpart, `AbstractUpdateAction::getRefresh`, has also been modified to mirror the accepted types of the setter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 * Ability to specify the type of authentication manually by the `auth_type` parameter (in the client class config) was added (allowed values are `basic, digest, gssnegotiate, ntlm`)
 ### Changed
-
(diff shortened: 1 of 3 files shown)


ruflin:
@knsarsaba Thanks for this contribution. It is so nice to be able to just merge such perfect contributions!

PR #1794, opened 2020-09-12
Support PHP 8.0

deguif:
#### Done in this PR:

- [x] Update composer constraint to allow PHP 8.0.
- [x] Update travis configuration in order to run tests on PHP 8.0 (nightly).
- [x] Update changelog

#### Done in previous PRs:
- [x] Deprecate `Elastica\Query\Match` class and add a new `Elastica\Query\MatchQuery` class as match is now a reserved keyword. [#1799]
- [x] Update phpunit version in order to be compatible with PHP 8.0. [#1811]
- [x] Bump `aws/aws-sdk-php` package version to fix a bug with PHP 8.0. [#1798]

#### Done in external PRs:
- [x] Wait for `aws/aws-sdk-php` to be compatible with PHP 8. [https://github.com/aws/aws-sdk-php/pull/2079]
- [ ] Wait for `elasticsearch/elasticsearch` to be compatible with PHP 8.0 and remove `--ignore-platform-reqs` composer option.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,9 +15,11 @@ matrix:
   fast_finish: true
   include:
     - php: 7.3
-      env: RUN_TEST=true  ES_VERSION=73
+      env: RUN_TEST=true ES_VERSION=73
     - php: 7.4
-      env: RUN_TEST=true  ES_VERSION=73
+      env: RUN_TEST=true ES_VERSION=73
+    - php: nightly
+      env: RUN_TEST=true ES_VERSION=73 COMPOSER_FLAGS="--ignore-platform-reqs"
     # Testing the library with different ES versions
     - php: 7.2
       env: RUN_TEST=true ES_VERSION=72
--- a/tests/ClientConfigurationTest.php
+++ b/tests/ClientConfigurationTest.php
@@ -15,9 +15,9 @@ class ClientConfigurationTest extends TestCase
     public function testInvalidDsn(): void
     {
         $this->expectException(\Elastica\Exception\InvalidException::class);
-        $this->expectExceptionMessage("DSN 'test:0' is invalid.");
+        $this->expectExceptionMessage("DSN ':0' is invalid.");
 
-        ClientConfiguration::fromDsn('test:0');
+        ClientConfiguration::fromDsn(':0');
     }
 
     public function testFromSimpleDsn(): void
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Ability to specify the type of authentication manually by the `auth_type` parameter (in the client class config) was added (allowed values are `basic, digest, gssnegotiate, ntlm`)
 * Added `if_seq_no` / `if_primary_term` to replace `version` for [optimistic concurrency control](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/optimistic-concurrency-control.html)
 * Added `Elastica\Aggregation\PercentilesBucket` aggregation [#1806](https://github.com/ruflin/Elastica/pull/1806)
+* Supported PHP 8.0 [#1794](https://github.com/ruflin/Elastica/pull/1794)
 ### Changed
 * Allow `string` such as `wait_for` to be passed to `AbstractUpdateAction::setRefresh` [#1791](https://github.com/ruflin/Elastica/pull/1791)
 * Changed the return type of `AbstractUpdateAction::getRefresh` to `boolean|string` [#1791](https://github.com/ruflin/Elastica/pull/1791)
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         }
     ],
     "require": {
-        "php": "^7.2",
+        "php": "^7.2 || ^8.0",
         "ext-json": "*",
         "psr/log": "^1.0",
         "elasticsearch/elasticsearch": "^7.1.1 !=7.4.0"


src/Util.php
```diff
@@ -221,7 +221,7 @@ public static function getParamName($class)
 
         $parts = \explode('\\', $class);
         $last = \array_pop($parts);
-        $last = \preg_replace('/Query$/', '', $last); // for BoolQuery
+        $last = \preg_replace('/(Query|_)$/', '', $last); // for BoolQuery and Match_
```
deguif, line 224: Replaced `_` suffix here, but the method can be overriden to return `match` instead. WDYT @ruflin ?

src/QueryBuilder/DSL/Query.php
```diff
@@ -72,9 +72,9 @@ public function getType(): string
      *
      * @param mixed $values
      */
-    public function match(?string $field = null, $values = null): Match
+    public function match(?string $field = null, $values = null): Match_
```
deguif, line 75: BC break here, but impact should be very low.

deguif:
One failure is remaining on PHP 8.0 but it's related to aws SDK.
See https://github.com/aws/aws-sdk-php/pull/2079 for details and the fix that should be merged soon (I hope).

src/Query/Match.php
```diff
@@ -2,180 +2,17 @@
 
 namespace Elastica\Query;
 
+\trigger_error('Elastica\Query\Match is deprecated. Use Match instead. From PHP 8 match is reserved word and this class will be removed in further Elastica releases', \E_USER_DEPRECATED);
```
jdreesen, line 5:
```suggestion
\trigger_error('Elastica\Query\Match is deprecated. Use Match_ instead. From PHP 8 match is reserved word and this class will be removed in further Elastica releases', \E_USER_DEPRECATED);
```

deguif: Thanks @jdreesen

ruflin:
@deguif Thanks for getting this started. We had in the past a similar issue with a naming conflict with the `Bool` query. What we did back then is rename it to `BoolQuery` instead: https://github.com/ruflin/Elastica/blob/master/src/Query/BoolQuery.php I wonder if we should follow the same pattern here instead of using the `_` for consistency? `MatchQuery`?

.travis.yml
```diff
@@ -15,20 +15,25 @@ matrix:
...
-      env: RUN_TEST=true  ES_VERSION=73
+      env: RUN_TEST=true ES_VERSION=73
     - php: 7.4
-      env: RUN_TEST=true  ES_VERSION=73
+      env: RUN_TEST=true ES_VERSION=73
+    - php: nightly
+      env: RUN_TEST=true ES_VERSION=73 COMPOSER_FLAGS="--ignore-platform-reqs"
     # Testing the library with different ES versions
-    - php: 7.2
+    - php: 7.3
```
ruflin, line 24: Any chance you could push the changes for PHP 7.3 into a separate PR sow e can get it in quickly? Then this PR gets smaller and we know potential failures are only related to these changes.

src/Query/Match.php
```diff
@@ -2,180 +2,17 @@
...
 namespace Elastica\Query;
 
+\trigger_error('Elastica\Query\Match is deprecated. Use Match_ instead. From PHP 8 match is reserved word and this class will be removed in further Elastica releases', \E_USER_DEPRECATED);
+
 /**
  * Match query.
  *
- * @author F21
- * @author WONG Wing Lun <luiges90@gmail.com>
+ * This class is for backward compatibility reason. For PHP 8 and above use Match_ as Match is reserved.
```
ruflin, line 10: Nice, so we can keep the old file around for compatiblity but mark it deprecated. This is great.

src/Util.php
ruflin: Not sure I follow? If we go with `Query` suffix instead of `_`, will it just work?

tests/PipelineTest.php
```diff
@@ -66,7 +66,8 @@ public function testPipelineCreate(): void
 
         $result = $pipeline->create();
 
-        $this->assertContains('acknowledged', $result->getData());
+        $this->assertArrayHasKey('acknowledged', $result->getData());
```
ruflin, line 69: Are these changes related to PHP 8? The new phpunit version? This this new version also work in 7.x? 

This feeds into the same question like the php 7.3 part above: Could we have these change in a separate PR to make it clear they are not related to the PHP 8.0 changes?

src/Util.php
deguif: Yes

ruflin:
@deguif This will also need an update of the CHANGELOG and I wonder if we should also update the readme that it now running at least tests with PHP 8.0?

tests/PipelineTest.php
deguif: yes, so increase of php version should be merged before increasing phpunit version to 9.3 which supports php 8.0 but doesn't support php 7.2 anymore.

ruflin: It seems like it is also compatible with 7.3, so if we jump to 7.3, we can go to 9 first in a separate PR and do this changes?

src/QueryBuilder/DSL/Query.php
BackEndTea: Not for users running php 7.4 and up.

As for them if they extend it they can speciyf their method returns the `Match` class.

deguif: It won't impact users on 7.4 just for covariance.
But probability that users care about this return is very low ;)
By care I mean `instanceof` or anything that is tied to the previous class name

GrahamCampbell:
> Bump aws/aws-sdk-php package version to fix a bug with PHP 8.0 (aws/aws-sdk-php#2079).

GrahamCampbell:
> Bump aws/aws-sdk-php package version to fix a bug with PHP 8.0 (aws/aws-sdk-php#2079).

There are definitely more similar bugs like the one that I fixed. Amazon's job is to fix that, though. I have just fixed the one that affected me.

GrahamCampbell:
Also, Guzzle, a dependency of the AWS SDK is not ready for PHP 8 yet.

ruflin:
Is there option to exclude the AWS part for 8.0 for now?

GrahamCampbell:
Guzzle is now ready, btw. 7.2.0 supports PHP 8.0.

tests/ClientConfigurationTest.php
```diff
@@ -15,9 +15,9 @@ class ClientConfigurationTest extends TestCase
     public function testInvalidDsn(): void
     {
         $this->expectException(\Elastica\Exception\InvalidException::class);
-        $this->expectExceptionMessage("DSN 'test:0' is invalid.");
+        $this->expectExceptionMessage("DSN ':0' is invalid.");
 
-        ClientConfiguration::fromDsn('test:0');
```
deguif, line 20: Since PHP 8.0 RC1 a 0 port number is considered valid: See https://github.com/php/php-src/pull/6152
For details see how it behaves on other PHP versions: https://3v4l.org/ZYhJc

deguif:
@ruflin I think this PR is now ready.
Testing on PHP 8.0 is now green with all the changes done in previous PRs.

I will later update the PHP version in travis configuration to 8.0 and also remove the composer flag `--ignore-platform-reqs` when `elasticsearch/elasticsearch` will officially support PHP 8.0.

ruflin:
@deguif Nice! Thank you!

PR #1795, opened 2020-09-15
[PHP 8] Support PHPUnit 9.4

deguif:


ruflin:
Is it really that we drop support on PHP 7.2? Or is it more that the testing is focused on 7.3? Do we have any feature at the moment or planned that will not work in 7.2?

deguif:
I can let support with the composer package and only test with 7.3 as PHPUnit version compatible with PHP 8.0 is `9.3` and doesn't support PHP 7.2 anymore.
So I'm reverting the change on composer.json and only drop testing on 7.2.
Will you be OK with that?

deguif:
Or maybe we can test on PHP 8.0 with more recent deps so that we can support two PHPUnit version in the meantime?

ruflin:
@deguif Is there an easy way to do this with 2 different deps?

deguif:
Will not be easy I think as PHPUnit is installed with phive and is not a dependency in composer.json dev dependencies.

ruflin:
I wonder if there is a way to run some CI with 7.2 without phpunit that will blow up in case we start using a feature only available in >=7.3. With this, we could switch everything to 7.3 but at least still know if we start to break 7.2. Otherwise we are just "blind" and might break it accendtially.

deguif:
Maybe we could move phpunit dependency to composer, so that composer would do the work for us.

ruflin:
@thePanz Perhaps you can chime in here as I see you were the one that touched https://github.com/ruflin/Elastica/blob/master/phive.xml recently? ^

deguif:
What would be nice with setting phpunit as a dev dependency is that we can then allow multi versions (`^8.5 || ^9.3`), and composer would automatically choose version `8.5.x` on PHP 7.2 as `9.3` is not compatible with PHP 7.2 ;)

ruflin:
@deguif I definitively like this idea. Are there any disadvantages to this?

deguif:
That's a common setup among PHP open source projects. Even if using phar is also common ;)
So I don't know what to answer about disadvantages.

ruflin:
@deguif I would say, lets do it then!

thePanz:
Fine for me to move PhpUnit to a `dev` dependency.
Unfortunately I don't see a quick solution for https://github.com/phar-io/phive/issues/219 :(

I would keep the rest of the tools (php-cs-fixer, php-coveralls, an maybe phpstan later) to be managed by `phive`. WDYT?

ruflin:
I'm good with that, no preference on my end.

deguif:
Just got PHPUnit 9.3 working alongside PHPUnit 8.5 in https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1707 in order to support PHP 7.2 and PHP 8.0 at the same time.
Will do a PR here in a near future.

deguif:
Here's the PR that will now install phpunit with composer instead of phive to allow several versions.
See https://github.com/ruflin/Elastica/pull/1801

deguif:
I'm closing this one. I'm not satisfied anymore with the branch name.
I'm opening another one to add support for PHPUnit 9.3/9.4 now that #1801 was merged.

PR #1798, opened 2020-09-25
[PHP 8.0] Use AWS SDK package version compatible

deguif:


ruflin (approved):
LGTM, could you add a changelog entry for awareness?

deguif:
That's a dev dependency, so is a changelog required ?

ruflin:
@deguif Good point. Probably not so I merged.

PR #1799, opened 2020-09-25
[PHP 8.0] Deprecate Match query class and introduce MatchQuery

deguif:


src/Query/Match.php
```diff
@@ -2,180 +2,17 @@
...
- * @author WONG Wing Lun <luiges90@gmail.com>
+ * This class is for backward compatibility reason. For PHP 8 and above use MatchQuery as Match is reserved.
+ *
+ * @author Nicolas Ruflin <spam@ruflin.com>
  *
- * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-match-query.html
+ * @deprecated Use MatchQuery instead. From PHP 8 match is reserved word and this class will be removed in further Elastica releases
  */
-class Match extends AbstractQuery
+class Match extends MatchQuery
```
ruflin, line 16: Nice!

ruflin (approved):

ruflin:
@deguif This is great. Thanks!

mfn:
Is there any time table for a release containing this fix?

I'm starting to run our tests again PHP8 and hit the "parse error" as outlined.

Although I should be able to use master, it actually doesn't work for me though I think it should => any ideas?
```
composer require ruflin/elastica:7.0.x-dev --prefer-dist
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - This package requires php ^8.0 but your PHP version (7.4.12) does not satisfy that requirement.


Installation failed, reverting ./composer.json to its original content.
```
I do run this with PHP 7.4.12 but according to this in master, it should work 🤔 https://github.com/ruflin/Elastica/blob/41ab9bf1ebca493832021e3eb8d199b50f6c0e79/composer.json#L15

Thank you

mfn:
PS: got it working by adding `--ignore-platform-reqs` to the composer command, still puzzles why it wouldn't work without it

deguif:
The behaviour you described is weird, as our CI jobs use several PHP 7.x versions and there are no issues like the one you encountered.
We also ignore platform requirements when testing on PHP 8 as the official elasticsearch library is not yet ready. (I see you commented on the PR on this repository too).

I think we should be able to publish a new `7.1.0` version in the coming days, but it would be better to wait for the elasticsearch official library to be officially compatible with PHP 8.0 before publishing the `7.1.0`.

reedy:
Not sure the best place to ask this (let me know if you want me to file an issue instead).. Is there any chance of getting this backported to the 6.x branch? Or some version also adding the MatchQuery class/alias also into 6.x so it can be used instead?

While presumably Elastica 6.x might not support PHP 8.0 (but 6.x is currently maintained based on the Readme), where a project can/will support multiple versions of elastica/elasticsearch php, being able to have the same class available in all versions would be really useful

Thanks!

olix21:
@reedy It could be interesting to open a dedicated issue/MR for that

reedy:
Will do!

PR #1800, opened 2020-10-02
[PHP 8.0] Fix \ReflectionParameter deprecations

deguif:
See https://php.watch/versions/8.0/deprecated-reflectionparameter-methods for details about the deprecations addressed here.

### DIFF
--- a/tests/QueryBuilder/DSL/AbstractDSLTest.php
+++ b/tests/QueryBuilder/DSL/AbstractDSLTest.php
@@ -72,16 +72,14 @@ abstract class AbstractDSLTest extends BaseTest
      */
     protected function _getHintName(\ReflectionParameter $param)
     {
-        if ($param->isCallable()) {
-            return 'callable';
+        if (null === $type = $param->getType()) {
+            return null;
         }
 
-        if ($param->isArray()) {
-            return 'array';
-        }
-
-        if ($class = $param->getClass()) {
-            return $class->getName();
+        if (\in_array($type->getName(), ['array', 'callable'], true)
+            || !$param->getType()->isBuiltin()
+        ) {
+            return $type->getName();
         }
 
         return null;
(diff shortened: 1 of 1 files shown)


tests/QueryBuilder/DSL/AbstractDSLTest.php
```diff
@@ -67,21 +67,16 @@ protected function _getDefaultValue(\ReflectionParameter $param)
...
-    protected function _getHintName(\ReflectionParameter $param)
+    protected function _getHintName(\ReflectionParameter $param): ?string
     {
-        if ($param->isCallable()) {
-            return 'callable';
-        }
-
-        if ($param->isArray()) {
-            return 'array';
+        if (null === $type = $param->getType()) {
```
ruflin, line 72: I must confess I'm still trying to get my head around this change. It does not change anything from a user perspective?

If this line would be removed, would it also be catched by line 82? Or would in_array blow up it type is null?

deguif: Sorry for the headache 😉
This line is just to ensure that `$type` is not null, so that it can be used safely below.

tests/QueryBuilder/DSL/AbstractDSLTest.php
```diff
@@ -67,21 +67,16 @@ protected function _getDefaultValue(\ReflectionParameter $param)
...
+            return null;
         }
 
-        if ($class = $param->getClass()) {
-            return $class->getName();
+        if (\in_array($type->getName(), ['array', 'callable'], true)
+            || !$param->getType()->isBuiltin()
```
deguif, line 77: By the way `$type` can be used here, I'm fixing it.

tests/QueryBuilder/DSL/AbstractDSLTest.php
deguif: Parameter type is stored inside `$type` to avoid calling the getter multiple time

ruflin (approved):
LGTM

I was not sure if we need a changelog or not. One the one hand, behaviour stays the same at the same time it is a nice improvement to be compatible with PHP 8.0. I'll merge and if you want, follow up with a PR for the changelog.

PR #1801, opened 2020-10-02
Use phpunit as a composer dev dependency

deguif:


deguif:
Just one failure (not related probably) which is related to memory usage?

ruflin:
Interesting failure: https://github.com/ruflin/Elastica/blob/master/tests/BulkTest.php#L685 It is only a little bit more and interestingly it only fails on one version of PHP. Also I didn't see this fail before, I wonder if it indicates a memory leak.

I'll run it again to see what happens, but for historical reasons here the output:

```
1) Elastica\Test\BulkTest::testMemoryUsage
Failed asserting that 1.3075842391681565 is less than 1.3.
/home/travis/build/ruflin/Elastica/tests/BulkTest.php:685
```

ruflin:
@deguif The test keeps failing even with multiple runs. I'm wondering what increased the memory usage here. Is it the different version of phpunit? At least the value seems to be consistent, so an easy fix would be just to increase the memory check to 1.31 or similar.

deguif:
Let's do that then.
I'm updating the failing test.

ruflin:
@deguif great thanks. Can you ping me with a comment when you updated it so I don't miss it as I don't think I get pings on new commits.

deguif:
@ruflin It's green now.

ruflin:
Great, thanks for the change. Lets see if the 1.31 will hunt us one day ;-)

PR #1796, opened 2020-09-18
uncommented version in _params on updateDocument() to support Elatica…

consolari:
… 7.9.1

The version parameter is removed from elasticsearch, tested on Elasticsearch 7.9.1

PR #1803, opened 2020-10-06
Version option deprecation fix and replacements

thomaspicquet:
As stated in https://github.com/ruflin/Elastica/pull/1796

* removed `version` and `version_type`
* added replacements `if_seq_no` and `if_primary_term`

### DIFF
--- a/src/AbstractUpdateAction.php
+++ b/src/AbstractUpdateAction.php
@@ -65,6 +65,71 @@ class AbstractUpdateAction extends Param
     }
 
     /**
+     * Sets the sequence number of a document for use with optimistic concurrency control.
+     *
+     * @param int $number Sequence Number
+     *
+     * @return $this
+     *
+     * @see https://www.elastic.co/guide/en/elasticsearch/reference/6.8/optimistic-concurrency-control.html
+     */
+    public function setSequenceNumber($number)
+    {
+        return $this->setParam('if_seq_no', $number);
+    }
+
+    /**
+     * Returns document version.
+     *
+     * @return int|string Document version
+     */
+    public function getSequenceNumber()
+    {
+        return $this->getParam('if_seq_no');
+    }
+
+    /**
+     * @return bool
+     */
+    public function hasSequenceNumber()
+    {
+        return $this->hasParam('if_seq_no');
+    }
+
+    /**
+     * Sets the prmary term of a document for use with optimistic concurrency control.
+     *
+     * @param int $term Primary Term
+     *
+     * @return $this
+     *
+     * @see https://www.elastic.co/guide/en/elasticsearch/reference/6.8/optimistic-concurrency-control.html
+     */
+    public function setPrimaryTerm($term)
+    {
+        return $this->setParam('if_primary_term', $term);
+    }
+
+    /**
+     * Returns document version.
+     *
+     * @return int|string Document version
+     */
+    public function getPrimaryTerm()
+    {
+        return $this->getParam('if_primary_term');
+    }
+
+    /**
+     * @return bool
+     */
+    public function hasPrimaryTerm()
+    {
+        return $this->hasParam('if_primary_term');
+    }
+
+    /**
+     * @deprecated
      * Sets the version of a document for use with optimistic concurrency control.
      *
      * @param int $version Document version
@@ -75,10 +140,13 @@ class AbstractUpdateAction extends Param
      */
     public function setVersion($version)
     {
-        return $this->setParam('version', (int) $version);
+        \trigger_error('Elastica\AbstractUpdateAction\setVersion is deprecated. Elastica\AbstractUpdateAction\setSequenceNumber and Elastica\AbstractUpdateAction\setPrimaryTerm instead.', E_USER_DEPRECATED);
+
+        return $this;
     }
 
     /**
+     * @deprecated
      * Returns document version.
      *
      * @return int|string Document version
@@ -89,6 +157,7 @@ class AbstractUpdateAction extends Param
     }
 
     /**
+     * @deprecated
      * @return bool
      */
     public function hasVersion()
@@ -97,6 +166,7 @@ class AbstractUpdateAction extends Param
     }
 
     /**
+     * @deprecated
      * Sets the version_type of a document
      * Default in ES is internal, but you can set to external to use custom versioning.
      *
@@ -106,10 +176,13 @@ class AbstractUpdateAction extends Param
      */
     public function setVersionType($versionType)
     {
-        return $this->setParam('version_type', $versionType);
+        \trigger_error('Elastica\AbstractUpdateAction\setVersionType is deprecated. Elastica\AbstractUpdateAction\setSequenceNumber and Elastica\AbstractUpdateAction\setPrimaryTerm instead.', E_USER_DEPRECATED);
+
+        return $this;
     }
 
     /**
+     * @deprecated
      * Returns document version type.
      *
      * @return int|string Document version type
@@ -120,6 +193,7 @@ class AbstractUpdateAction extends Param
     }
 
     /**
+     * @deprecated
      * @return bool
      */
     public function hasVersionType()
--- a/src/AbstractUpdateAction.php
+++ b/src/AbstractUpdateAction.php
@@ -438,4 +438,26 @@ class AbstractUpdateAction extends Param
 
         return \array_filter($this->getParams());
     }
+
+    /**
+     * @param array $responseData
+     *
+     * @return $this
+     */
+    public function setVersionParams(array $responseData) :self
+    {
+        if (isset($responseData['_version'])) {
+            $this->setVersion($responseData['_version']);
+        }
+
+        if (isset($data['_seq_no'])) {
+            $this->setSequenceNumber($responseData['_seq_no']);
+        }
+
+        if (isset($data['_primary_term'])) {
+            $this->setPrimaryTerm($responseData['_primary_term']);
+        }
+
+        return $this;
+    }
 }
--- a/src/AbstractUpdateAction.php
+++ b/src/AbstractUpdateAction.php
@@ -59,7 +59,7 @@ class AbstractUpdateAction extends Param
      *
      * @return string Index name
      */
-    public function getIndex()
+    public function getIndex(): string
     {
         return $this->getParam('_index');
     }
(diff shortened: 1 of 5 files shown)


deguif:
I'm still reading the changelog for 7.x and I'm not sure anymore what to do: https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking-changes-7.0.html#breaking_70_api_changes

It clearly states: 
```
Elasticsearch maintains a numeric version field for each document it stores. 
That field is incremented by one with every change to the document. 
Until 7.0.0 the API allowed using that field for optimistic concurrency control, i.e., making
a write operation conditional on the current document version. 
Sadly, that approach is flawed because the value of the version doesn’t always uniquely represent
a change to the document. If a primary fails while handling a write operation,
it may expose a version that will then be reused by the new primary.

Due to that issue, internal versioning can no longer be used and is replaced by a new method
based on sequence numbers. See Optimistic concurrency control for more details.

Note that the external versioning type is still fully supported.
```

src/AbstractUpdateAction.php
```diff
@@ -65,6 +65,71 @@ public function getIndex()
...
     /**
+     * Sets the sequence number of a document for use with optimistic concurrency control.
+     *
+     * @param int $number Sequence Number
+     *
+     * @return $this
+     *
+     * @see https://www.elastic.co/guide/en/elasticsearch/reference/6.8/optimistic-concurrency-control.html
+     */
+    public function setSequenceNumber($number)
```
deguif, line 76:
```suggestion
    public function setSequenceNumber(int $number)
```

src/AbstractUpdateAction.php
```diff
@@ -65,6 +65,71 @@ public function getIndex()
...
+    /**
+     * Sets the prmary term of a document for use with optimistic concurrency control.
+     *
+     * @param int $term Primary Term
+     *
+     * @return $this
+     *
+     * @see https://www.elastic.co/guide/en/elasticsearch/reference/6.8/optimistic-concurrency-control.html
+     */
+    public function setPrimaryTerm($term)
```
deguif, line 108:
```suggestion
    public function setPrimaryTerm(int $term)
```

src/AbstractUpdateAction.php
```diff
@@ -65,6 +65,71 @@ public function getIndex()
...
+    /**
+     * @return bool
+     */
+    public function hasSequenceNumber()
+    {
+        return $this->hasParam('if_seq_no');
+    }
+
+    /**
+     * Sets the prmary term of a document for use with optimistic concurrency control.
```
deguif, line 100:
```suggestion
     * Sets the primary term of a document for use with optimistic concurrency control.
```

src/AbstractUpdateAction.php
```diff
@@ -65,6 +65,71 @@ public function getIndex()
...
+    {
+        return $this->setParam('if_primary_term', $term);
+    }
+
+    /**
+     * Returns document version.
+     *
+     * @return int|string Document version
+     */
+    public function getPrimaryTerm()
```
deguif, line 118:
```suggestion
    public function getPrimaryTerm(): int
```

src/AbstractUpdateAction.php
```diff
@@ -65,6 +65,71 @@ public function getIndex()
...
+    {
+        return $this->getParam('if_primary_term');
+    }
+
+    /**
+     * @return bool
+     */
+    public function hasPrimaryTerm()
```
deguif, line 126:
```suggestion
    public function hasPrimaryTerm(): bool
```

src/AbstractUpdateAction.php
```diff
@@ -65,6 +65,71 @@ public function getIndex()
...
+     */
+    public function getSequenceNumber()
+    {
+        return $this->getParam('if_seq_no');
+    }
+
+    /**
+     * @return bool
+     */
+    public function hasSequenceNumber()
```
deguif, line 94:
```suggestion
    public function hasSequenceNumber(): bool
```

src/AbstractUpdateAction.php
```diff
@@ -65,6 +65,71 @@ public function getIndex()
...
+    {
+        return $this->setParam('if_seq_no', $number);
+    }
+
+    /**
+     * Returns document version.
+     *
+     * @return int|string Document version
+     */
+    public function getSequenceNumber()
```
deguif, line 86:
```suggestion
    public function getSequenceNumber(): int
```

src/Client.php
```diff
@@ -311,8 +311,11 @@ public function updateDocument($id, $data, $index, array $options = []): Respons
             && ($data->isAutoPopulate() || $this->getConfigValue(['document', 'autoPopulate'], false))
         ) {
             $responseData = $response->getData();
-            if (isset($responseData['_version'])) {
-                $data->setVersion($responseData['_version']);
```
deguif, line 315: Not sure this should be removed finally. As it's still documented as returned.

src/Client.php
```diff
@@ -277,8 +277,8 @@ public function updateDocument($id, $data, $index, array $options = []): Respons
 
             $docOptions = $data->getOptions(
                 [
-                    'version',
-                    'version_type',
```
deguif, line 281: Seems the right fix as this not documented anymore: https://www.elastic.co/guide/en/elasticsearch/reference/7.x/docs-update.html

src/Index.php
```diff
@@ -215,8 +215,11 @@ public function addDocument(Document $doc): Response
             if (isset($data['_id']) && !$doc->hasId()) {
                 $doc->setId($data['_id']);
             }
-            if (isset($data['_version'])) {
-                $doc->setVersion($data['_version']);
```
deguif, line 219: Not sure this should be removed finally. As it's still documented as returned.

src/Index.php
```diff
@@ -272,10 +275,7 @@ public function getDocument($id, array $options = []): Document
             $data = [];
         }
 
-        $document = new Document($id, $data, $this->getName());
-        $document->setVersion($result['_version']);
```
deguif, line 276: Not sure this should be removed finally. As it's still documented as returned.

src/Index.php
```diff
@@ -272,10 +275,7 @@ public function getDocument($id, array $options = []): Document
...
-
-        return $document;
+        return new Document($id, $data, $this->getName());
```
deguif, line 278: We can probably add the `seq_no`  and `primary_term` to document too as it's documented as returned: https://www.elastic.co/guide/en/elasticsearch/reference/7.9/docs-get.html#docs-get-api-response-body

thomaspicquet: I was debating doing that since the param for the document update is different, I think it needs to be separated between what we get back and what is sent for a document update

src/Index.php
thomaspicquet: I'll put back setVersion, just remove it for a document update

src/Client.php
```diff
@@ -277,15 +277,15 @@ public function updateDocument($id, $data, $index, array $options = []): Respons
...
-                    'routing',
-                    'percolate',
-                    'parent',
-                    'retry_on_conflict',
                     'consistency',
-                    'replication',
+                    'if_primary_term',
+                    'if_seq_no',
```
thomaspicquet, line 282: @deguif @ruflin after reading the docs, I think we should remove those from the default params. It shouldn't be forced since there are set when it gets a document right now

thomaspicquet:
@deguif @ruflin this is ready for review

src/AbstractUpdateAction.php
```diff
@@ -59,72 +59,118 @@ public function setIndex($index): self
...
+     * Sets the version of a document.
+     *
+     * @param int $version Document version
+     *
+     * @return $this
+     *
+     * @see https://www.elastic.co/blog/versioning
      */
-    public function hasVersionType()
+    public function setVersion(int $version): self
```
deguif, line 156: BC break too.

deguif (commented):
Many type-hints changes not in scope of this PR.
That would break the backward compatibility.

src/AbstractUpdateAction.php
```diff
@@ -59,72 +59,118 @@ public function setIndex($index): self
...
+    {
+        return $this->setParam('_version', $version);
```
deguif, line 158:
```suggestion
        return $this->setParam('version', $version);
```

thomaspicquet:
> Many type-hints changes not in scope of this PR.
> That would break the backward compatibility.

I reverted the type hints outside of the scope

src/AbstractUpdateAction.php
```diff
@@ -65,66 +65,115 @@ public function getIndex()
     }
 
     /**
-     * Sets the version of a document for use with optimistic concurrency control.
+     * Sets the version parameters of a document for use with optimistic concurrency control.
      *
-     * @param int $version Document version
+     * @return $this
+     */
+    public function setVersionParams(array $responseData): self
```
ruflin, line 72: Do we still need this method or could we remove it?

src/Bulk.php
```diff
@@ -323,9 +323,7 @@ protected function _processResponse(Response $response): ResponseSet
                         if (!$data->hasId() && isset($bulkResponseData['_id'])) {
                             $data->setId($bulkResponseData['_id']);
                         }
-                        if (isset($bulkResponseData['_version'])) {
-                            $data->setVersion($bulkResponseData['_version']);
-                        }
+                        $data->setVersionParams($bulkResponseData);
```
ruflin, line 326: Now I see where we need this. Is there a better way to do this?

ruflin (commented):
Is any of this going to be a breaking change for some of the users?

src/AbstractUpdateAction.php
thomaspicquet: its still used to set the document version on GET

src/Bulk.php
thomaspicquet: I'm not sure there's a better way, I was trying to avoid to copy/paste code

ruflin (approved):
LGTM.

@deguif Could you also give a 👍  or check if your concerns were addressed?

src/Bulk.php
deguif: I think copy/paste in this case is fine, adding another public method just for avoiding calling each setter doesn't bring so much value.

deguif:
@ruflin I'm not really confident with the impacts it can have but else I'm 👍

ruflin:
@deguif That leaves a lot of room for interpretation 😆  The good news is, that this is master. I'll move forward with it.

PR #1808, opened 2020-10-11
Added support for pools using DSN

Nyholm:
I was about to update https://github.com/Happyr/elastica-dsn which basically adds DSN support to 6.x. But then I saw that 7.x will have support for DNS but not for multiple hosts. 

Im sorry that I introduce a new dependency like this for such a small feature. I do need this feature in my app and I really wanted to try out nyholm/dsn:2.0 before I tag a stable release.

This will support any crazy DSN you may think of. See the full specification here: https://github.com/nyholm/dsn

Let me know what you think.

---

The PR is in draft as it should not be merged before nyholm/dsn:2.0 is released.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added `Elastica\Aggregation\PercentilesBucket` aggregation [#1806](https://github.com/ruflin/Elastica/pull/1806)
 * Added `weighted_avg` to aggregations DSL [#1814](https://github.com/ruflin/Elastica/pull/1814)
 * Supported PHP 8.0 [#1794](https://github.com/ruflin/Elastica/pull/1794)
+* Added support for defining a connection pool with DSN. Example `pool(http://127.0.0.1 http://127.0.0.2/bar?timeout=4)` [#1808](https://github.com/ruflin/Elastica/pull/1808)
 ### Changed
 * Allow `string` such as `wait_for` to be passed to `AbstractUpdateAction::setRefresh` [#1791](https://github.com/ruflin/Elastica/pull/1791)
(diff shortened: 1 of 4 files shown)


ruflin:
Thanks for the addition. I'm always a bit on the fence about adding more external dependency at the same time, get rid of code to maintain is great. My initial reaction was that this might remove quite a bit of code on our end up it seems it is not the case. What else does the DSN lib to add as features to Elastica besides multi hosts or are we just missing multi hosts?

Do you know when 2.0 is planned to be shipped?

Nyholm:
> What else does the DSN lib to add as features to Elastica besides multi hosts or are we just missing multi hosts?

Not much. The usual. =)
- Safer parsing with some error messages
- The ability to do DSN functions like `roundRobin(http://host1 http://host2)`

> My initial reaction was that this might remove quite a bit of code on our end up it seems it is not the case.

True. I wanted to keep BC. Ideally I would remove the `connections` key and force people to use a DSN function. That would also make it possible to validate the options to make sure the user dont misspell stuff.

> Do you know when 2.0 is planned to be shipped?

I plan to release it as soon we are happy here.

ruflin (commented):
Haha, I missed the most important bit that you are the owner of the lib 🤦‍♂️ That simplifies things of course as you are in control of it and the release.

I'm good with moving forward here. This PR would also need a changelog entry.

Nyholm:
I've released 2.0, PR is ready for review.

thomaspicquet (approved):

ruflin:
@Nyholm Just wanted to merge this but unfortunately there is now a conflict with master :-( Could you rebased / merge in master? Sorry for the delay on my end which probably caused this.

Nyholm:
No worries. 

PR is rebased. Tests should pass.

ruflin:
@Nyholm Thanks for the contribution!

Nyholm:
Wohoo. Thank you for merging

PR #1811, opened 2020-10-19
[PHP 8.0] Allow PHPUnit 9.4

deguif:
This PR will allow to support PHP 8.0 in #1794 

It allows to run PHPUnit **8.5** for PHP **7.2** support, but also to run PHPUnit **9.4** for versions supporting it (7.3, 7.4, 8.0) as PHP **8.0** is not supported in PHPUnit **8.5**.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
     "require-dev": {
         "aws/aws-sdk-php": "^3.155",
         "guzzlehttp/guzzle": "^6.3",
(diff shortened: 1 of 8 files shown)


deguif:
ping @ruflin

ruflin:
Thanks! Merged to get it in. Perhaps we should have a changelog entry for it even though it is dev only?

Issue #1812, opened 2020-10-19
Missing Pipeline processors


ThibautSF:
Comparing declared classes in [Elastica\Processor](https://elastica.io/api/latest/namespaces/Elastica.Processor.html) against [Elasticsearch documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/ingest-processors.html), some processors seem never to have been implemented.

- [ ] [Bytes](https://www.elastic.co/guide/en/elasticsearch/reference/current/bytes-processor.html)
- [ ] [Circle](https://www.elastic.co/guide/en/elasticsearch/reference/current/ingest-circle-processor.html)
- [ ] [CSV](https://www.elastic.co/guide/en/elasticsearch/reference/current/csv-processor.html)
- [ ] [Dissect](https://www.elastic.co/guide/en/elasticsearch/reference/current/dissect-processor.html)
- [ ] [Foreach](https://www.elastic.co/guide/en/elasticsearch/reference/current/foreach-processor.html)
- [ ] [GeoIP](https://www.elastic.co/guide/en/elasticsearch/reference/current/geoip-processor.html)
- [ ] [Grok](https://www.elastic.co/guide/en/elasticsearch/reference/current/grok-processor.html)
- [ ] [Gsub](https://www.elastic.co/guide/en/elasticsearch/reference/current/gsub-processor.html)
- [ ] [HTML strip](https://www.elastic.co/guide/en/elasticsearch/reference/current/htmlstrip-processor.html)
- [ ] [Inference](https://www.elastic.co/guide/en/elasticsearch/reference/current/inference-processor.html)
- [ ] [Pipeline](https://www.elastic.co/guide/en/elasticsearch/reference/current/pipeline-processor.html)
- [ ] [Script](https://www.elastic.co/guide/en/elasticsearch/reference/current/script-processor.html)
- [ ] [Set security user](https://www.elastic.co/guide/en/elasticsearch/reference/current/ingest-node-set-security-user-processor.html)
- [ ] [URL decode](https://www.elastic.co/guide/en/elasticsearch/reference/current/urldecode-processor.html)
- [ ] [User agent](https://www.elastic.co/guide/en/elasticsearch/reference/current/user-agent-processor.html)

ThibautSF:
Link to #1810 & #1809

ruflin:
Thanks for starting to track this!

PR #1813, opened 2020-10-19
Add foreach processor

ThibautSF:
Add Foreach processor (and test files)

Complete list in #1812


ThibautSF:
Note: I will do some other processor in the list, only did one for now because want to check : 
- if the procedure is ok (nothing is missing)
- check the list, if everything can be added

src/Processor/ForeachProcessor.php
```diff
@@ -0,0 +1,69 @@
+<?php
+
+namespace Elastica\Processor;
+
+/**
+ * Elastica Attachment Processor.
```
deguif, line 6:
```suggestion
 * Elastica Foreach Processor.
```

src/Processor/ForeachProcessor.php
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/foreach-processor.html
+ */
+class ForeachProcessor extends AbstractProcessor
+{
+    public const DEFAULT_IGNORE_MISSING_VALUE = false;
+
+    /**
+     * Foreach constructor.
+     */
```
deguif, line 19: Doesn't bring much value, could probably be removed

ThibautSF: Resolved in https://github.com/ruflin/Elastica/pull/1813/commits/e7659be9d92078ad97a48bd994f580eaa01abb1a

deguif:
As `field` and `processor` are required I would make them parameters to the constructor to avoid constructing an invalid foreach processor. Is there any reason it can't be done this way?
I see your comment about `setRawProcessor` but I'm  not sure to understand why we need this.

Wouldn't a `setProcessor(new Remove('user_agent'))` or `new ForeachProcessor('field', new Remove('user_agent'))` work as expected (to use the example you put in annotation of the `setRawProcessor()`method)?

ThibautSF:
> As `field` and `processor` are required as stated in the table options of the official documentation ([official doc](https://www.elastic.co/guide/en/elasticsearch/reference/master/foreach-processor.html#foreach-processor)) I would make them parameters to the constructor to avoid constructing an invalid foreach processor. Is there any reason it can't be done this way?

PHP is one of the few languages which doesn't allow the constructor overloading... And I often forgot that.
My initial idea was to use 2 constructors :
- __construct(string $field, AbstractProcessor $processor)
- __construct(string $field, array $rawprocessor)

Similar behaviour can be made with public static class and private constructor: but thus mean that this class won't be instantiated like the other processors. Might be a bit strange as final user?

> I see your comment about `setRawProcessor` but I'm not sure to understand why we need this.

The setRawProcessor is needed to easily add a processor which doesn't exist as an Elastica class.

> Wouldn't a `setProcessor(new Remove('user_agent'))` or `new ForeachProcessor('field', new Remove('user_agent'))` work as expected (to use the example you put in annotation of the `setRawProcessor()`method)?

Yes of course.


But while I'm thinking about it, the most frequent instance will be with AbstractProcessor.
A thing that can be done is : 
```php
     /**
     * Foreach constructor.
     */
    public function __construct(string $field, AbstractProcessor $processor)
    {
        $this->setField($field);
        $this->setProcessor($processor);
    }
```

Thus if the user wants to use a custom processor: 
```php
$processor = ForeachProcessor('field', null);
$processor->setRawProcessor(['remove' => ['field' => 'user_agent']]);
```

And in other case:
```php
$processor = ForeachProcessor('field', new Remove('user_agent'));
```

deguif:
@ThibautSF I better understand now with your explanation.

We can kind of simulate use of constructor overloading with:
```php
/**
 * @param AbstractProcessor|array $processor
 */
public function __construct(string $field, $processor)
{
    $this->setField($field);
    
    if ($processor instanceof AbstractProcessor) {
        $this->setProcessor($processor);
    } elseif (\is_array($processor)) {
        $this->setRawProcessor($processor);
    } else {
        throw new \TypeError('Expected xxx');
    }
}
``` 

That can later be replaced with native union type when dropping support for PHP versions < 8.0

Not sure if we have to support the raw processor format, but @ruflin will be of better help for this.

ThibautSF:
I was working on 
```php
    /**
     * Foreach constructor.
     *
     * @param AbstractProcessor $processor
     */
    public function __construct(string $field, ?AbstractProcessor $processor)
    {
        $this->setField($field);
        $this->setProcessor($processor);
    }
```

But your solution looks better.

src/Processor/ForeachProcessor.php
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /**
+     * Set processor.
+     *
+     * @param AbstractProcessor
+     *
+     * @return $this
+     */
+    public function setProcessor(?AbstractProcessor $processor): self
```
deguif, line 51:
```suggestion
    public function setProcessor(AbstractProcessor $processor): self
```

src/Processor/ForeachProcessor.php
```diff
@@ -0,0 +1,90 @@
...
+     * Param's name
+     * Picks the last part of the class name and makes it snake_case
+     * You can override this method if you want to change the name.
+     *
+     * @return string name
+     */
+    protected function _getBaseName()
+    {
+        return self::PROCESSOR_NAME;
+    }
```
deguif, line 89: To avoid this, maybe the class should be named `ForEach`?

src/Processor/ForeachProcessor.php
```diff
@@ -0,0 +1,90 @@
...
+    public function __construct(string $field, $processor)
+    {
+        $this->setField($field);
+
+        if ($processor instanceof AbstractProcessor) {
+            $this->setProcessor($processor);
+        } elseif (\is_array($processor)) {
+            $this->setRawProcessor($processor);
+        } else {
+            throw new \TypeError('Expected Elastica\Processor\AbstractProcessor or array');
```
deguif, line 30:
```suggestion
            throw new \TypeError(\sprintf('Argument 2 passed to %s::__construct() must be of type %s|array, %s given.', self::class, AbstractProcessor::class, \is_object($processor) ? \get_class($processor) : \gettype($processor)));

```

src/Processor/ForeachProcessor.php
ThibautSF: "foreach" is a php construct, don't think it's a good idea.

And my PHP code check doesn't like that.
```
syntax error, unexpected 'ForEach' (T_FOREACH), expecting identifier (T_STRING)
Unexpected 'foreach'. Expected 'Name'.intelephense(1001)
Expression is not writable.intelephense(1024)
```

deguif: Sorry @ThibautSF I meant `ForEachProcessor`, yes `Foreach` is a reserved keyword.

ThibautSF: `Elastica\Processor\ForeachProcessor._getBaseName()` is an override of default `Elastica\Param._getBaseName()` which only calls and return the static method `Elastica\Utils.getParamName($class)`

```php
    /**
     * Tries to guess the name of the param, based on its class
     * Example: \Elastica\Query\MatchAll => match_all.
     *
     * @param object|string Object or class name
     * @param mixed $class
     *
     * @return string parameter name
     */
    public static function getParamName($class)
    {
        if (\is_object($class)) {
            $class = \get_class($class);
        }

        $parts = \explode('\\', $class);
        $last = \array_pop($parts);
        $last = \preg_replace('/Query$/', '', $last); // for BoolQuery and MatchQuery

        return self::toSnakeCase($last);
    }
```

As we can see, this method with "ForEachProcessor" class may return "for_each_processor".
Note: When I tried "ForeachProcessor" before adding the method override, the generated processor for elasticsearch was "foreach_processor" thus I pushed https://github.com/ruflin/Elastica/pull/1813/commits/e7659be9d92078ad97a48bd994f580eaa01abb1a
> Fix processor name being "foreach_processor" instead of "processor" due to class name (note: 'Foreach' is reserved)

So, the override is there to ensure that the param name used will be "foreach".
Note: The presence of the const aims to simplify processor name modification (idk, in case Elasticsearch dev wants to change "foreach" with "for")

deguif: Thanks for your detailed explanation, I better understand now ;)

What about in this case adding:
```php
abstract class AbstractProcessor extends Param
{
    protected function _getBaseName()
    {
        $shortName = (new \ReflectionClass($this))->getShortName();
        $shortName = \preg_replace('/Processor$/', '', $shortName);

        return Util::toSnakeCase($shortName);
    }
}
```

Or add `\preg_replace('/Processor$/', '', $last);` in `Util::getParamName()` but impact can be bigger.

For fun I benchmarked the use of `ReflectionClass` vs `explode` to see if it impacts on performance ;)
See https://3v4l.org/9rlBF

ThibautSF: > What about in this case adding:
> 
> ```
> abstract class AbstractProcessor extends Param
> {
>     protected function _getBaseName()
>     {
>         $shortName = (new \ReflectionClass($this))->getShortName();
>         $shortName = \preg_replace('/Processor$/', '', $shortName);
> 
>         return Util::toSnakeCase($shortName);
>     }
> }
> ```
> 
> Or add `\preg_replace('/Processor$/', '', $last);` in `Util::getParamName()` but impact can be bigger.

I don't know... the current implementation is working for every case (generic) and was designed to allow custom name (adaptability).

In my opinion, we need to keep in mind that the issue ("foreach" being reserved) which forces us to use the non-generic way is uncommon. And if tomorrow we need a class "FooProcessor" which need to be interpreted by _getBaseName() as "foo_processor" the `preg_replace('/Processor$/', '', $shortName);` might be annoying.


> For fun I benchmarked the use of `ReflectionClass` vs `explode` to see if it impacts on performance ;)
> See https://3v4l.org/9rlBF

You might be interested to look at this comment https://www.php.net/manual/fr/function.get-class.php#114568
It's 6 years old so might need to benchmark again, but `strrpos()` looks like better than `explode()`

ruflin:
Sorry, a bit late to comment but I'm curious what we need to move this forward? @ThibautSF @deguif ?

ThibautSF:
Should be OK

Just, code will need to be a bit updated if there is any modification due to #1824

ruflin:
Sounds like we should resolve #1824 then. @deguif WDYT?

deguif:
@ThibautSF can you rebase the PR. Overriding `_getBaseName()` is not needed anymore as the base class will do the job now.

ThibautSF:
First time rebasing a PR : Followed this guide https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

deguif:
It seems rebase has not worked as it should as there are lot of commit not introduced in this PR.
Would you mind allowing maintainers to push to your branch ?

ThibautSF:
Should be already allowed.

image

deguif:
@ThibautSF should be good now.

src/Processor/ForeachProcessor.php
```diff
@@ -0,0 +1,77 @@
...
+     */
+    public function setField(string $field): self
+    {
+        return $this->setParam('field', $field);
+    }
+
+    /**
+     * Set processor.
+     *
+     * @param AbstractProcessor
```
deguif, line 46: Not needed

ThibautSF: 29d2c6f

deguif (approved):

ruflin:
Merging this even though I assume we didn't really align yet on https://github.com/ruflin/Elastica/issues/1824 ;-) But we should not block contributions based on this discussion and we can add an alias at any time.

deguif:
This one shouldn't be impacted as it conflicts with reserved keywords and will need a `Processor` suffix ;)

PR #1771, opened 2020-04-21
Add WeightedAvg aggregation

romainneutron:
Replace #1770

PR #1814, opened 2020-10-20
Add weighted_avg to aggregration DSL

deguif:
This follows PR #1771 which added this aggregation but didn't expose it to DSL.


deguif:
@ruflin just rebased, can be merged now.

PR #1815, opened 2020-10-20
Fix sampler aggregation DSL

deguif:


src/QueryBuilder/DSL/Aggregation.php
```diff
@@ -422,13 +422,12 @@ public function adjacency_matrix(string $name): AdjacencyMatrix
...
-    /** sampler aggregation.
+    /**
+     * sampler aggregation.
      *
      * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-sampler-aggregation.html
-     *
-     * @param string $name
      */
-    public function sampler($name): Sampler
+    public function sampler(string $name): Sampler
```
deguif, line 430: BC break as type-hint is changed, but this is currently breaking when calling the constructor with another type-hint as constructor only accepts a `string` for name argument.

ruflin:
Could you add a changelog entry? We could consider this a bugfix :-)

deguif:
@ruflin changelog added, can be merged ;)

deguif:
Merging as I added the changelog entry ;)

ruflin:
Thanks!

ruflin:
BTW glad to see you used squash merge to keep the history clean.

PR #1817, opened 2020-10-20
Fix deprecated exceptions guzzle request option

deguif:
This was deprecated in 6.0 version and  a `http_errors` was introduced at this time. This won't work anymore with guzzle 7.x as `exceptions` option was removed.


deguif:
By the way I'm not sure which `guzzle\guzzle` versions are supported. It seems it's only 6.x currently.
@ruflin can you confirm?

deguif:
@ruflin should now be ready to merge.

ruflin:
@deguif PR LGTM. For the guzzle question, TBH not sure. Why do you mean exactly by "supported"?

deguif:
By `supported` I mean the minimal version which works with `ruflin/elastica` when using guzzle transport shipped there.

ruflin:
Good question, we haven't discuss this TBH. We should probably specify it?

deguif:
A hard composer requirement would probably not be a good choice.
But adding a conflict could work, but can be sources of issues, as using `ruflin/elastica` doesn't mean you will use its guzzle transport.
Or at least check at runtime when using the guzzle transport, that would be the less intrusive solution.

deguif:
@ruflin I think this can be merged. The point about minimum version of `guzzle/guzzle` supported can be addressed in a future PR. Currently it seems it supports v6 and upper, but it needs some adaptations for v7 ;)

PR #1818, opened 2020-10-20
Remove HHVM proxy detection

deguif:
HHVM is not compatible anymore with PHP (last version supporting PHP was v3.30 released on 17th december 2018).
I think it's time move forward as many packages already did and drop code related to HHVM.


ruflin:
@deguif I'm good with removing it but perhaps you can some details on why you remove it?

deguif:
@ruflin sure. I just added a better description.

ruflin:
Thank you. This reminds me how old some code in Elastica is :-)

PR #1821, opened 2020-10-20
Fix transient test

deguif:
This should fix this failing test which occurs regularly:
```
There was 1 failure:

1) Elastica\Test\ClientFunctionalTest::testEndpointParamsRequest
Failed asserting that 2 is identical to 1.
```

This was occurring when replica(s) shard(s) was/were not started before adding a document.
There also a `wait_for_active_shards` options that could be passed to query when creating an index, but would require to adapt `Index::create()` method (I will probably propose another PR for this).

To better understand the change here the json part Elasticsearch returns when calling `GET /_cluster/state`:
```json
{
   "routing_table":{
      "indices":{
         "elasticatestclientfunctionaltestffbf":{
            "shards":{
               "0":[
                  {
                     "state":"STARTED",
                     "primary":true,
                     "node":"4uHcxqlPRVKRZw9Ay8MerQ",
                     "relocating_node":null,
                     "shard":0,
                     "index":"elasticatestclientfunctionaltestffbf",
                     "allocation_id":{
                        "id":"1GMyYkpQRTq4IHsy_YwKmg"
                     }
                  },
                  {
                     "state":"STARTED",
                     "primary":false,
                     "node":"FB_0fAuzSEuj9pS5FUF_Rw",
                     "relocating_node":null,
                     "shard":0,
                     "index":"elasticatestclientfunctionaltestffbf",
                     "allocation_id":{
                        "id":"sqc10Z0gQ-K3ipZef6YiFQ"
                     }
                  }
               ]
            }
         }
      }
   }
}
```

### DIFF
--- a/tests/ClientFunctionalTest.php
+++ b/tests/ClientFunctionalTest.php
@@ -937,6 +937,7 @@ class ClientFunctionalTest extends BaseTest
     public function testEndpointParamsRequest(): void
     {
         $index = $this->_createIndex();
+        $this->_waitForAllocation($index);
         $client = $index->getClient();
         $doc = new Document(null, ['foo' => 'bar']);
         $doc->setRouting('first_routing');
@@ -952,7 +953,7 @@ class ClientFunctionalTest extends BaseTest
         $this->assertArrayHasKey('index_total', $response->getData()['indices'][$index->getName()]['total']['indexing']);
 
         $this->assertSame(
-            1,
+            2,
             $response->getData()['indices'][$index->getName()]['total']['indexing']['index_total']
         );
     }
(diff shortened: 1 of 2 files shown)


tests/ClientFunctionalTest.php
```diff
@@ -952,7 +953,7 @@ public function testEndpointParamsRequest(): void
         $this->assertArrayHasKey('index_total', $response->getData()['indices'][$index->getName()]['total']['indexing']);
 
         $this->assertSame(
-            1,
+            2,
```
deguif, line 956: `2` is the right value as there is a replica shard configured.

ruflin:
Thanks for the fix.

PR #1822, opened 2020-10-20
Review options handling for index creation

deguif:
There are multiple changes. Maybe I should split the PR.

- Better check for options parameters by throwing a `TypeError` if not an expected type.
- Ensure `$options` is always converted to an array (I was planning to add deprecations with `\trigger_error('xxx', E_USER_DEPRECATED)` for other types in a following PR)
- Avoid duplicating the delete call when handling boolean or array (now it's unified and treat only the array option, as boolean option filled the array appropriatly).
- More user friendly messages when option(s) is/are invalid.

### DIFF
--- a/src/Index.php
+++ b/src/Index.php
@@ -381,26 +381,31 @@ class Index implements SearchableInterface
      */
     public function create(array $args = [], $options = null): Response
     {
-        if (\is_bool($options) && $options) {
+        $options = $options ?? [];
+
+        if (\is_bool($options)) {
+            $options = ['recreate' => $options];
+        } elseif (!\is_array($options)) {
+            throw new \TypeError(\sprintf('Argument 2 passed to %s::create() must be of type array|bool|null, %s given.', self::class, \is_object($options) ? \get_class($options) : \gettype($options)));
+        }
+
+        $invalidOptions = \array_diff(\array_keys($options), $allowedOptions = [
+            'recreate',
+        ]);
+
+        if (1 === $invalidOptionCount = \count($invalidOptions)) {
+            throw new InvalidException(\sprintf('"%s" is not a valid option. Allowed options are "%s".', \implode('", "', $invalidOptions), \implode('", "', $allowedOptions)));
+        }
+
+        if ($invalidOptionCount > 1) {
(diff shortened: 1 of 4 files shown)


ruflin:
I'll probably need a bit more details in the PR description and a changelog ;-)

deguif:
Just added a better description. I'm adding a changelog.

src/Index.php
```diff
@@ -381,26 +381,31 @@ public function refresh(): Response
      */
     public function create(array $args = [], $options = null): Response
```
ruflin, line 382: General thought: Should we become more strict on what is allowed to be passed to options? The code below is way too complex as we piled onto it over to time not have a breaking change I guess. Good news is, your PR seems to simplify it at least.

tests/Base.php
```diff
@@ -96,7 +96,9 @@ protected function _createIndex(?string $name = null, bool $delete = true, int $
         $client = $this->_getClient();
         $index = $client->getIndex($name);
 
-        $index->create(['settings' => ['index' => ['number_of_shards' => $shards, 'number_of_replicas' => 1]]], $delete);
+        $index->create(['settings' => ['index' => ['number_of_shards' => $shards, 'number_of_replicas' => 1]]], [
```
ruflin, line 99: Out of curiosity: The old one would still work but we discourage using it like this?

src/Index.php
ruflin: Update: Just read your PR message again. ++ on deprecation additions.

ruflin (approved):
I suggest we get this in as is and follow up with deprecation and more clean up. Does that work for you?

tests/Base.php
deguif: Yes the old one will still work, but it will be deprecated so this is to avoid triggering the future deprecation.

src/Index.php
deguif: Yes exactly, I will deprecate all types except array. So this will be backward compatible and will be easy to identify which part to drop when bumping a new major version.

deguif: By the way there's this PR #1823 that is tied to deprecations, would be nice if you could have a look so that we can go forward and use this mechanism for all futures deprecations. I have several ones in mind ;)

PR #1823, opened 2020-10-21
Review deprecations triggering

deguif:
See https://github.com/elastic/elasticsearch-php/pull/973 for details about why unsilenced deprecations can cause issues.
Symfony link related in the elasticsearch issue is outdated as it points to current version which use `trigger_deprecation`, see [this link](https://symfony.com/doc/4.4/contributing/code/conventions.html#deprecating-code) instead.

Here what is stated in the symfony documentation:
```
Without the @-silencing operator, users would need to opt-out from deprecation notices. Silencing 
swaps this behavior and allows users to opt-in when they are ready to cope with them (by adding 
a custom error handler like the one used by the Web Debug Toolbar or by the PHPUnit bridge).
```
I added `symfony/deprecation-contracts` to better handle deprecations (they take care of silencing the deprecations).

### DIFF
--- a/src/Query/Match.php
+++ b/src/Query/Match.php
@@ -2,7 +2,7 @@
 
 namespace Elastica\Query;
 
-\trigger_error('Elastica\Query\Match is deprecated. Use Elastica\Query\MatchQuery instead. From PHP 8 match is reserved word and this class will be removed in further Elastica releases', \E_USER_DEPRECATED);
+trigger_deprecation('ruflin/elastica', '7.1.0', 'The "%s" class is deprecated, use "%s" instead. "match" is a reserved keyword starting from PHP 8.0. It will be removed in 8.0.', Match::class, MatchQuery::class);
 
 /**
  * Match query.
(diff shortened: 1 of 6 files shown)


src/Query/Match.php
```diff
@@ -2,7 +2,7 @@
 
 namespace Elastica\Query;
 
-\trigger_error('Elastica\Query\Match is deprecated. Use Elastica\Query\MatchQuery instead. From PHP 8 match is reserved word and this class will be removed in further Elastica releases', \E_USER_DEPRECATED);
+trigger_deprecation('ruflin/elastica', '7.0.1', 'The "%s" class is deprecated, use "%s" instead. "match" is a reserved keyword starting from PHP 8.0. It will be removed in 8.0.', Match::class, MatchQuery::class);
```
deguif, line 5: Maybe `7.1.0` instead? I don't know which next version you have in mind @ruflin

ruflin: Lets put 7.1.0 for now.

ruflin (commented):
I'm undecided if we should just go with silencing with `@` or add the dependency. I'm always a bit torn on adding more dependencies at the same time it is nice to unify.

deguif:
Yes I added silencing with `@` on new PRs which triggers deprecations, but `symfony/deprecation-contracts` would add a way to enforce proper deprecations messages and would be a very small dependency introduction, but I'm also mitigated with its interests as this can be done with conventions with `@trigger` ;)

src/Query/Match.php
deguif: Perfect, I'm updating it and previously merged deprecations

nicolas-grekas (approved):
Thanks, I fully support this change!
Silencing the deprecations is a must because this is what makes them opt-in:
- without the silencing operator, ppl MUST configure their prod to ignore the deprecations: that means ppl are forced to opt-out;
- with silencing, deprecations are silent, unless one decides to use a tool that gathers them and reports them appropriately (such as Symfony's): this is opt-in.

About `symfony/deprecation-contracts`, it's been designed to help the community standardize on a very simple and generic signature and implementation. I wish all PHP projects would adopt it. That would help build better tools to display nice deprecation reports.

:+1: on my side!

ruflin (approved):

PR #1816, opened 2020-10-20
Allow guzzlehttp/guzzle 7.2 version

deguif:
Support `guzzlehttp/guzzle:7.2` as that's where PHP 8.0 support was added.
It requires some work:
- [x] Fix deprecated option not available anymore [#1817]
- [x] Fix `ConnectException` extending `TransferException` now instead of `RequestException` [#1825]

PR #1825, opened 2020-10-22
Fix Guzzle ConnectException inheritance change

deguif:
Second PR to prepare support for Guzzle 7.x support.
Guzzle `ConnectException` now inherits `TransferException` instead of `RequestException`, see https://github.com/guzzle/guzzle/pull/2541 or https://github.com/guzzle/guzzle/blob/master/CHANGELOG.md#changed-3

The tests updated in this PR were relying on this exception being triggered but catch `RequestException`.

I think there's no need to add a changelog for this as it only impacts tests, I will by the way add a changelog for the PR which will introduce testing with Guzzle 7.2, [#1816] to claim that Guzzle 7.x is now officially supported.


deguif:
@ruflin should be the last PR to fully support Guzzle 7.x

ruflin (approved):

PR #1826, opened 2020-10-23
Deprecate global_agg() DSL, replace it by global()

deguif:
Since PHP 7.0, there's no more reason to not use reserved keyword as class method name.

See original [RFC](https://wiki.php.net/rfc/context_sensitive_lexer).
See [code sample](https://3v4l.org/JPLmc)

### DIFF
--- a/src/QueryBuilder/DSL/Aggregation.php
+++ b/src/QueryBuilder/DSL/Aggregation.php
@@ -227,9 +227,19 @@ class Aggregation implements DSL
      *
      * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-global-aggregation.html
      */
-    public function global_agg(string $name): GlobalAggregation
+    public function global(string $name): GlobalAggregation
     {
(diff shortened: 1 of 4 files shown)


src/QueryBuilder/DSL/Aggregation.php
```diff
@@ -227,11 +227,21 @@ public function scripted_metric(
...
     {
         return new GlobalAggregation($name);
     }
 
+    /**
+     * @deprecated since Elastica 7.0.1, use the "global()" method instead.
+     */
+    public function global_agg(string $name): GlobalAggregation
+    {
+        @\trigger_error(\sprintf('The "%s()" method is deprecated since Elastica 7.0.1, use global() instead.', __METHOD__), E_USER_DEPRECATED);
```
deguif, line 240: Is related to https://github.com/ruflin/Elastica/pull/1823
So it can be adapted here once the linked PR gets merged, or vice versa.

ruflin (approved):

PR #1827, opened 2020-10-23
Remove specific hack for global aggregation

deguif:
There is a specific hack for `global_aggregation`.
`ParentAggregation` which has the same issue does it by overriding `_getBasename()`, so let's do it the same way.


ruflin:
++ on unifying. Seems like CI is not happy.

deguif:
CI is now green, just a test to adapt ;)

ruflin (approved):

PR #1831, opened 2020-10-26
Refactor Util::toSnakeCase()

deguif:
Current implementation is buggy when string is not starting with an upper letter.
For fun I also benchmarked current implement vs new implementation: https://3v4l.org/a9EdW

Here's the previous result when testing snake casing with `fooBar`:
```
1) Elastica\Test\UtilTest::testToSnakeCase
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'foo_bar'
+'oo_bar'
```


deguif:
Just added test covering the change.

ruflin (approved):
I wonder how this bug managed to stay around for so long and nobody noticed? Thanks for fixing. Probably worth a mention in the changelog.

Nice benchmarks. Does it mean 8.0 is slower? :-D

deguif:
@ruflin I'm adding a changelog as bug fix.
I think bug is pretty limited as it's used with class name and class names always start with upper case (at least here).

ruflin (approved):

PR #1832, opened 2020-10-26
Deprecate Util::getParamName()

deguif:
See comment: https://github.com/ruflin/Elastica/pull/1813#discussion_r509259083
I'm proposing to deprecate `Elastica\Util::getParamName()` as it's too generic and used in context of multiple classes with different naming conventions like `Query`, `Aggregation` and `Processor` (it checks for example `Query` suffix for every class, even when the class is not related to a Query). Instead I moved logic to `Query` suffix in the base Query class. This can be replicated for base classes `Aggregation` and `Processor` later to handle their suffix in an unified way.


ruflin:
Can you share some details behind this change?

deguif:
Just added a description to share the details I have in mind behind this change.

ruflin (approved):

PR #1833, opened 2020-10-26
Add native_constant_invocation CS rule

deguif:


ruflin (approved):
LGTM, seems to be in line with the function part.

PR #1842, opened 2020-10-27
Use late state binding class constant

deguif:
See ongoing RFC about deprecations for PHP 8.1 https://wiki.php.net/rfc/deprecations_php_8_0


ruflin (approved):
++ on getting rid of the "old" name even if they will not be removed.

PR #1843, opened 2020-10-27
Use argument unpacking

deguif:
See this old RFC: https://wiki.php.net/rfc/argument_unpacking


ruflin (approved):
Nice, I didn't even know PHP supports this. And it was available since 5.6!

Issue #1846, opened 2020-10-27
6.x branch DateRange aggregation does not support timezone parameter


dinamic:
Hey guys 👋

Thanks for keeping an eye on and maintaining this library!

Could we add support for the timezone on the date range aggregation, please? Is it a matter of just adding a setter on the DateRange class?

Here's the official documentation:
- https://www.elastic.co/guide/en/elasticsearch/reference/6.8/search-aggregations-bucket-daterange-aggregation.html#time-zones

deguif:
See #1847 thats targets master.
Can probably be backported to 6.x.

deguif:
#1847 was merged into master.

PR #1847, opened 2020-10-27
Add time zone option for date range aggregation

deguif:
See #1846


ruflin (approved):
Someone is on 🔥  Thanks!

PR #1848, opened 2020-10-27
Deprecate Range::setKeyedResponse() in favor of setKeyed()

deguif:
`setKeyed()` is used in:
- `Elastica\Aggregation\Percentiles`
- `Elastica\Aggregation\PercentilesBucket`

Let's use it there too. Feels more natural.


ruflin (approved):
We need to make sure to clean all these up before we go with 8.0 ;-)

PR #1851, opened 2020-10-28
Fix wrong namespaces PSR-4/PSR-0

deguif:


ruflin (approved):
Interesting that it still worked 🤔

PR #1856, opened 2020-10-29
Fix PHPUnit expectations misorder

deguif:
PHPUnit methods like `assertEquals`, `assertSame` are waiting the expected result as first argument and the current result as second. Sometimes they're misordered in the tests.


ruflin (approved):

ruflin:
CI does not seem to be happy.

deguif:
Green now, I'm merging.

PR #1859, opened 2020-10-30
Allow to pass native endpoint options to index creation

deguif:
This PR will allow to pass endpoint query parameters to index creation: https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-create-index.html#indices-create-api-query-params

The list of available options is retrieved from the corresponding elasticsearch endpoint class to avoid maintaining it here.

### DIFF
--- a/src/Index.php
+++ b/src/Index.php
@@ -402,9 +402,10 @@ class Index implements SearchableInterface
             throw new \TypeError(\sprintf('Argument 2 passed to "%s()" must be of type array|bool|null, %s given.', __METHOD__, \is_object($options) ? \get_class($options) : \gettype($options)));
         }
 
-        $invalidOptions = \array_diff(\array_keys($options), $allowedOptions = [
+        $endpoint = new Create();
+        $invalidOptions = \array_diff(\array_keys($options), $allowedOptions = \array_merge($endpoint->getParamWhitelist(), [
             'recreate',
-        ]);
+        ]));
 
         if (1 === $invalidOptionCount = \count($invalidOptions)) {
             throw new InvalidException(\sprintf('"%s" is not a valid option. Allowed options are "%s".', \implode('", "', $invalidOptions), \implode('", "', $allowedOptions)));
@@ -422,7 +423,9 @@ class Index implements SearchableInterface
             }
         }
 
-        $endpoint = new Create();
+        unset($options['recreate']);
+
+        $endpoint->setParams($options);
         $endpoint->setBody($args);
 
         return $this->requestEndpoint($endpoint);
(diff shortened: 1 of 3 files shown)


src/Index.php
```diff
@@ -422,7 +423,9 @@ public function create(array $args = [], $options = null): Response
             }
         }
 
-        $endpoint = new Create();
+        unset($options['recreate']);
```
ruflin, line 426: Can you share why this is needed now?

deguif: Yes, I had in mind to use option `wait_for_active_shards` when creating indexes on ES.
This would remove the need to have the function `Elastica\Test\Base::_waitForAllocation()` for example by simply passing this option when creating the index.

src/Index.php
```diff
@@ -402,9 +402,10 @@ public function create(array $args = [], $options = null): Response
             throw new \TypeError(\sprintf('Argument 2 passed to "%s()" must be of type array|bool|null, %s given.', __METHOD__, \is_object($options) ? \get_class($options) : \gettype($options)));
         }
 
-        $invalidOptions = \array_diff(\array_keys($options), $allowedOptions = [
+        $endpoint = new Create();
+        $invalidOptions = \array_diff(\array_keys($options), $allowedOptions = \array_merge($endpoint->getParamWhitelist(), [
```
ruflin, line 406: If I understand this correctly, it adds an additional call to Elasticsearch each time before an index is created? Assuming index creation is not happening too often, this is probably not a problem but worth mentioning.

src/Index.php
ruflin: I think I need to read up on my end what this flag does or doesn't. 

Is it now with the new code that `recreate` or was this always the case and now we remove it but didn't before? Trying to understand which part is related to the change of getting the options and what is further improvements.

src/Index.php
deguif: No extra call to Elasticsearch, just the valid options are retrieved from the underlying endpoint class.

src/Index.php
deguif: The unset is here only to remove invalid endpoint options so that these options can be passed to the endpoint via `setParams()`.

src/Index.php
ruflin: Ah, good to know. I somehow assumed it gets it from ES each time.

ruflin (approved):

deguif:
Waiting for #1867 to be merged before merging this one as I will have to update the tests.

tests/IndexTest.php
```diff
@@ -582,16 +582,17 @@ public function testCreate(): void
         $indexName = 'test';
 
         $index = $client->getIndex($indexName);
-        $index->create([]);
-        $this->_waitForAllocation($index);
+        $index->create([], [
+            'wait_for_active_shards' => 'all',
```
deguif, line 586: Here's a concrete example of the use of native options when creating an index.
This allows to replace the use of `$this->_waitForAllocation()` by a native option.

PR #1862, opened 2020-11-03
Remove optional argument before required one

deguif:


src/Transport/HttpAdapter.php
```diff
@@ -30,7 +30,7 @@ class HttpAdapter extends AbstractTransport
     /**
      * Construct transport.
      */
-    public function __construct(?Connection $connection = null, HttpAdapterInterface $httpAdapter)
+    public function __construct(?Connection $connection, HttpAdapterInterface $httpAdapter)
```
ruflin, line 33: This is odd indeed. I wonder if the adapter should be the first param. But I know, it would be a breaking change.

ruflin (approved):

src/Transport/HttpAdapter.php
deguif: Yes, I had the same thought.

PR #1864, opened 2020-11-04
Support previously unsupported elasticsearch library version 7.4.0

deguif:
Now that endpoints classes changes were supported with a BC layer (work of these last days), previously unsupported version 7.4.0 of `elasticsearch` php library is now supported.

See https://github.com/ruflin/Elastica/pull/1715 for history.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added `weighted_avg` to aggregations DSL [#1814](https://github.com/ruflin/Elastica/pull/1814)
 * Supported `guzzlehttp/guzzle` 7.x [#1816](https://github.com/ruflin/Elastica/pull/1816)
 * Supported PHP 8.0 [#1794](https://github.com/ruflin/Elastica/pull/1794)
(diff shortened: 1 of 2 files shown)


ruflin (approved):
Nice!

coreation:
hey @ruflin @deguif thanks so much for keeping this library so alive! I had a question though, which might result in some contributions on my side, the "FunctionScore" query for ES7.4 seems to be broken in terms of syntax.

This is the query Elastica v7.1 is producing when it comes to a Function Score query (shortened)

```
{
	"function_score": {
		"boost_mode": "replace",
		"query": {
			"bool": {}
		},
		"functions": [{
			"script_score": {
				"script": {
					"source": "doc['media.full_body_dense_vector'].size() == 0 ? 0 : 1.0"
				}
			}
		}]
	}
}
```

[This is how the DSL is described in the docs:](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html#function-script-score)

```
{
  "query": {
    "function_score": {
      "query": {
        "match": { "message": "elasticsearch" }
      },
      "script_score": {
        "script": {
          "source": "Math.log(2 + doc['my-int'].value)"
        }
      }
    }
  }
}
```

If I use the FunctionScore, it works ok in 7.3, in 7.4 however, it throws a parse exception:

```
"Unknown key for a START_OBJECT in [function_score]."
```

Looking forward to see if 7.4 is in the works (WIP) and if I have time, happy to help.

PR #1868, opened 2020-11-12
Use microsecond precision instead of second for connection timeout and connect-timeout

romainneutron:
Current implementation only allows second precision for connection timeout and connect-timeout.
This PR provides millisecond precision by accepting float values, and using them if libcurl is compatible with this.

### DIFF
--- a/src/Transport/Http.php
+++ b/src/Transport/Http.php
@@ -76,7 +76,7 @@ class Http extends AbstractTransport
         }
 
         \curl_setopt($conn, \CURLOPT_URL, $baseUri);
-        \curl_setopt($conn, \CURLOPT_TIMEOUT, $connection->getTimeout());
+        \curl_setopt($conn, \CURLOPT_TIMEOUT_MS, $connection->getTimeout() * 1000);
         \curl_setopt($conn, \CURLOPT_FORBID_REUSE, 0);
 
         // Tell ES that we support the compressed responses
@@ -85,9 +85,12 @@ class Http extends AbstractTransport
         \curl_setopt($conn, \CURLOPT_ENCODING, '');
 
         /* @see Connection::setConnectTimeout() */
-        $connectTimeout = $connection->getConnectTimeout();
-        if ($connectTimeout > 0) {
-            \curl_setopt($conn, \CURLOPT_CONNECTTIMEOUT, $connectTimeout);
+        $connectTimeoutMs = $connection->getConnectTimeout() * 1000;
+
+        // Let's only apply this value if the number of ms is greater than or equal to "1".
+        // In case "0" is passed as an argument, the value is reset to its default (300 s)
+        if ($connectTimeoutMs > 1) {
+            \curl_setopt($conn, \CURLOPT_CONNECTTIMEOUT_MS, $connectTimeoutMs);
         }
 
         if (null !== $proxy = $connection->getProxy()) {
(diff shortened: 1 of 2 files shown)


src/Transport/Http.php
```diff
@@ -87,7 +94,16 @@ public function exec(Request $request, array $params): Response
         /* @see Connection::setConnectTimeout() */
         $connectTimeout = $connection->getConnectTimeout();
         if ($connectTimeout > 0) {
-            \curl_setopt($conn, \CURLOPT_CONNECTTIMEOUT, $connectTimeout);
+            // If cURL >= 7.16.2, timeout can be expressed in ms
+            if (\defined('CURLOPT_CONNECTTIMEOUT_MS')) {
+                // since we're accepting float, and the option uses an int, let's round and use a
+                // minimum value of "1" for the option. "0" means "no timeout"
+                \curl_setopt($conn, \CURLOPT_CONNECTTIMEOUT_MS, \max(1, \round($connectTimeout * 1000)));
```
deguif, line 101: Do we need to use `max` here, as there is already a check for only allowing value greater than `0`.

romainneutron: yes, round(0.1) produces `0` whereas `0` as a different meaning

deguif: Ok, so that's for avoiding rounding issue which could happen when using a value like `0.0001`.
Can't we make the code minimalist (for example as guzzle or symfony http client do) by just multiplying by 1000.
```php
\curl_setopt($conn, \CURLOPT_CONNECTTIMEOUT_MS, $connectTimeout * 1000);
```

src/Transport/Http.php
```diff
@@ -87,7 +94,16 @@ public function exec(Request $request, array $params): Response
         /* @see Connection::setConnectTimeout() */
         $connectTimeout = $connection->getConnectTimeout();
         if ($connectTimeout > 0) {
-            \curl_setopt($conn, \CURLOPT_CONNECTTIMEOUT, $connectTimeout);
+            // If cURL >= 7.16.2, timeout can be expressed in ms
+            if (\defined('CURLOPT_CONNECTTIMEOUT_MS')) {
```
deguif, line 98: We can probably remove this and just assume that curl is greater than `7.16.2` which was released a long way back (April 11 2007).

src/Transport/Http.php
romainneutron: This constant is an int, not a float, see `https://curl.se/libcurl/c/CURLOPT_CONNECTTIMEOUT_MS.html`

romainneutron: As a reminder, Symfony HTTP client does not provide interaction with `CURLOPT_CONNECTTIMEOUT_MS`, only for `CURLOPT_TIMEOUT`

As mentioned in the comment of this PR, passing `0` to `CURLOPT_CONNECTTIMEOUT_MS` is a complete different meaning than passing `0` to `CURLOPT_TIMEOUT`. In the latest case, it sets the value to `0`, in the first case it reset to the default connect timeout that equals to `300` in std cURL versions

romainneutron:
I've updated the PR according to comments

src/Transport/Http.php
deguif: Yes but this would be automatically casted to long by using `zval_get_long`, which does under the hood an `zend_dval_to_lval` call when it's a double, so this doesn't require to pass an int exclusively.
See: 
- https://github.com/php/php-src/blob/master/ext/curl/interface.c#L2366

Just to confirm it works:
```php
<?php

$ch = curl_init();
curl_setopt($ch, CURLOPT_URL, "example.com");
curl_setopt($ch, CURLOPT_CONNECTTIMEOUT_MS, 0.1);
curl_setopt($ch, CURLOPT_TIMEOUT_MS, 0.1);
curl_exec($ch);
```

deguif (approved):

deguif:
Thanks @romainneutron.

PR #1870, opened 2020-11-16
Add static_lambda CS rule

deguif:


ruflin (approved):
Interesting that this mostly only affects tests.

PR #1872, opened 2020-11-18
Allow the Terms query to accept arrays of strings, ints and floats

mfn:
As shown in https://github.com/ruflin/Elastica/pull/1765#discussion_r525927129 , ElasticSearch accepts all of them

Note: the change of `public function addTerm` was not previously discussed, but is made for consistency, however poses a breaking change if someone extended this class previously => https://3v4l.org/Ws5bM

### TODO
- [x] Add changelog
- [ ] Discuss breaking change outcome
- [x] Provide a functional test

### DIFF
--- a/src/Query/Terms.php
+++ b/src/Query/Terms.php
@@ -61,6 +61,10 @@ class Terms extends AbstractQuery
      */
     public function addTerm($term): self
     {
+        if (!\is_string($term) && !\is_int($term) && !\is_float($term)) {
+            throw new \TypeError(\sprintf('Argument 1 passed to "%s()" must be of type float|int|string, %s given.', __METHOD__, \is_object($term) ? \get_class($term) : \gettype($term)));
+        }
+
         $this->terms[] = $term;
 
         return $this;
--- a/src/Query/Terms.php
+++ b/src/Query/Terms.php
@@ -20,7 +20,7 @@ class Terms extends AbstractQuery
     private $field;
 
     /**
-     * @var string[]
+     * @var float[]|int[]|string[]
      */
     private $terms;
 
@@ -30,7 +30,7 @@ class Terms extends AbstractQuery
     private $lookup;
 
     /**
-     * @param string[] $terms Terms list, leave empty if building a terms-lookup query
+     * @param float[]|int[]|string[] $terms Terms list, leave empty if building a terms-lookup query
      */
     public function __construct(string $field, array $terms = [])
     {
@@ -45,7 +45,7 @@ class Terms extends AbstractQuery
     /**
      * Sets terms for the query.
      *
-     * @param string[]
+     * @param float[]|int[]|string[]
      */
     public function setTerms(array $terms): self
     {
@@ -56,8 +56,10 @@ class Terms extends AbstractQuery
 
     /**
      * Adds a single term to the list.
+     *
+     * @param float|int|string $term
      */
-    public function addTerm(string $term): self
+    public function addTerm($term): self
     {
         $this->terms[] = $term;
 
--- a/src/Query/Terms.php
+++ b/src/Query/Terms.php
@@ -20,7 +20,7 @@ class Terms extends AbstractQuery
     private $field;
 
     /**
-     * @var float[]|int[]|string[]
+     * @var array<float|int|string>
      */
     private $terms;
 
@@ -30,7 +30,7 @@ class Terms extends AbstractQuery
     private $lookup;
 
     /**
-     * @param float[]|int[]|string[] $terms Terms list, leave empty if building a terms-lookup query
+     * @param array<float|int|string> $terms Terms list, leave empty if building a terms-lookup query
      */
     public function __construct(string $field, array $terms = [])
     {
@@ -45,7 +45,7 @@ class Terms extends AbstractQuery
     /**
      * Sets terms for the query.
      *
-     * @param float[]|int[]|string[]
+     * @param array<float|int|string>
      */
     public function setTerms(array $terms): self
     {
--- a/tests/Query/TermsTest.php
+++ b/tests/Query/TermsTest.php
@@ -96,4 +96,83 @@ class TermsTest extends BaseTest
 
         $this->assertEquals(2, $resultSet->count());
     }
+
+    /**
+     * @group functional
+     */
+    public function testVariousDataTypesViaConstructor(): void
+    {
+        $index = $this->_createIndex();
+
+        $index->addDocuments([
+            new Document(1, ['some_numeric_field' => 9876]),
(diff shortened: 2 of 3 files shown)


ruflin:
Did not realise this could be a breaking change but I would argue this one is acceptable as it is actually a bug fix. @thePanz @deguif WDYT?

@mfn Can you add a changelog entry?

mfn:
> Can you add a changelog entry?

On it, actually i also would like to provide a test, but need a bit time => moving into draft state!

ruflin:
Awesome, thanks!

mfn:
Hmm, a force-push doesn't seem to retrigger the build?

mfn:
Re-opening the PR does 😄

mfn:
@ruflin technically all done / green, only BC break needs be discussed

mfn:
PS: great contributing guide, although I don't have much docker experience, was able to get this easily running and also execute the phpunit tests before pushing, thank you! 👍

deguif:
@mfn @ruflin yes this is a BC break, but impact should be very limited.

I'm also wondering if several types can be combined in the same array like mixing string, int and/or float at the same time.

Maybe this can be something like that if types can be mixed:
```php
array<string|int|float>
```

src/Query/Terms.php
```diff
@@ -56,8 +56,10 @@ public function setTerms(array $terms): self
 
     /**
      * Adds a single term to the list.
+     *
+     * @param float|int|string $term
      */
-    public function addTerm(string $term): self
+    public function addTerm($term): self
     {
         $this->terms[] = $term;
```
deguif, line 64: Would probably be judicious to throw an error when type is not accepted? That would make the type-hint change when supporting PHP 8.0 as minimal version less impacting, as code passing an invalid type would already be breaking.

```suggestion
        if (!\is_string($term) && !\is_int($term) && !\is_float($term)) {
            throw new \TypeError(\sprintf('Argument 1 passed to "%s()" must be of type float|int|string, %s given.', __METHOD__, \is_object($term) ? \get_class($term) : \gettype($term)));
        }
        
        $this->terms[] = $term;
```

mfn:
> `array<string|int|float>`

Should be fine; I didn't see any existing `array<>` syntax in the project so it didn't occur to me. Also wish that would be rather something a code style fix change just do themselves 😅 At least I saw that php-cs did re-order the types I provided 🤷‍♀️

> Would probably be judicious to throw an error when type is not accepted?

🤷‍♀️

The existing code base doesn't seem much to use such a pattern, couldn't find any. So I made the change as un-intrusive as possible

---
I'm not sure if you want me to go ahead and change things, the feedback doesn't seem to much match the existing code base.

I'm happy to change it, just wanted to make sure I don't to unnecessary work, let me know.

Thank you.

ruflin:
I'm personally comfortable with getting this breaking change in. I'm pretty sure we will have 1-2 users hit by it but I personally would even consider it a bug fix. The 1-2 users that will hit by will hopefully have an IDE in place that tells them there is a problem? In the changelog, lets put it under breaking change to make it more obvious. Or do I miss something here?

No strong preference on my end on the way we put it in. I'm following the guidance of @deguif as he put a lot of effort into unifying the code base recently and probably knows best.

I'm also ok with merging the PR as is and then follow up. @deguif WDYT?

thePanz:
@ruflin I do not see this as a breaking change, as the method, is now accepting a wider range of types.

mfn:
>  I'm following the guidance of @deguif as he put a lot of effort into unifying the code base recently and probably knows best.

Roger that! Suggestion applied 😄 

> In the changelog, lets put it under breaking change to make it more obvious.

Also done!

mfn:
@thePanz it's a technical one, see https://3v4l.org/Ws5bM => if someone extended `addTerm` previously => 💥

src/Query/Terms.php
mfn: Applied!

src/Query/Terms.php
```diff
@@ -56,9 +57,15 @@ public function setTerms(array $terms): self
...
     /**
      * Adds a single term to the list.
+     *
+     * @param float|int|string $term
      */
-    public function addTerm(string $term): self
+    public function addTerm($term): self
     {
+        if (!\is_string($term) && !\is_int($term) && !\is_float($term)) {
+            throw new TypeError(\sprintf('Argument 1 passed to "%s()" must be of type float|int|string, %s given.', __METHOD__, \is_object($term) ? \get_class($term) : \gettype($term)));
```
thePanz, line 66: Can we have a `throw new \TypeError` to avoid the `use` of it at the beginning of the class?

mfn: So what's the rule when to import and when not?

thePanz: Not sure, IMHO I would try to not import something from the root namespace, but just use `\TypeError`. WDYT?

thePanz:
> @thePanz it's a technical one, see https://3v4l.org/Ws5bM => if someone extended `addTerm` previously => boom

@mfn you're right! This would open the discussion on making the query classes final, but let's not open the Pandora's vase now :)

tests/Query/TermsTest.php
```diff
@@ -96,4 +97,72 @@ public function testFilteredSearchWithLookup(): void
 
         $this->assertEquals(2, $resultSet->count());
     }
+
+    /**
+     * @group functional
+     */
+    public function testVariousDataTypesViaConstructor(): void
```
thePanz, line 104: Can you add a mixed test too? `Terms('some_numeric_field', ['9876', 1234])`

src/Query/Terms.php
deguif: Yes I'm 👍 with @thePanz comment ;)

tests/Query/TermsTest.php
mfn: Done! => `testVariousMixedDataTypesViaConstructor`

src/Query/Terms.php
mfn: Done!

mfn:
All done!

deguif (approved):

ruflin (approved):

ruflin:
Thanks everyone for getting this in with reviews and @mfn for the quick turn arounds on all the changes!

mfn:
`\o/`

Thank you @ruflin , @deguif and @thePanz !

romainneutron:
I think it should also accept booleans since elasticsearch accepts those

deguif:
@romainneutron I think you're right.
@mfn would you be able to provide another PR to support boolean?

PR #1874, opened 2020-11-23
Deprecate using "interval" configuration for date histograms aggregations

romainneutron:
The combined `interval` field has been deprecated in favor of fixed_interval or calendar_interval. See https://www.elastic.co/guide/en/elasticsearch/reference/7.2/search-aggregations-bucket-datehistogram-aggregation.html

### DIFF
--- a/src/Aggregation/DateHistogram.php
+++ b/src/Aggregation/DateHistogram.php
@@ -7,9 +7,69 @@ namespace Elastica\Aggregation;
  *
  * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-datehistogram-aggregation.html
  */
-class DateHistogram extends Histogram
+class DateHistogram extends AbstractSimpleAggregation
 {
+    use Traits\KeyedTrait;
+    use Traits\MissingTrait;
+
     public const DEFAULT_TIMEZONE_VALUE = 'UTC';
+
+    /**
+     * @param string     $name     the name of this aggregation
+     * @param string     $field    the name of the field on which to perform the aggregation
+     * @param int|string $interval the interval by which documents will be bucketed
+     */
+    public function __construct(string $name, string $field, $interval = null)
+    {
+        parent::__construct($name, $field);
+        $this->setField($field);
+
+        if (null !== $interval) {
+            trigger_deprecation('ruflin/elastica', '7.1.0', 'Argument 3 passed to "%s()" is deprecated, use "setDateInterval()" or "setCalendarInterval()" instead. It will be removed in 8.0.', __METHOD__);
+
+            $this->setParam('interval', $interval);
+        }
+    }
+
+    /**
+     * Set the interval by which documents will be bucketed.
+     *
+     * @deprecated Deprecated since 7.1.0
+     *
+     * @param int|string $interval
+     *
+     * @return $this
+     */
+    public function setInterval($interval): self
+    {
+        trigger_deprecation('ruflin/elastica', '7.1.0', 'The "%s()" method is deprecated, use "setDateInterval()" or "setCalendarInterval()" instead. It will be removed in 8.0.', __METHOD__);
+
+        return $this->setParam('interval', $interval);
+    }
+
+    /**
+     * Set the fixed interval by which documents will be bucketed.
+     *
+     * @param int|string $interval
+     *
+     * @return $this
+     */
+    public function setFixedInterval($interval): self
+    {
+        return $this->setParam('fixed_interval', $interval);
+    }
+
+    /**
+     * Set the calendar interval by which documents will be bucketed.
+     *
+     * @param int|string $interval
+     *
+     * @return $this
+     */
+    public function setCalendarInterval($interval): self
+    {
+        return $this->setParam('calendar_interval', $interval);
+    }
 
     /**
      * Set time_zone option.
@@ -79,5 +139,30 @@ class DateHistogram extends Histogram
         }
 
         return $this->setParam('extended_bounds', $bounds);
+    }
+
+    /**
+     * Set the bucket sort order.
+     *
+     * @param string $order     "_count", "_term", or the name of a sub-aggregation or sub-aggregation response field
+     * @param string $direction "asc" or "desc"
+     *
+     * @return $this
+     */
+    public function setOrder(string $order, string $direction): self
+    {
+        return $this->setParam('order', [$order => $direction]);
+    }
+
+    /**
+     * Set the minimum number of documents which must fall into a bucket in order for the bucket to be returned.
+     *
+     * @param int $count set to 0 to include empty buckets
+     *
+     * @return $this
+     */
+    public function setMinimumDocumentCount(int $count): self
+    {
+        return $this->setParam('min_doc_count', $count);
     }
 }
--- a/src/QueryBuilder/DSL/Aggregation.php
+++ b/src/QueryBuilder/DSL/Aggregation.php
@@ -355,9 +355,9 @@ class Aggregation implements DSL
      *
      * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-histogram-aggregation.html
      *
-     * @param string $name     the name of this aggregation
-     * @param string $field    the name of the field on which to perform the aggregation
-     * @param int    $interval the interval by which documents will be bucketed
+     * @param string     $name     the name of this aggregation
+     * @param string     $field    the name of the field on which to perform the aggregation
+     * @param int|string $interval the interval by which documents will be bucketed
      */
     public function histogram(string $name, string $field, $interval): Histogram
     {
@@ -373,7 +373,7 @@ class Aggregation implements DSL
      * @param string     $field    the name of the field on which to perform the aggregation
      * @param int|string $interval the interval by which documents will be bucketed
      */
-    public function date_histogram(string $name, string $field, $interval): DateHistogram
+    public function date_histogram(string $name, string $field, $interval = null): DateHistogram
     {
         return new DateHistogram($name, $field, $interval);
     }
--- a/tests/Aggregation/BucketSelectorTest.php
+++ b/tests/Aggregation/BucketSelectorTest.php
@@ -46,9 +46,12 @@ class BucketSelectorTest extends BaseAggregationTest
      */
     public function testMaxAggregation(): void
     {
+        $this->_checkVersion('7.2');
+
         $index = $this->_getIndexForTest();
 
-        $dateHistogramAgg = new DateHistogram('histogram_agg', 'date', 'day');
+        $dateHistogramAgg = new DateHistogram('histogram_agg', 'date');
+        $dateHistogramAgg->setFixedInterval('1d');
         $dateHistogramAgg->setFormat('yyyy-MM-dd');
 
         $maxAgg = new Max('max_agg');
--- a/tests/Aggregation/DateHistogramTest.php
+++ b/tests/Aggregation/DateHistogramTest.php
@@ -7,18 +7,86 @@ use Elastica\Document;
 use Elastica\Index;
 use Elastica\Mapping;
 use Elastica\Query;
+use Symfony\Bridge\PhpUnit\ExpectDeprecationTrait;
 
 /**
  * @internal
  */
 class DateHistogramTest extends BaseAggregationTest
 {
+    use ExpectDeprecationTrait;
+
     /**
      * @group functional
      */
     public function testDateHistogramAggregation(): void
     {
-        $agg = new DateHistogram('hist', 'created', '1h');
+        $agg = new DateHistogram('hist', 'created');
+
+        $version = $this->_getVersion();
+
+        if (\version_compare($version, '7.2') < 0) {
+            $agg->setParam('interval', '1h');
+        } else {
+            $agg->setFixedInterval('1h');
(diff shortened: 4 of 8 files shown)


src/QueryBuilder/DSL/Aggregation.php
```diff
@@ -354,9 +354,9 @@ public function ipv4_range(string $name, string $field): IpRange
...
      * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-histogram-aggregation.html
      *
-     * @param string $name     the name of this aggregation
-     * @param string $field    the name of the field on which to perform the aggregation
-     * @param int    $interval the interval by which documents will be bucketed
+     * @param string     $name     the name of this aggregation
+     * @param string     $field    the name of the field on which to perform the aggregation
+     * @param int|string $interval the interval by which documents will be bucketed
      */
     public function histogram(string $name, string $field, $interval): Histogram
```
deguif, line 361:
```suggestion
    public function histogram(string $name, string $field, $interval = null): Histogram
```

To reflect the new constructor default value for `$interval`?
Should probably also be applied to `date_histogram` method in this class?

deguif (approved):
👍

src/QueryBuilder/DSL/Aggregation.php
romainneutron: fixed!

deguif:
Thanks @romainneutron 
Would you be able to add a unit test to ensure deprecation is triggered? See https://github.com/ruflin/Elastica/blob/master/tests/IndexTest.php#L604 for example

src/Aggregation/Histogram.php
```diff
@@ -14,11 +14,16 @@ class Histogram extends AbstractSimpleAggregation
...
      */
-    public function __construct(string $name, string $field, $interval)
+    public function __construct(string $name, string $field, $interval = null)
     {
         parent::__construct($name);
         $this->setField($field);
-        $this->setInterval($interval);
+
+        if (null !== $interval) {
+            trigger_deprecation('ruflin/elastica', '7.1.0', 'The "%s::__construct" third argument is deprecated, use "setDateInterval()" or "setCalendarInterval()" instead. It will be removed in 8.0.', __CLASS__);
```
deguif, line 23:
```suggestion
            trigger_deprecation('ruflin/elastica', '7.1.0', 'Argument 3 passed to "%s()" is deprecated, use "setDateInterval()" or "setCalendarInterval()" instead. It will be removed in 8.0.', __METHOD__);
```

tests/Aggregation/DateHistogramTest.php
```diff
@@ -18,7 +18,62 @@ class DateHistogramTest extends BaseAggregationTest
...
+
+    /**
+     * @group unit
+     * @group legacy
+     */
+    public function testDateHistogramAggregationSetIntervalTriggersADeprecation(): void
+    {
+        $agg = new DateHistogram('hist', 'created');
+
+        $this->expectDeprecation('Since ruflin/elastica 7.1.0: The "setInterval()" method is deprecated, use "setDateInterval()" or "setCalendarInterval()" instead. It will be removed in 8.0.');
```
romainneutron, line 63: This does not work @deguif 
Do you have an idea why?

deguif: You don't call the `setInterval()`method, or am I missing something?
By the way __METHOD__ generate a full namespaced method name.

deguif: Ok tested locally, it misses the inclusion of Symfony provided trait for deprecations.
See https://github.com/ruflin/Elastica/blob/master/tests/IndexTest.php#L27

src/Aggregation/DateHistogram.php
```diff
@@ -7,10 +7,67 @@
  *
  * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-datehistogram-aggregation.html
  */
-class DateHistogram extends Histogram
+class DateHistogram extends AbstractSimpleAggregation
```
deguif, line 10: That would be a BC break.

deguif (changes requested):

src/Aggregation/DateHistogram.php
romainneutron: I thought BC breaks were allowed

deguif: Sometimes but should not ;)
I think in this case, we can avoid it or I'm missing something?

romainneutron: so, what should we do?

tests/Aggregation/DateHistogramTest.php
```diff
@@ -18,7 +18,62 @@ class DateHistogramTest extends BaseAggregationTest
      */
     public function testDateHistogramAggregation(): void
     {
-        $agg = new DateHistogram('hist', 'created', '1h');
+        $this->_checkVersion('7.2');
+
+        $agg = new DateHistogram('hist', 'created');
+        $agg->setFixedInterval('1h');
```
deguif, line 24:
```suggestion
        $version = $this->_getVersion();

        if (\version_compare($version, '7.2') < 0) {
            $this->expectDeprecation('Since ruflin/elastica 7.1.0: The "Elastica\Aggregation\DateHistogram::setInterval()" method is deprecated, use "setDateInterval()" or "setCalendarInterval()" instead. It will be removed in 8.0.');

            $agg = (new DateHistogram('hist', 'created'))
                ->setInterval('1h')
            ;
        } else {
            $agg = (new DateHistogram('hist', 'created'))
                ->setFixedInterval('1h')
            ;
        }
```

Cannot we use a code like this to test on version lower than 7.2?

deguif:
```suggestion
        $version = $this->_getVersion();

        if (\version_compare($version, '7.2') < 0) {
            $agg = (new DateHistogram('hist', 'created'))
                ->setParam('interval', '1h')
            ;
        } else {
            $agg = (new DateHistogram('hist', 'created'))
                ->setFixedInterval('1h')
            ;
        }
```

Or like this?

deguif: Maybe the deprecation is a little too violent, as this can't be fixed for versions lower than 7.2 unless using `setParam()` directly? WDYT @ruflin @thePanz ? Is there any advices/hints for this kind of deprecations?

Not sure if it's a blocker as the deprecation is also very helpful for upper compatibility.

src/Aggregation/DateHistogram.php
deguif: Wouldn't this be helpful (https://github.com/ruflin/Elastica/pull/1874#discussion_r528717001) to avoid changing the inheritance?

deguif: Ok I think I got the point.
DateHistogram `interval` parameter is deprecated but still useful for `Histogram`.
Can you confirm it's what you're addressing by changing the inheritance chain?

romainneutron: yes. We could move methods if we want to keep those implementation. 
Maintaining BC on interfaces and userland API yes, but should we maintain BC on abstract internal classes?

romainneutron:
PR udpated.
I dont think we should consider still consider using Histogram as base for DateHistogram. They're both histogram but differ on some of their behavior.

ruflin:
If understand it there are 2 things here:

* Changing the inheritance of DateHistogram
* Deprecating interval

We need the first one to only deprecate interval for DateHistogram?

It seems the current code change is backward compatible expect in the case someone checked for specifically inherited classes? Is there more the could break users code?

romainneutron:
PR rebased

ruflin:
Seems like one of the tests fail because of the deprecation :-) Can you fix it and also rebased the changelog?

romainneutron:
PR rebased, tests are green :)

ruflin:
I'll refer to @deguif on the final 👍  as he had the most interactions here.

romainneutron:
any news?

ruflin:
@deguif Any chance to have a quick loo?

tests/Aggregation/BucketSelectorTest.php
```diff
@@ -46,9 +46,12 @@ public function testToArray(): void
      */
     public function testMaxAggregation(): void
     {
+        $this->_checkVersion('7.2');
```
deguif, line 49: Cannot we do the same you've done later for `DateHistogramTest::testDateHistogramAggregation()`, I mean checking the version of elasticsearch and using `setInterval()` or `setFixedInterval()` depending the version?

tests/Aggregation/DerivativeTest.php
```diff
@@ -45,9 +45,12 @@ public function testToArray(): void
      */
     public function testMaxAggregation(): void
     {
+        $this->_checkVersion('7.2');
```
deguif, line 48: Same here:
> Cannot we do the same you've done later for DateHistogramTest::testDateHistogramAggregation(), I mean checking the version of elasticsearch and using setInterval() or setFixedInterval() depending the version?

tests/Aggregation/SerialDiffTest.php
```diff
@@ -20,7 +20,10 @@ class SerialDiffTest extends BaseAggregationTest
      */
     public function testSerialDiffAggregation(): void
     {
-        $dateHistogramAggregation = new DateHistogram('measurements', 'measured_at', 'hour');
+        $this->_checkVersion('7.2');
```
deguif, line 23: Same here:
> Cannot we do the same you've done later for DateHistogramTest::testDateHistogramAggregation(), I mean checking the version of elasticsearch and using setInterval() or setFixedInterval() depending the version?

src/Aggregation/DateHistogram.php
```diff
@@ -80,4 +140,29 @@ public function setExtendedBounds(string $min = '', string $max = ''): self
...
+
+    /**
+     * Set the bucket sort order.
+     *
+     * @param string $order     "_count", "_term", or the name of a sub-aggregation or sub-aggregation response field
+     * @param string $direction "asc" or "desc"
+     *
+     * @return $this
+     */
+    public function setOrder(string $order, string $direction): self
```
deguif, line 152: It seems it can also be a multi-dimensional array?
Like this: 
```
"order": [ { "rock>playback_stats.avg": "desc" }, { "_count": "desc" } ]
```

src/Aggregation/DateHistogram.php
```diff
@@ -7,10 +7,70 @@
...
+     *
+     * @deprecated Deprecated since 7.1.0
+     *
+     * @param int|string $interval
+     *
+     * @return $this
+     */
+    public function setInterval($interval): self
+    {
+        trigger_deprecation('ruflin/elastica', '7.1.0', 'The "%s()" method is deprecated, use "setDateInterval()" or "setCalendarInterval()" instead. It will be removed in 8.0.', __METHOD__);
```
deguif, line 45: I'm dubious about this deprecation, as `interval` parameter is still legit for ES version lower than `7.2`. It was only deprecated in 7.2 version but is still supported (until at least 8.0).

I think we should remove the deprecation here. But I would let the deprecation in the constructor so that users would be required to update their code to explicitly call `setInterval()`, `setFixedInterval()` or `setCalendarInterval()`.

Later when bumping minimum version of ES to a version later than 7.2 (probably when bumping to 8.0) we should be able to drop this method and only let the two others.

Would that make sense? ping @ruflin @thePanz

ruflin: There are two different deprecations:

* Deprecation in Elasticsearch
* Deprecation in Elastica

In the best case, they are aligned. As Elastica lags behind in releases, we kind of have the option to deprecate things already in older releases, assuming there is already an option available which is not deprecated.

To remove a method in 8.0, we should have it deprecated already in 7.x. I would say the earlier we can tell devs to adjust their code the better.

deguif: Yes but that means in this case, that users of version older than 7.2 will get this deprecation but won't be able to remove the deprecation as the parameters (`setFixedInterval`, `setCalendarInterval`) can't work with their ES version. But maybe that's not an issue.

ruflin: I see. Maybe we can leave in a note in the docs for this. In general it would be great if we can push users to use the newest version of Elastica and the related version of Elasticsearch and would optimise for this. Maybe we could even adjust the README that this version of Elastica is supported for Elasticsearch >= 7.3 or similar. It does not mean, it will not work anymore with the older versions.

tests/Aggregation/DateHistogramTest.php
thePanz: I like having deprecations, and the first example (where we test the deprecation) if :+1:  for me :)

src/Aggregation/DateHistogram.php
thePanz: I would add the PHP Doc to the methods, stating that those are not supported by ES >= 8.x
When the library will support ES v8.x only, we can remove those methods.
WDYT?

ruflin: No strong preference on my end. SGTM

VincentLanglet:
This https://github.com/ruflin/Elastica/pull/2099 could be a simpler way to do it, isn't it ?

thePanz:
Fixed in #2099

PR #1877, opened 2020-11-23
Fix Composite::addAfter(null) aggregation behavior

romainneutron:
`Elastica\Aggregation\Composite::addAfter` accepts `null` value. However, it's not allowed by elasticsearch It returns an error: `[composite] after doesn't support values of type: VALUE_NULL`

This PR fixes this bug

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Replaced `_routing` and `_retry_on_conflict` by `routing` and `retry_on_conflict` in `AbstractUpdateAction` [#1807](https://github.com/ruflin/Elastica/issues/1807)
 * Fixed `Elastica\Util::toSnakeCase()` with first letter being lower cased [#1831](https://github.com/ruflin/Elastica/pull/1831)
 * Fixed handling precision as string in `Elastica\Aggregation\GeohashGrid::setPrecision()` [#1884](https://github.com/ruflin/Elastica/pull/1884)
+* Fixed calling `Elastica\Aggregation\Composite::addAfter()` with the `null` value [1877](https://github.com/ruflin/Elastica/pull/1877)
 ### Security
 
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -52,6 +52,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fixed type-hint for `Elastica\QueryBuilder\DSL\Aggregation::sampler()` not consistent with the underlying constructor call [#1815](https://github.com/ruflin/Elastica/pull/1815)
 * Replaced `_routing` and `_retry_on_conflict` by `routing` and `retry_on_conflict` in `AbstractUpdateAction` [#1807](https://github.com/ruflin/Elastica/issues/1807)
 * Fixed `Elastica\Util::toSnakeCase()` with first letter being lower cased [#1831](https://github.com/ruflin/Elastica/pull/1831)
+* Fix calling `Elastica\Aggregation\Composite::addAfter()` with the `null` value [1877](https://github.com/ruflin/Elastica/pull/1877)
```
deguif, line 55:
```suggestion
* Fixed calling `Elastica\Aggregation\Composite::addAfter()` with the `null` value [1877](https://github.com/ruflin/Elastica/pull/1877)
```

src/Aggregation/Composite.php
```diff
@@ -25,6 +25,10 @@ public function addSource(AbstractAggregation $aggregation): self
      */
     public function addAfter(?array $checkpoint): self
```
deguif, line 26: Cannot we deprecate passing null? So that we can change the type-hint in 8.0 to only allow array and remove the check for null?

romainneutron: I'd be in favor of keeping it, It's very handy for writing code:

```
do {
    $result = $this->doYourAggregation($after ?? null);

    // ...

    $after = $result->getAggregations()['composite']['after_key'];
} while (true);
```

romainneutron:
PR rebased

ruflin (approved):
LGTM

@deguif WDYT?

deguif (approved):

PR #1878, opened 2020-11-27
Use github action for coding style checks

deguif:
See result here: https://github.com/deguif/Elastica/runs/1462981807
Should be accessible here once merged.

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -1,0 +1,20 @@
+name: 'Continuous integration'
+on: ['push', 'pull_request']
+jobs:
+  cs:
+    runs-on: 'ubuntu-latest'
+    name: 'Coding style'
+    steps:
+      - name: 'Checkout'
+        uses: 'actions/checkout@v2'
+
+      - name: 'Setup PHP'
+        uses: 'shivammathur/setup-php@v2'
+        with:
+          php-version: '7.2'
+          coverage: 'none'
+          extensions: 'json, mbstring, tokenizer'
+          tools: 'php-cs-fixer:2.16.4'
+
+      - name: 'Check PHP code'
+        run: 'php-cs-fixer fix --diff --dry-run --allow-risky=yes --using-cache=false'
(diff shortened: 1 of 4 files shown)


ruflin:
Nice! This looks great. I would suggest we add a changelog entry even if though this is not related to the actual code. Still it is a big change to how the project does CI.

WDYT about for now keep the travis part in an only remove it when we see it sucessfully working on 2-3 PR's?

.github/workflows/continuous-integration.yaml
```diff
@@ -0,0 +1,24 @@
...
+          coverage: 'none'
+          extensions: 'json, mbstring, tokenizer'
+          tools: 'php-cs-fixer:2.16.4'
+
+      - name: 'Display tools versions'
+        run: |
+          php-cs-fixer --version
+
+      - name: 'Check PHP code'
+        run: 'php-cs-fixer fix --diff --dry-run --allow-risky=yes --using-cache=false'
```
ruflin, line 24: Is it possible to use `make run-phpcs` instead here? Like this we have only 1 place where we specify the exact params to be run like we have for travis. So local and CI are always in sync.

deguif: I would be in favor to let the `setup-php` github action setup this, that's used on many PHP projects on github.
Advantage: No probem with extra downloads done by the makefile to install the tools (shipped directly in the GA)

Downside is that we will have to maintain two versions as you stated, the local environment (via makefile) for contributors and the CI one, but I think we can live with it. The CI one will be the source of trust as it would make the pull request / merge checks fail.

Would that be a blocker for you?

Regarding keeping the travis part, I can let the old part but I'm familiar with this setup as I already configured this way on many projects and it works really well. See for example:
- https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/.github/workflows/continuous-integration.yaml
- https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/blob/master/.github/workflows/continuous-integration.yaml
- ...

ruflin:
@deguif Not blocker at all. I just merged it. I'm much more keen to get CI running again then anything else! Thanks for taking this on.

Issue #1824, opened 2020-10-21
Aggregation/Query/Processor class namings standardisation


deguif:
Previously some class were required to be named with a suffix to avoid conflicting with some reserved keywords.
For examples these classes use different naming than others:
- `Elastica\Query\BoolQuery` instead of `Elastica\Query\Bool` previously
- `Elastica\Query\MatchQuery` instead of `Elastica\Query\Match` previously
- `Elastica\Aggregation\GlobalAggregation`
- `Elastica\Aggregation\ParentAggregation`
- `Elastica\Processor\ForeachProcessor` in PR #1813 

I think it would be nice to standardise the suffix strategy by suffixing all classes from `Aggregation`, `Query` and `Processor` namespaces. It could be done in a BC manner by creating a new class with the suffix part and making the previous class extends from it (which would trigger a deprecation too to help fixing it).

What do you think?

ruflin:
I'm not a big fan of having the directory name repeated also in the class name itself. My hope initially was that this would be the exception, but it seems we have more and more use cases around it.

Is there any specific issue that triggered this or more generally to unify things?

PR #1880, opened 2020-12-01
added geotile grid support

krewetka:
In ElasticSearch 7 and above there is new aggregation type which is not supported here yet -> [GeoTile Grid](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-geotilegrid-aggregation.html
)

I have added support of it. It is very similar to existing [GeoHash Grid]( https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-geohashgrid-aggregation.html ) except it has more detailed precision levels.

I also renamed one test in other class which had incorrect name.

### DIFF
--- a/src/Aggregation/GeotileGrid.php
+++ b/src/Aggregation/GeotileGrid.php
@@ -1,0 +1,70 @@
+<?php
+
+namespace Elastica\Aggregation;
+
+/**
+ * Class GeotileGrid.
+ *
+ * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-geotilegrid-aggregation.html
+ */
+class GeotileGrid extends AbstractAggregation
+{
+    public const DEFAULT_PRECISION_VALUE = 5;
+    public const DEFAULT_SIZE_VALUE = 10000;
+
+    /**
+     * @param string $name  the name of this aggregation
+     * @param string $field the field on which to perform this aggregation
+     */
+    public function __construct(string $name, string $field)
+    {
+        parent::__construct($name);
+        $this->setField($field);
+    }
+
+    /**
+     * Set the field for this aggregation.
+     *
+     * @param string $field the name of the document field on which to perform this aggregation
+     *
+     * @return $this
+     */
+    public function setField(string $field): self
+    {
+        return $this->setParam('field', $field);
+    }
+
+    /**
+     * Set the precision for this aggregation.
+     *
+     * @param int $precision an integer between 1 and 12, inclusive. Defaults to 5.
+     *
+     * @return $this
+     */
+    public function setPrecision(int $precision): self
+    {
+        return $this->setParam('precision', $precision);
+    }
+
+    /**
+     * Set the maximum number of buckets to return.
+     *
+     * @param int $size defaults to 10,000
+     *
+     * @return $this
+     */
+    public function setSize(int $size): self
+    {
+        return $this->setParam('size', $size);
+    }
+
+    /**
+     * Set the number of results returned from each shard.
+     *
+     * @return $this
+     */
+    public function setShardSize(int $shardSize): self
+    {
+        return $this->setParam('shard_size', $shardSize);
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added `Elastica\Aggregation\GeoDistance::setKeyed()` [#1876](https://github.com/ruflin/Elastica/pull/1876)
 * Added `Elastica\Aggregation\Histogram::setKeyed()` [#1876](https://github.com/ruflin/Elastica/pull/1876)
 * Added `Elastica\Aggregation\IpRange::setKeyed()` [#1876](https://github.com/ruflin/Elastica/pull/1876)
-* Added `Elastica\Aggregation\GeotileGrid` [#1880](https://github.com/ruflin/Elastica/pull/1880)
+* Added `Elastica\Aggregation\GeotileGridAggregation` [#1880](https://github.com/ruflin/Elastica/pull/1880)
 ### Changed
 * Allow `string` such as `wait_for` to be passed to `AbstractUpdateAction::setRefresh` [#1791](https://github.com/ruflin/Elastica/pull/1791)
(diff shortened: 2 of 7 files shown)


ruflin (commented):
Hi, this looks good. Could you add a changelog?

There is an other issue we have at the moment with travis and CI is not run. I would like to wait with getting this PR in until we have CI running again if that is ok for you?

krewetka:
> Hi, this looks good. Could you add a changelog?

sure I will add it there :) 

> I would like to wait with getting this PR in until we have CI running again if that is ok for you?

if this will be a week or so sure - I switched my project to use fork right now

ruflin:
@krewetka As soon as https://github.com/ruflin/Elastica/pull/1885 is merged, could you rebase your PR?

deguif:
It's merged

src/Aggregation/GeotileGrid.php
```diff
@@ -0,0 +1,70 @@
+<?php
+
+namespace Elastica\Aggregation;
+
+/**
+ * Class GeotileGrid.
+ *
+ * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-geotilegrid-aggregation.html
+ */
+class GeotileGrid extends AbstractAggregation
```
deguif, line 10: We should probably start marking these kind of class as `final` as @thePanz recently commented.
And as it's a new class also suffix the its name with `Aggregation`. This requires #1887 to be merged before.

src/Aggregation/GeotileGrid.php
```diff
@@ -0,0 +1,70 @@
...
+    {
+        return $this->setParam('size', $size);
+    }
+
+    /**
+     * Set the number of results returned from each shard.
+     *
+     * @return $this
+     */
+    public function setShardSize(int $shardSize): self
```
deguif, line 66: Now that #1889 was merged, you can use the trait to have this behavior.

src/Aggregation/GeotileGrid.php
krewetka: ok, no problem

src/Aggregation/GeotileGrid.php
krewetka: ok

src/Aggregation/GeotileGrid.php
deguif: ping @ruflin for his confirmation.

krewetka: @ruflin  is this ok for you? prefer confirming before doing this changes

ruflin: @deguif can you point me to the final comment? I'm not sure I fully follow why we should make it final? Should we really prevent others from extending these classes?

deguif: Let's not mark the class as final for now as it would need discussing this subject more in details.
Will the name `GeotileGridAggregation` be OK for you?

deguif:
Thanks @krewetka for your PR.

I see on the elasticsearch documentation that there is also an optional `bounds` parameter, it can probably be added without too much work. WDYT?

For details: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-geotilegrid-aggregation.html#_requests_with_additional_bounding_box_filtering_2

krewetka:
> I see on the elasticsearch documentation that there is also an optional `bounds` parameter, it can probably be added without too much work. WDYT?

Yes, it could. But it also exists in geohash grid and is not there in Ruflin elastica so maybe better as separate PR adding it to both? :thinking:

src/Aggregation/GeotileGrid.php
ruflin: @deguif Currently we always skip the Aggregation part as it is already in the name of the directory. So to be consistent it should be `GoetileGrid`?

ruflin:
@krewetka Getting an additional PR later on SGTM.

krewetka:
> @krewetka Getting an additional PR later on SGTM.

ok :) so this one should be ready IMO to be merged @ruflin , @deguif

src/Aggregation/GeotileGrid.php
deguif: Yes but as stated in #1824 I would be in favor to deprecate all aggregations classes and suffix them with `Aggregation` to avoid them colliding with reserved keywords and unifying all classes names. 

Here's the class is new so it could directly be named accordingly and avoid having to deprecate it later.
Would that sound good for you?

krewetka: @ruflin  @deguif please align about it. I really would like it to be merged soon - one way or another ;) 

I have added this `Aggregation` at the end per request here. I have also changed code in my project to support this new name already.

Btw, I personally like adding Aggregation to the name.  Easlier my projects code looked like this  

`use Elastica\Aggregation\GeotileGrid as GeotileGridAggregation;`

for readibility reason later in code. Otherwise it is much less visible in code that this is aggregation.

krewetka:
> Thanks @krewetka for your PR.
> 
> I see on the elasticsearch documentation that there is also an optional `bounds` parameter, it can probably be added without too much work. WDYT?
> 
> For details: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-geotilegrid-aggregation.html#_requests_with_additional_bounding_box_filtering_2

@deguif after checking this parameter and info that is supports alll formats from https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-geo-bounding-box-query.html#query-dsl-geo-bounding-box-query-accepted-formats I am not sure I am able to handle this one propely.

I have checked https://github.com/ruflin/Elastica/blob/a2f5b0d7171691da06a4c31b6829b9c1dafdfbf5/src/Query/GeoBoundingBox.php and even for original one there is just one format supported now in ruflin/Eastica 

I can add the same one type for both Geohash and Geotile but to be honest I don't have time for more, not the knowledge how to do it properly :)

deguif:
@krewetka no problem for that, that's not a blocker for me.
Thank you for this nice contribution and as @ruflin already commented we can add this parameter later ;)

ruflin (approved):
LGTM. Did a commit to resolve the changelog conflict. @deguif If ok for you, please merge directly.

deguif (approved):

krewetka:
Thanks for merging :)

PR #1882, opened 2020-12-02
Run unit tests on github action

deguif:
See result here: https://github.com/deguif/Elastica/actions/runs/395847967 or at the end of the PR in the checks section.

I had to make several changes in tests to fix unit tests that were functional tests, here's the result without the changes on test classes: https://github.com/deguif/Elastica/runs/1480767481

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -2,7 +2,7 @@ name: 'Continuous integration'
 on: ['push', 'pull_request']
 jobs:
   cs:
-    runs-on: 'ubuntu-latest'
+    runs-on: 'ubuntu-20.04'
     name: 'Coding style'
     steps:
       - name: 'Checkout'
@@ -23,7 +23,7 @@ jobs:
         run: 'composer-normalize --diff --dry-run --indent-size=4 --indent-style=space --no-update-lock'
 
   phpunit:
-    runs-on: 'ubuntu-latest'
+    runs-on: 'ubuntu-20.04'
     name: 'PHPUnit (PHP ${{ matrix.php }}, ES ${{ matrix.elasticsearch }})'
     timeout-minutes: 5
     strategy:
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -21,3 +21,53 @@ jobs:
 
       - name: 'Check composer.json'
         run: 'composer-normalize --diff --dry-run --indent-size=4 --indent-style=space --no-update-lock'
+
+  phpunit:
+    runs-on: 'ubuntu-latest'
+    name: 'PHPUnit (PHP ${{ matrix.php }}, ES ${{ matrix.elasticsearch }})'
+    timeout-minutes: 5
+    strategy:
+      matrix:
+        include:
+          - php: '7.2'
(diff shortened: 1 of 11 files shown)


.github/workflows/continuous-integration.yaml
```diff
@@ -18,3 +18,53 @@ jobs:
...
+          key: '${{ runner.os }}-composer-php${{ matrix.php }}-${{ hashFiles(''**/composer.json'') }}'
+          restore-keys: |
+            ${{ runner.os }}-composer-php${{ matrix.php }}-
+            ${{ runner.os }}-composer-
+
+      - name: 'Update dependencies'
+        run: 'composer update --prefer-dist --no-interaction --no-progress --ansi ${{ matrix.composer_flags }}'
+
+      - name: 'Run tests'
+        run: 'vendor/bin/phpunit --group unit'
```
deguif, line 70: Just running unit test for the moment as booting a functional stack is more complicated. Will be done in a following PR if you don't mind.

tests/Query/TermsTest.php
```diff
@@ -9,10 +9,12 @@
 
 /**
  * @internal
- * @group unit
```
deguif, line 12: Adding this annotation at the class level, mark all methods as unit test.

deguif:
It's blazing fast (without the functional tests part).
Screenshot 2020-12-02 at 12 02 51

tests/IndexTest.php
```diff
@@ -626,17 +688,19 @@ public function testCreateWithInvalidOption(): void
...
-        $index = $client->getIndex($indexName);
+        $client = $this->createMock(Client::class);
+        $index = new Index($client, 'test');
 
         $opts = ['testing_invalid_option' => true];
         $index->create([], $opts);
     }
 
+    /**
+     * @group unit
```
ruflin, line 699: unit?

tests/PipelineTest.php
```diff
@@ -24,7 +25,8 @@ public function testProcessor(): void
         $rename = new Rename('foo', 'target.field');
         $set = new Set('field4', 324);
 
-        $processors = new Pipeline($this->_getClient());
+        $client = $this->createMock(Client::class);
```
ruflin, line 28: Can you share some background on this change?

tests/Query/ParentIdTest.php
```diff
@@ -124,7 +124,7 @@ public function testParentId(): void
     }
 
     /**
-     * @group unit
+     * @group functional
```
ruflin, line 127: 🤦 Nice catch.

ruflin (approved):
Awesome. Left one question but should not block merging this PR! Please get it in.

tests/IndexTest.php
deguif: Yes unit as there is no call done on the elasticsearch stack. Mocking a client is sufficient as it's not used for this test

tests/PipelineTest.php
deguif: Yes a client mock is sufficient here to satisfy the pipeline object constructor.
No call is made on the elasticsearch server, as only the `toArray()` method is tested here.

deguif:
I just added a changelog.

PR #1884, opened 2020-12-02
added distance precision to GeohashGrid.php

krewetka:
As described in https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-geohashgrid-aggregation.html#_options_3 there is possibility to pass precision as string.

It was also working in earlier versions of Elastica.

Got broken after adding `int` to parameter in `setPrecision`.

I think it is better to add second method for that.

### DIFF
--- a/src/Aggregation/GeohashGrid.php
+++ b/src/Aggregation/GeohashGrid.php
@@ -39,24 +39,16 @@ class GeohashGrid extends AbstractAggregation
     /**
      * Set the precision for this aggregation.
      *
-     * @param int $precision an integer between 1 and 12, inclusive. Defaults to 5.
+     * @param int|string $precision an integer between 1 and 12, inclusive. Defaults to 5 or distance like 1km, 10m
      *
      * @return $this
      */
-    public function setPrecision(int $precision): self
+    public function setPrecision($precision): self
     {
-        return $this->setParam('precision', $precision);
-    }
+        if (!\is_int($precision) && !\is_string($precision)) {
+            throw new \TypeError(\sprintf('Argument 1 passed to "%s()" must be a int or string, %s given.', __METHOD__, \is_object($precision) ? \get_class($precision) : \gettype($precision)));
+        }
 
-    /**
-     * Set the precision for this aggregation in meters or kilometers.
-     *
-     * @param string $precision a string like 100m or 1km
-     *
-     * @return $this
-     */
-    public function setDistancePrecision(string $precision): self
-    {
         return $this->setParam('precision', $precision);
     }
 
--- a/src/Aggregation/GeohashGrid.php
+++ b/src/Aggregation/GeohashGrid.php
@@ -46,7 +46,7 @@ class GeohashGrid extends AbstractAggregation
     public function setPrecision($precision): self
     {
         if (!\is_int($precision) && !\is_string($precision)) {
-            throw new \TypeError(\sprintf('Argument 1 passed to "%s()" must be a int or string, %s given.', __METHOD__, \is_object($precision) ? \get_class($precision) : \gettype($precision)));
+            throw new \TypeError(\sprintf('Argument 1 passed to "%s()" must be of type int|string, %s given.', __METHOD__, \is_object($precision) ? \get_class($precision) : \gettype($precision)));
         }
 
         return $this->setParam('precision', $precision);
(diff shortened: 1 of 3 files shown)


ruflin:
@krewetka Could you also rebase this PR now that CI changes are in?

src/Aggregation/GeohashGrid.php
```diff
@@ -46,6 +46,18 @@ public function setPrecision(int $precision): self
...
     }
 
+    /**
+     * Set the precision for this aggregation in meters or kilometers.
+     *
+     * @param string $precision a string like 100m or 1km
+     *
+     * @return $this
+     */
+    public function setDistancePrecision(string $precision): self
```
deguif, line 56: I would be in favor to use the previous setter and widen its type.
To simplify the migration path to multi arguments types available with PHP 8.0 I would add a check to ensure it's an int or a string in the setter and throw a \TypeError if the type is anything else than an int or string.

See for examples: https://github.com/ruflin/Elastica/blob/master/src/Query/Terms.php#L65 or https://github.com/ruflin/Elastica/blob/master/src/Index.php#L402

Changing the type of the method is a BC break, but in this case it's also a bug fix, so adding a changelog for documenting this BC should be fine. WDYT @ruflin ?

krewetka: that was my concern that it is BC and also bugfix 

I can make it one method  and check inside it if argument is int or string

ruflin: ++ on treating it as a bug fix and use the previous one.

krewetka: ok, changed

krewetka:
> @krewetka Could you also rebase this PR now that CI changes are in?

just so you know CI is failing sometimes for some strange reason. I had a fail, change just single test name a bit , restarted it and next time it was ok

image

tests/Aggregation/GeohashGridTest.php
```diff
@@ -29,6 +29,37 @@ public function testGeohashGridAggregation(): void
...
+        $query = new Query();
+        $query->addAggregation($agg);
+        $results = $this->_getIndexForTest()->search($query)->getAggregation('hash');
+
+        $this->assertEquals(2, $results['buckets'][0]['doc_count']);
+        $this->assertEquals(1, $results['buckets'][1]['doc_count']);
+    }
+
+    /**
+     * @group functional
```
deguif, line 49: That could a simple unit test here.

tests/Aggregation/GeohashGridTest.php
```diff
@@ -29,6 +29,37 @@ public function testGeohashGridAggregation(): void
...
+    public function testGeohashGridAggregationWithNotAllowedPrecision(): void
+    {
+        $this->expectException(\TypeError::class);
+
+        $agg = new GeohashGrid('hash', 'location');
+        $agg->setPrecision(1.5);
+
+        $query = new Query();
+        $query->addAggregation($agg);
+        $this->_createIndex()->search($query)->getAggregation('hash');
```
deguif, line 60: Not needed as exception should already be thrown.

src/Aggregation/GeohashGrid.php
```diff
@@ -39,12 +39,16 @@ public function setField(string $field): self
...
-     * @param int $precision an integer between 1 and 12, inclusive. Defaults to 5.
+     * @param int|string $precision an integer between 1 and 12, inclusive. Defaults to 5 or distance like 1km, 10m
      *
      * @return $this
      */
-    public function setPrecision(int $precision): self
+    public function setPrecision($precision): self
     {
+        if (!\is_int($precision) && !\is_string($precision)) {
+            throw new \TypeError(\sprintf('Argument 1 passed to "%s()" must be a int or string, %s given.', __METHOD__, \is_object($precision) ? \get_class($precision) : \gettype($precision)));
```
deguif, line 49:
```suggestion
            throw new \TypeError(\sprintf('Argument 1 passed to "%s()" must be of type int|string, %s given.', __METHOD__, \is_object($precision) ? \get_class($precision) : \gettype($precision)));
```

tests/Aggregation/GeohashGridTest.php
krewetka: yeap, you are right, will change it

ruflin (approved):
Change LGTM, but waiting on green. Also @deguif should have a final look. Please merge directly if it looks good to you.

src/Aggregation/GeohashGrid.php
krewetka: done :heavy_check_mark:

deguif:
Thanks @krewetka

PR #1888, opened 2020-12-03
Use dedicated namespace for aggregation traits

deguif:
This follow the recent addition of traits by @romainneutron for aggregations (first PR merged, second is awaiting a rebase).
As the usage of traits should increase in the future, let's use a dedicated namespace for them.


ruflin (approved):
LGTM. See my comment on the other PR if we then still need the postfix.

PR #1889, opened 2020-12-03
Add ShardSizeTrait

deguif:


src/Aggregation/Traits/ShardSizeTrait.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Elastica\Aggregation\Traits;
+
+trait ShardSizeTrait
```
ruflin, line 5: Nit: Should we call it `ShardSize` or `ShardSizeTrait`? That it is a Trait is kind of already given by the namespace. Are there some conventions around this from other projects?

ruflin (approved):

src/Aggregation/Traits/ShardSizeTrait.php
deguif: Yes for example Symfony suffix them with `Trait`.
For exemple: https://github.com/symfony/symfony/tree/5.x/src/Symfony/Component/Cache/Traits

ruflin: 👍

PR #1893, opened 2020-12-04
Deprecate unsuffixed processor classes

deguif:


ruflin:
Can you share some background on why you want to suffix them all? Is the plan to suffix all the other assets too like Aggregations (see the other discussion).

deguif:
Yes, see my comment: https://github.com/ruflin/Elastica/pull/1880#discussion_r537397993
I'm starting by processors as that should impact less library users (deprecations triggering) and that would let us see how it's received by them.
The plan I had in mind is to apply to all aggregations / queries classes later (for 7.2.0 for example).
With that, we should be less impacted by new reserved keywords introduced in newer PHP versions and have the same naming strategy for all these classes.

ruflin:
I must confess I'm torn on this. I definitively don't like the conflicts but at the same time it seems extreme for 2-3 conflicts to rename ALL classes and with it basically force all users to modify their app at one stage. Instead I would probably more have a default fallback way that we use for the conflicts but keep as is.

If I understand this change right and we follow through with it, it will affect the majority of classes in the library?

deguif:
Yes at least, for the main ones (and most subject to colliding with reserved keywords) in the `Query`, `Aggregation` and `Processor` namespaces.
It's done in a backward compatible way so should have no impact for library users (just deprecations).

There's also the comment of @krewetka regarding class name and their readibility that I 👍 : https://github.com/ruflin/Elastica/pull/1880#discussion_r537471039

But that's not a blocker, the goal here is to clean / unify things, this can be ignored (and this PR closed) if we're not agreeing on this point ;)

ruflin:
@thePanz Would be great to get one more opinion on this, could you chime in?

thePanz:
@ruflin  @deguif I like the approach, each class name is now "self-explanatory"... the most relevant example is the "Date" processor, being renamed to `DateProcessor` makes things more clean IMHO

Just a niptick: can we remove `@return $this` when as the method's signature is returning `self`?

ruflin (approved):
@deguif @thePanz Ok, lets move forward with this.

deguif:
> Just a niptick: can we remove @return $this when as the method's signature is returning self?

We can but in case of inheritance this allow to properly infer the right type. This should be solved with `static` type-hint with PHP 8.0 ;)

deguif:
May I merge this one ?

ruflin:
@deguif Go for it.

PR #1894, opened 2020-12-28
Adding new ElasticSearch 7.10 'case_insensitive' param to Wildcard.

ktherage:
Elasticsearch 7.x introduced a new parameter in Wildcard query which was missing in the Elastica object, the purpose of this pull request is to add it.

### DIFF
--- a/src/Query/Wildcard.php
+++ b/src/Query/Wildcard.php
@@ -30,7 +30,7 @@ class Wildcard extends AbstractQuery
         $this->setParam($field, [
             'value' => $value,
             'boost' => $boost,
-            'case_insensitive' => $caseInsensitive
+            'case_insensitive' => $caseInsensitive,
         ]);
     }
 
--- a/src/Query/Wildcard.php
+++ b/src/Query/Wildcard.php
@@ -23,13 +23,14 @@ class Wildcard extends AbstractQuery
      */
     private $field;
 
-    public function __construct(string $field, string $value, float $boost = 1.0)
+    public function __construct(string $field, string $value, float $boost = 1.0, bool $caseInsensitive = true)
     {
         $this->field = $field;
 
         $this->setParam($field, [
             'value' => $value,
             'boost' => $boost,
+            'case_insensitive' => $caseInsensitive
         ]);
     }
 
--- a/src/Query/Wildcard.php
+++ b/src/Query/Wildcard.php
@@ -31,7 +31,7 @@ class Wildcard extends AbstractQuery
             'value' => $value,
             'boost' => $boost,
         ]);
-        if (null !== $caseInsensitive){
+        if (null !== $caseInsensitive) {
             $this->setCaseInsensitive($caseInsensitive);
         }
     }
--- a/src/Query/Wildcard.php
+++ b/src/Query/Wildcard.php
@@ -23,15 +23,17 @@ class Wildcard extends AbstractQuery
      */
     private $field;
 
-    public function __construct(string $field, string $value, float $boost = 1.0, bool $caseInsensitive = true)
+    public function __construct(string $field, string $value, float $boost = 1.0, ?bool $caseInsensitive = null)
     {
         $this->field = $field;
 
         $this->setParam($field, [
             'value' => $value,
             'boost' => $boost,
-            'case_insensitive' => $caseInsensitive,
         ]);
+        if (null !== $caseInsensitive){
+            $this->setCaseInsensitive($caseInsensitive);
+        }
     }
 
     public function getField(): string
--- a/src/Query/Wildcard.php
+++ b/src/Query/Wildcard.php
@@ -23,13 +23,14 @@ class Wildcard extends AbstractQuery
      */
     private $field;
 
-    public function __construct(string $field, string $value, float $boost = 1.0)
+    public function __construct(string $field, string $value, float $boost = 1.0, bool $caseInsensitive = false)
     {
         $this->field = $field;
 
         $this->setParam($field, [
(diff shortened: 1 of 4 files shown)


src/Query/Wildcard.php
```diff
@@ -23,13 +23,14 @@ class Wildcard extends AbstractQuery
      */
     private $field;
 
-    public function __construct(string $field, string $value, float $boost = 1.0)
+    public function __construct(string $field, string $value, float $boost = 1.0, bool $caseInsensitive = false)
```
deguif, line 26: Should probably not be passed as argument to constructor as it's optional.
As an example `rewrite` option is not passed as argument here.

ruflin:
Could you also add a changelog? Can you also check CI?

ktherage:
> Could you also add a changelog? Can you also check CI?

@ruflin, I've adapted things but CI is, IMO, fakely green as the test I have added needs to have a 7.10 version of elasticsearch for being run.

ruflin:
@ktherage Thanks for the CI check. I guess it is time to update the running ES version to 7.10? If you have time, could you open a separate PR for this to get it in first and then rebase this one?

Could you add a changelog entry?

deguif:
@ruflin I think we still not have fixed the issue you encountered when trying the version 7.4 (https://github.com/ruflin/Elastica/pull/1686) so it's probably a little bit harder than just updating the ES version ;)

src/Query/Wildcard.php
```diff
@@ -67,4 +67,15 @@ public function setRewrite(string $rewriteMode): self
 
         return $this;
     }
+
+    /**
+     * @see https://www.elastic.co/guide/en/elasticsearch/reference/7.10/query-dsl-wildcard-query.html#wildcard-query-field-params
+     */
```
thePanz, line 73: Should we add a note here, stating that this is only working for ES >= v7.10?

ruflin:
@deguif Pushed a few fixes to the other PR. Could you take a look?

src/Query/Wildcard.php
ruflin: ++, that would be a nice indicator.

deguif: Yes that will be the same for #1874 and setFixedInterval() and setCalendarInterval() but for ES 7.2

ktherage:
Hello,

I just need a small update on that PR, must I change something on that ?
I didn't followed all what happened on it.

ruflin:
@ktherage Sorry for the late response here. What I did was pushing an update to https://github.com/ruflin/Elastica/pull/1686 but somehow it failed CI. I need to work again on this to make sure we are compatible with 7.10.2 which should also help this PR. If you get to it before, feel free to pull the changes form this PR and open a new one with a fix.

ktherage:
@ruflin, PR with CI seems OK now is there anything blocking that would require some work ?

ruflin:
Unfortunately I have not fixed https://github.com/ruflin/Elastica/pull/1686 yet. I assume as long as this is not merged, this PR will always be green as the test is not run? I'll try to find some time for https://github.com/ruflin/Elastica/pull/1686 to get it in.

Could you have a look at this comment? https://github.com/ruflin/Elastica/pull/1894#discussion_r559423334

ruflin:
@ktherage I just merged https://github.com/ruflin/Elastica/pull/1686. Could you rebase on master or merge in master and push again to run CI on the change?

ktherage:
@ruflin, I've rebased the branch on master but it seems that an approval is needed to get the CI running.

tests/Query/WildcardTest.php
```diff
@@ -106,4 +106,31 @@ public function testSearchWithAnalyzer(): void
...
+
+    /**
+     * @group functional
+     */
+    public function testCaseInsensitive(): void
+    {
+        // feature doesn't exist on version prior 7.10;
+        $this->_checkVersion('7.10');
+
+        foreach ([true, false] as $expected) {
```
thePanz, line 118: Can we use a @dataprovider annotation instead of having a foreach?

ktherage: I've changed it.

src/Query/Wildcard.php
```diff
@@ -23,13 +23,14 @@ class Wildcard extends AbstractQuery
      */
     private $field;
 
-    public function __construct(string $field, string $value, float $boost = 1.0)
+    public function __construct(string $field, string $value, float $boost = 1.0, bool $caseInsensitive = true)
```
thePanz, line 26: From the ES documentation:
- the parameter is optional, thus "nullable" IMHO → in that case it should not appear in the query at all
- the default value is "false", not "true"

Would it make sense to set  the parameter only if a value is passed, and have it null by default? wdyt?

ktherage: why not, I'm changing it.

thePanz (approved):

thePanz (approved):

thePanz:
@ktherage : "Could you also add a changelog?" (ref @ruflin  :) )

src/Query/Wildcard.php
thePanz: @ktherage I completely missed this, my bad
→ makes sense to remove it from the constructor, as for the "rewrite" parameter

ktherage: @thePanz, removed :wink:

thePanz (approved):

thePanz:
Thank you @ktherage! Merged :tada:

ruflin:
Thanks for getting this in!

PR #1898, opened 2021-02-11
Add MatchQuery class for forward compat

### LINKED ISSUE #1879: Backport MatchQuery into 6.x branch
Related to #1799 and #1793

While it's not so important that the 6.x branch supports PHP 8.0 (at least for us, for other people it might be), it'd be useful if `MatchQuery` in some form or another is backported into the 6.x branch, allowing code that needs to support 6.x | 7.x is able to set a version constraint to bring in the newer version of 6.x (ie 6.2 onwards or similar), and have the `MatchQuery` class available for both branches, to ease later migration burdens, and also not having to support both `Match` and `MatchQuery` in extends and use clauses etc.

In the simplest form, I guess a `class_alias` or `class MatchQuery extends Match {}` would do it. But for ease, it might be "easier" to backport all of #1799 to keep any subsequent backports and such easier (if they happen to touch the same code).

reedy:
Fixes #1879

### DIFF
--- a/lib/Elastica/Query/Match.php
+++ b/lib/Elastica/Query/Match.php
@@ -8,6 +8,8 @@ namespace Elastica\Query;
  * @author F21
  * @author WONG Wing Lun <luiges90@gmail.com>
  *
+ * @deprecated since version 6.1.2, use the MatchQuery class instead.
+ *
  * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-match-query.html
  */
 class Match extends AbstractQuery
--- a/lib/Elastica/Query/MatchQuery.php
+++ b/lib/Elastica/Query/MatchQuery.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Elastica\Query;
+
+/**
+ * Match query.
+ *
+ * This class is for forward compatibility reasons. For PHP 8 and above use MatchQuery as Match is reserved.
+ */
+class MatchQuery extends Match
+{
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Change Log
 All notable changes to this project will be documented in this file based on the [Keep a Changelog](http://keepachangelog.com/) Standard. This project adheres to [Semantic Versioning](http://semver.org/).
 
-## [Unreleased](https://github.com/ruflin/Elastica/compare/6.1.1...master)
+## [Unreleased](https://github.com/ruflin/Elastica/compare/6.1.1...6.x)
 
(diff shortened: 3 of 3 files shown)


lib/Elastica/Query/Match.php
```diff
@@ -8,6 +8,8 @@
  * @author F21
  * @author WONG Wing Lun <luiges90@gmail.com>
  *
+ * @deprecated since version 6.1.2, use the MatchQuery class instead.
```
reedy, line 11: Guessed that 6.1.2 would be the next release of 6.x..

lib/Elastica/Query/MatchQuery.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Elastica\Query;
+
+/**
+ * Match query.
+ *
+ * This class is for forward compatibility reasons. For PHP 8 and above use MatchQuery as Match is reserved.
+ */
+class MatchQuery extends Match
```
deguif, line 10: I think that won't work as `Match` is reserved (even when using extend).

reedy: It doesn't need to.

6.x only supports PHP ^7.0, so unless you force it, it won't run on PHP 8.0 - https://github.com/ruflin/Elastica/blob/6.x/composer.json#L15

I'm not adding it to add PHP 8.0 support, I'm doing it so people can migrate the code and be able to use newer 6.x versions or 7.x with the MatchQuery class, and not need to have conditional support for the class existing (which can be a pain when you're extending this class, for example)

reedy: See comments on https://github.com/ruflin/Elastica/issues/1879

It was suggested to avoid backporting all of https://github.com/ruflin/Elastica/pull/1799

deguif: Thanks, sorry I didn't understand at first glance ;)

deguif (approved):
LGTM.

Could you add a changelog?

ruflin (approved):

lib/Elastica/Query/MatchQuery.php
```diff
@@ -0,0 +1,11 @@
...
+{
```
reedy, line 11: Well, this is embarrassing. And apparently CI didn't pick this up.

This isn't valid PHP, there's a missing }

deguif: Yes CI didn't pick it on 6.x as it was broken when Travis stopped testing when quota was reached.
On 7.x CI was migrated to github action.

Issue #1908, opened 2021-03-07
ruflin/elastica 6.x is not compatible with elasticsearch/elasticsearch-php 6.8


reedy:
Downstream task: https://phabricator.wikimedia.org/T276320

6.x of ruflin/elastica as per https://github.com/ruflin/Elastica/blob/6.x/composer.json#L17 supports `"elasticsearch/elasticsearch": "^6.0"`

However, trying to use 6.x of ruflin/elastica with 6.8 of elasticsearch/elasticsearch causes errors due to https://github.com/ruflin/Elastica/blob/6.x/lib/Elastica/Index.php#L451

```
[3e8bfb9228c410ca976223cf] [no req]   Error from line 451 of /var/www/html/vendor/ruflin/elastica/lib/Elastica/Index.php: Class 'Elasticsearch\Endpoints\Indices\Aliases\Update' not found
Backtrace:
#0 /var/www/html/extensions/CirrusSearch/includes/Maintenance/Validators/SpecificAliasValidator.php(137): Elastica\Index->addAlias(string, boolean)
#1 /var/www/html/extensions/CirrusSearch/includes/Maintenance/Validators/SpecificAliasValidator.php(79): CirrusSearch\Maintenance\Validators\SpecificAliasValidator->updateFreeIndices(array)
#2 /var/www/html/extensions/CirrusSearch/includes/Maintenance/Validators/IndexAliasValidator.php(98): CirrusSearch\Maintenance\Validators\SpecificAliasValidator->updateIndices(array, array)
#3 /var/www/html/extensions/CirrusSearch/maintenance/UpdateOneSearchIndexConfig.php(458): CirrusSearch\Maintenance\Validators\IndexAliasValidator->validate()
#4 /var/www/html/extensions/CirrusSearch/maintenance/UpdateOneSearchIndexConfig.php(411): CirrusSearch\Maintenance\UpdateOneSearchIndexConfig->validateSpecificAlias()
#5 /var/www/html/extensions/CirrusSearch/maintenance/UpdateOneSearchIndexConfig.php(269): CirrusSearch\Maintenance\UpdateOneSearchIndexConfig->validateAlias()
#6 /var/www/html/extensions/CirrusSearch/maintenance/UpdateSearchIndexConfig.php(61): CirrusSearch\Maintenance\UpdateOneSearchIndexConfig->execute()
[...]

PR #1918, opened 2021-03-16
Migrate CI to github actions

deguif:


deguif:
It wasn't much fun but that should work.
CI is broken on PHP 7.3 and 7.4 because of https://github.com/ruflin/Elastica/issues/1908 but @ezimuel is fixing it.

I think we should a add a conflict to avoid installing `elasticsearch/elasticsearch` 6.8.0 as it's broken.

ezimuel:
I've the PR https://github.com/elastic/elasticsearch-php/pull/1114 for fixing 6.8.0.

deguif:
Yes, thanks @ezimuel for that ;)

I've just added the conflict for specific version (`6.8.0`), this should allow to install `6.8.1` when it will be released, but disallow `6.8.0`.

deguif:
@ruflin could you have a look.
I think we should release a `6.1.4` after merging this one to avoid the issue with `elasticsearch/elasticsearch` 6.8.0

ruflin (approved):
LGTM. Thanks for taking this on.

PR #1920, opened 2021-03-17
Fix isset() on unexisting variable

deguif:
This was caught with PHPStan as I'm working on adding static code analysis on the project.
See https://github.com/deguif/Elastica/runs/2130228841?check_suite_focus=true for WIP


ruflin (commented):
Nice, this looks like a bug. Changelog?

deguif:
@ruflin I just added a changelog entry.

ruflin (approved):

PR #1923, opened 2021-03-17
Replace call_user_func() by direct calls

deguif:


ruflin (approved):
Beautiful new world

kissifrot:
Out of curiosity, does it improve performance too?

deguif:
> Out of curiosity, does it improve performance too?

It should, but very very little as it avoids an extra function call

PR #1929, opened 2021-03-18
Use strict option for in_array() function

deguif:


src/Status.php
```diff
@@ -69,35 +69,29 @@ public function getIndexNames()
     /**
      * Checks if the given index exists.
      *
-     * @param string $name Index name to check
-     *
      * @return bool True if index exists
      */
-    public function indexExists($name)
+    public function indexExists(string $name)
```
deguif, line 74: It should not be a BC break as since PHP 7.2 and the parameter type widening  feature (https://wiki.php.net/rfc/parameter-no-type-variance) a subclass is allowed to widen the type, so that shouldn't break existing code overriding this method.

ruflin (approved):
Just in case it will break something, should we mention it in the changelog?

deguif:
> Just in case it will break something, should we mention it in the changelog?

You're right, I'll add a changelog.

PR #1936, opened 2021-03-19
Remove some functional test skip conditions

deguif:


ruflin (approved):
LGTM. Trying to remember why we had it in and not need it anymore.

PR #1940, opened 2021-03-22
Deprecate ivory http adapter transport

deguif:
https://github.com/egeloen/ivory-http-adapter is deprecated, let's deprecate it in this project too.


ruflin (approved):
LGTM: Curious to see if there will be some push back on this.

Issue #1963, opened 2021-05-28
The Phrase suggester should support multiple direct_generators


sergiuionescu:
The current Phrase suggester(https://github.com/ruflin/Elastica/blob/master/src/Suggest/Phrase.php#L12) only supports one generator. Subsequently calling \Elastica\Suggest\Phrase::addCandidateGenerator always overwrites the previous generator value.

According to https://www.elastic.co/guide/en/elasticsearch/reference/current/search-suggesters.html#_direct_generators multiple generators can be used. Ex:
```
{
    "suggest": {
        "text": "obel prize",
        "simple_phrase": {
            "phrase": {
                "field": "title.trigram",
                "size": 1,
                "direct_generator": [
                    {
                        "field": "title.trigram",
                        "suggest_mode": "always"
                    },
                    {
                        "field": "title.reverse",
                        "suggest_mode": "always",
                        "pre_filter": "reverse",
                        "post_filter": "reverse"
                    }
                ]
            }
        }
    }
}
```

At the moment the workaround we can use is:

```
\Elastica\Suggest\Phrase::setParam('direct_generator', [
    $trigramGenerator->toArray()['direct_generator'],
    $reverseGenerator->toArray()['direct_generator'],
])
```
Is it possible to update the Phrase suggester class to support a list of generators?

Thanks!

ruflin:
@sergiuionescu Interested to open a PR? 😇

sergiuionescu:
@ruflin Please have a look at https://github.com/ruflin/Elastica/pull/1964 . 
Please runt the github action tests, i haven't been able to spin up the integration env locally.

ruflin:
Thanks! Test should always be run automatically.

deguif:
@ruflin they are not run anymore automatically for new contributors.
I had to approve @sergiuionescu work for running the tests as github recently changed this behaviour to avoid abuses (crypto mining, ...).

deguif:
See https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks for more details

deguif:
Closing as #1964 was merged and fixed that.

PR #1964, opened 2021-05-31
1963 - support multiple direct_generators

sergiuionescu:
See https://github.com/ruflin/Elastica/issues/1963

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,15 +6,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/7.1.1...master)
 ### Backward Compatibility Breaks
+* Changed `Elastica\Exception\JSONParseException` inheritance, it now extends `\JsonException` instead of `\RuntimeException` [#1949](https://github.com/ruflin/Elastica/pull/1949)
 ### Added
-
 * Added `Elastica\Aggregation\NormalizeAggregation` [#1956](https://github.com/ruflin/Elastica/pull/1956)
-* Added `\Elastica\Suggest\Phrase::addDirectGenerator` to align with ES specification [#1964](https://github.com/ruflin/Elastica/pull/1964)
-
+* Added `Elastica\Suggest\Phrase::addDirectGenerator` to align with ES specification [#1964](https://github.com/ruflin/Elastica/pull/1964)
 ### Changed
 * Updated `php-cs-fixer` to `2.18.6` [#1955](https://github.com/ruflin/Elastica/pull/1955)
+* Updated `php-cs-fixer` to `3.0.0` [#1959](https://github.com/ruflin/Elastica/pull/1959)
 * Using default Elasticsearch images for testing instead of OSS https://github.com/ruflin/Elastica/pull/1954
-
 ### Deprecated
 ### Removed
 ### Fixed
@@ -35,6 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Updated `composer-normalize` to `2.13.3` [#1927](https://github.com/ruflin/Elastica/pull/1927)
 ### Deprecated
 * Deprecated `Elastica\Transport\HttpAdapter` class [#1940](https://github.com/ruflin/Elastica/pull/1940)
+* Deprecated `Elastica\Exception\JSONParseException` exception, catch `\JsonException` instead [#1949](https://github.com/ruflin/Elastica/pull/1949)
 ### Fixed
 * Fixed wrong `ltrim` usage in guzzle transport [#1783](https://github.com/ruflin/Elastica/pull/1783)
 * Fixed `_seq_no` and `_primary_term` wrong initialization [#1920](https://github.com/ruflin/Elastica/pull/1920)
@@ -108,7 +108,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fixed handling precision as string in `Elastica\Aggregation\GeohashGrid::setPrecision()` [#1884](https://github.com/ruflin/Elastica/pull/1884)
 * Fixed calling `Elastica\Aggregation\Composite::addAfter()` with the `null` value [1877](https://github.com/ruflin/Elastica/pull/1877)
 * Replaced `_routing` and `_retry_on_conflict` by `routing` and `retry_on_conflict` in `AbstractUpdateAction` [#1807](https://github.com/ruflin/Elastica/issues/1807)
-### Security
 
 
 ## [7.0.0](https://github.com/ruflin/Elastica/compare/7.0.0-beta.4...7.0.0)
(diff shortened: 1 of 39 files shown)


src/Suggest/Phrase.php
```diff
@@ -140,7 +140,7 @@ public function setSmoothingModel(string $model, array $params): Phrase
      */
     public function addCandidateGenerator(AbstractCandidateGenerator $generator): Phrase
```
deguif, line 141: What about deprecating this method, and instead add a new one `addDirectGenerator(DirectGenerator $generator)` as this suggester only allow configuring `direct_generator`. That would this way avoid to trick to `toArray` method.

ruflin (commented):
Thanks for the contribution. Could you also add a test and a changelog entry?

Also see comment from @deguif

sergiuionescu:
> Thanks for the contribution. Could you also add a test and a changelog entry?
> 
> Also see comment from @deguif

@deguif I added the new method and deprecated the old one since it was a bit confusing to use. The toArray trick was required by `\Elastica\Suggest\CandidateGenerator\DirectGenerator::toArray` which is also updated.

@ruflin I updated the changelog. When do you think you could include this in a patch version?

Thanks.

src/Suggest/Phrase.php
```diff
@@ -138,30 +139,18 @@ public function setSmoothingModel(string $model, array $params): Phrase
...
-    public function addCandidateGenerator(AbstractCandidateGenerator $generator): Phrase
+    public function addDirectGenerator(DirectGenerator $generator): Phrase
     {
-        return $this->setParam('candidate_generator', $generator);
+        return $this->addParam('direct_generator', $generator);
     }
 
     /**
-     * {@inheritdoc}
+     * @deprecated Please use addDirectGenerator instead.
```
deguif, line 148:
```suggestion
     * @deprecated since version 7.2.0, use the "addDirectGenerator()" method instead.
```

src/Suggest/Phrase.php
```diff
@@ -138,30 +139,18 @@ public function setSmoothingModel(string $model, array $params): Phrase
...
-            unset($array[$baseName]['candidate_generator']);
-
-            $keys = \array_keys($generator);
-            $values = \array_values($generator);
-
-            $array[$baseName][$keys[0]][] = $values[0];
-        }
-
-        return $array;
+        return $this->addParam('candidate_generator', $generator);
```
deguif, line 154:
```suggestion
        \trigger_deprecation('ruflin/elastica', '7.2.0', 'The "%s()" method is deprecated, use the "addDirectGenerator()" method instead. It will be removed in 8.0.', __METHOD__);

        return $this->addParam('candidate_generator', $generator);
```

src/Suggest/Phrase.php
```diff
@@ -138,30 +139,18 @@ public function setSmoothingModel(string $model, array $params): Phrase
     /**
      * @return $this
      */
-    public function addCandidateGenerator(AbstractCandidateGenerator $generator): Phrase
+    public function addDirectGenerator(DirectGenerator $generator): Phrase
```
deguif, line 142:
```suggestion
    public function addDirectGenerator(DirectGenerator $generator): self
```

src/Suggest/CandidateGenerator/DirectGenerator.php
```diff
@@ -132,4 +132,20 @@ public function setPostFilter(string $post)
...
+     */
+    public function toArray()
+    {
+        $data = $this->getParams();
+
+        if (!empty($this->_rawParams)) {
+            $data = \array_merge($data, $this->_rawParams);
+        }
+
+        return $this->_convertArrayable($data);
```
deguif, line 149: Implementing NameableInterface here should do the trick and avoid customizing the `toArray()`

deguif (commented):
Thanks @sergiuionescu for your work.
Some new comments related to how we handle deprecations.
Another one related to `toArray` change that can probably be done another way more standard.

src/Suggest/CandidateGenerator/DirectGenerator.php
sergiuionescu: @deguif Thanks for suggesting the changes and align this with the future changes for the lib.

 I do not see how implementing NameableInterface will do the trick here. The handling for that interface is:  
`$arr[$value instanceof NameableInterface ? $value->getName() : $key] = $value->toArray();` So we will still get a map with the "direct_generator" name and params, instead of a list of direct_generator's params.

We are looking to get 
```
{
    "suggest": {
        "text": "obel prize",
        "simple_phrase": {
            "phrase": {
                "field": "title.trigram",
                "size": 1,
                "direct_generator": [
                    {
                        "field": "title.trigram",
                        "suggest_mode": "always"
                    },
                    {
                        "field": "title.reverse",
                        "suggest_mode": "always",
                        "pre_filter": "reverse",
                        "post_filter": "reverse"
                    }
                ]
            }
        }
    }
}
```

I think the override is the more transparent approach. Am i missing something with the NameableInterface, do you have an example where that is used to build a list?

sergiuionescu: @deguif Please run the tests whenever possible, i reverted to the `\Elastica\Suggest\Phrase::toArray` override.

deguif: Yes done again, I did it this morning but Github was encountering issues.

CHANGELOG.md
```diff
@@ -19,6 +19,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 ### Security
 
+## [7.1.2](https://github.com/ruflin/Elastica/compare/7.1.1...7.1.2)
+### Added
+* Added `\Elastica\Suggest\Phrase::addDirectGenerator` to align with ES specification [#1964](https://github.com/ruflin/Elastica/pull/1964)
```
deguif, line 24: This should be moved to the upper `Added section`

CHANGELOG.md
```diff
@@ -19,6 +19,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 ### Security
 
+## [7.1.2](https://github.com/ruflin/Elastica/compare/7.1.1...7.1.2)
```
deguif, line 22: This should be removed as it's done upper. Proper version will be edited when preparing the next release.

tests/Suggest/PhraseTest.php
```diff
@@ -46,6 +46,29 @@ public function testToArray(): void
...
+        // 3 suggestions should be returned: One in which both misspellings are corrected, and two in which only one misspelling is corrected.
+        $this->assertCount(3, $suggests['suggest1'][0]['options']);
+
+        $this->assertEquals('elasticsearch is <suggest>bonsai cool</suggest>', $suggests['suggest1'][0]['options'][0]['highlighted']);
+        $this->assertEquals('elasticsearch is bonsai cool', $suggests['suggest1'][0]['options'][0]['text']);
+    }
+
+    /**
+     * @group functional
+     */
```
deguif, line 70:
```suggestion
     * @group legacy
     */
```

deguif: And probably add a check for deprecation triggering (using the proper trait): 
```
$this->expectDeprecation('Since ruflin/elastica 7.2.0: XXXX . It will be removed in 8.0.');
```

deguif (approved):

deguif:
Thanks @sergiuionescu

sergiuionescu:
> Thanks @sergiuionescu

Thank you @deguif , @ruflin for the quick action and support!

PR #1966, opened 2021-06-02
Replace avoidable empty() calls

deguif:


src/Node.php
```diff
@@ -72,7 +72,7 @@ public function setId(string $id): self
      */
     public function getName(): string
     {
-        if (empty($this->_name)) {
+        if (null === $this->_name) {
```
ruflin, line 75: If _name is '', what should happen?

deguif: I'm pretty sure here the goal was to initialize the property as its default value is null. So when unitiliazed, the call to call getInfo is made, and when initiliazed it returns the previous value.

ruflin (approved):
LGTM and all tests are still green (I wonder if we test all this properly :-D )

PR #1973, opened 2021-08-19
Fix getDocument to return the version parameters

dsgrillo:


ruflin:
Any chance you could add a changelog and a test that confirms this fix?

dsgrillo:
Hi @ruflin, I was in a rush last week to have this working, sorry for the low quality PR.
I'd be happy to add a couple of tests to confirm the issue and update the changelog accordingly :)

ruflin:
Thank you! I really appreciate the contribution!

tests/ClientFunctionalTest.php
```diff
@@ -219,6 +219,28 @@ public function testUpdateDocuments(): void
...
+            1,
+            ['name' => 'hans', 'age' => $initialValue],
+            $indexName
+        );
+        $data = [$doc1];
+        $client->addDocuments($data);
+
+        $doc1 = $index->getDocument(1);
+
+        $this->assertGreaterThan(0, $doc1->getVersion());
```
ruflin, line 241: Before this was 0?

src/Query/MoreLikeThis.php
```diff
@@ -170,6 +170,8 @@ public function toArray(): array
             $doc = $array['more_like_this']['like'];
             $doc['doc'] = $array['more_like_this']['like']['_source'];
             unset($doc['_id'], $doc['_source']);
+            unset($doc['if_seq_no']);
```
ruflin, line 173: Trying to figure out how this change is related?

dsgrillo: The `if_seq_no` and `if_primary_term` parameters are now set by the fix and those parameters are not expected by the MoreLikeThis query. 
Find out because of a failing test.

tests/ClientFunctionalTest.php
dsgrillo: Previously it would fail with: `Elastica\Exception\InvalidException: Param version does not exist`

PR #1977, opened 2021-09-10
6.x compatible with php8.0

AskaoAhmedSaad:
make version 6.x compatible with php8.0


lib/Elastica/Query/Match.php
```diff
@@ -1,216 +0,0 @@
-<?php
```
ruflin, line 1: This looks like a breaking change?

ruflin (commented):
@AskaoAhmedSaad It would be nice if we could make 6.x compatible with PHP 8.0 but we should not make it a breaking change for existing users. Any ideas on how we can make it non breaking? I remember we had similar challenges with other PHP versions and reserved terms. Can you check it how we have solved it in other places / PR?

reedy:
I was going to try and make a similar patch...

Don't remove the `Match` class; if it exists and isn't executed, it won't be a problem for PHP 8.0

Swap `Match` and `MatchQuery` around (so the `Match` alias extends `MatchQuery` instead of the other way around), and then update documentation.

If people are still using `Match` as the canonical in their own code, they'll need to update that to work on PHP 8.0 anyway, so that's not a problem. Anyone still using `Match` and not needing to support PHP 8.0 doens't need to make any changes

composer.json
```diff
@@ -12,7 +12,7 @@
         }
     ],
     "require": {
-        "php": "^7.0",
+        "php": "^7.0 | ^8.0",
```
reedy, line 15: Use `||` as per https://github.com/ruflin/Elastica/blob/master/composer.json#L18 for consistency

lib/Elastica/Query/Match.php
```diff
@@ -1,216 +0,0 @@
...
-/**
- * Match query.
- *
- * @author F21
- * @author WONG Wing Lun <luiges90@gmail.com>
- *
- * @deprecated since version 6.1.2, use the MatchQuery class instead.
- * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-match-query.html
- */
-class Match extends AbstractQuery
```
reedy, line 14: This should become `class Match extends MatchQuery` and otherwise be empty of content

lib/Elastica/Query/MatchNone.php
```diff
@@ -3,7 +3,7 @@
 namespace Elastica\Query;
 
 /**
- * Match none query. Returns no results.
+ * MatchQuery none query. Returns no results.
```
reedy, line 6: This shouldn't be replaced. See master https://github.com/ruflin/Elastica/blob/master/src/Query/MatchNone.php#L6

lib/Elastica/Query/MatchAll.php
```diff
@@ -3,7 +3,7 @@
 namespace Elastica\Query;
 
 /**
- * Match all query. Returns all results.
+ * MatchQuery all query. Returns all results.
```
reedy, line 6: This shouldn't be replaced.

lib/Elastica/Query/MatchPhrase.php
```diff
@@ -3,7 +3,7 @@
 namespace Elastica\Query;
 
 /**
- * Match Phrase query.
+ * MatchQuery Phrase query.
```
reedy, line 6: This shouldn't be replaced.

lib/Elastica/Query/MatchPhrasePrefix.php
```diff
@@ -3,7 +3,7 @@
 namespace Elastica\Query;
 
 /**
- * Match Phrase Prefix query.
+ * MatchQuery Phrase Prefix query.
```
reedy, line 6: This shouldn't be replaced.

lib/Elastica/Query/MatchQuery.php
```diff
@@ -3,10 +3,214 @@
 namespace Elastica\Query;
 
 /**
- * Match query.
+ * MatchQuery query.
```
reedy, line 6: This shouldn't be replaced.

lib/Elastica/Query/MatchQuery.php
```diff
@@ -3,10 +3,214 @@
...
  *
- * This class is for forward compatibility reasons. For PHP 8 and above use MatchQuery as Match is reserved.
+ * @author F21
+ * @author WONG Wing Lun <luiges90@gmail.com>
+ *
+ * @deprecated since version 6.1.2, use the MatchQuery class instead.
```
reedy, line 11: This is incorrect, it's not deprecated.

lib/Elastica/Query/MultiMatch.php
```diff
@@ -3,7 +3,7 @@
 namespace Elastica\Query;
 
 /**
- * Multi Match.
+ * Multi MatchQuery.
```
reedy, line 6: This shouldn't be replaced.

lib/Elastica/Query/MultiMatch.php
```diff
@@ -80,7 +80,7 @@ public function setTieBreaker($tieBreaker = 0.0)
     }
 
     /**
-     * Sets operator for Match Query.
+     * Sets operator for MatchQuery Query.
```
reedy, line 83: This shouldn't be replaced.

lib/Elastica/Query/MultiMatch.php
```diff
@@ -94,7 +94,7 @@ public function setOperator($operator = self::OPERATOR_OR)
     }
 
     /**
-     * Set field minimum should match for Match Query.
+     * Set field minimum should match for MatchQuery Query.
```
reedy, line 97: This shouldn't be replaced.

lib/Elastica/Query/MultiMatch.php
```diff
@@ -106,7 +106,7 @@ public function setMinimumShouldMatch($minimumShouldMatch)
     }
 
     /**
-     * Set zero terms query for Match Query.
+     * Set zero terms query for MatchQuery Query.
```
reedy, line 109: This shouldn't be replaced.

lib/Elastica/Query/MultiMatch.php
```diff
@@ -120,7 +120,7 @@ public function setZeroTermsQuery($zeroTermQuery = self::ZERO_TERM_NONE)
     }
 
     /**
-     * Set cutoff frequency for Match Query.
+     * Set cutoff frequency for MatchQuery Query.
```
reedy, line 123: This shouldn't be replaced.

test/Elastica/Query/PercolateTest.php
```diff
@@ -26,7 +26,7 @@ public function testPercolateQueryOnNewDocument()
         $doc = new Document(2, ['message' => 'A new bonsai tree in the office']);
         $this->documentType->addDocument($queryDoc);
         $this->index->refresh();
-        //Match a document to the registered percolator queries:
+        //MatchQuery a document to the registered percolator queries:
```
reedy, line 29: This shouldn't be replaced.

reedy (changes requested):

reedy:
Looks like this has just been done as a literal find and replace. Only really the code uses of the `Match` query needed updating. Replacing most (all?) of the `Match` comments with `MatchQuery` make no sense and don't align with what's on master.

lib/Elastica/Query/Match.php
reedy: It definitely is :)

lib/Elastica/Query/Match.php
reedy: Keeping the comments above too, of course.

AskaoAhmedSaad:
@reedy @ruflin having a class with the name "Match" gives a parser error at PHP 8 environment, do you have a suggestion to overcome this and at the same time not to break the other PHP versions ?

I think maybe we can have a special version name with removing Match class totally and replace it with MatchQuery

reedy:
> @reedy @ruflin having a class with the name "Match" gives a parser error at PHP 8 environment, do you have a suggestion to overcome this and at the same time not to break the other PHP versions ?
> 
> I think maybe we can have a special version name with removing Match class totally and replace it with MatchQuery

It should only error if you execute the file using it (or lint it etc)... What are you doing to cause/see that error?

reedy:
I re-made your commit without the comment changes in https://github.com/reedy/Elastica/tree/6xphp8

composer.json:
```
{
	"require": {
		"ruflin/elastica": "dev-6xphp8 as 6.2"
	},
	"repositories": [
		{
			"type": "git",
			"url": "https://github.com/reedy/elastica.git"
		}
	]
}
```

Test PHP script:
```
<?php

require __DIR__ . '/vendor/autoload.php';

use Elastica\Query\MatchQuery;

$mq = new MatchQuery();
echo "Done\n";
```

Run it

```
reedy@MacBook-Pro elasticatest % php --version    
PHP 8.0.10 (cli) (built: Aug 26 2021 15:36:17) ( NTS )
Copyright (c) The PHP Group
Zend Engine v4.0.10, Copyright (c) Zend Technologies
    with Zend OPcache v8.0.10, Copyright (c), by Zend Technologies
reedy@MacBook-Pro elasticatest % php test.php 
Done
```

reedy:
CI seems fine with the `Match` class existing too...

I've got https://github.com/ruflin/Elastica/pull/1981 to the point where I'm just fiddling with fixing up some minor test breakages (function calls that need updating)

ruflin:
As https://github.com/ruflin/Elastica/pull/1981 goes more into the direction we want to go I'm going to close this PR for now. @AskaoAhmedSaad Thanks for kicking this off and I would appreciate if you could chime in on https://github.com/ruflin/Elastica/pull/1981 In case we change direction, we can still reopen this one.

PR #1980, opened 2021-09-22
[6.x] Update Versions & Dependencies table

reedy:
Copy updated table from https://github.com/ruflin/Elastica/commit/8ecc5cdadd5bf0b69c1eb20e56246a940b43bdb7

Escapes `|` primarily so that the 7.x supported PHP version shows `^7.2 || ^8.0` rather than just `^7.2`


ruflin:
Thx for the fix.

Issue #1978, opened 2021-09-16
More Match class PHP 8 fun (in 6.x)


reedy:
From our downstream issue - https://phabricator.wikimedia.org/T268861#7359008

Because of how I did `MatchQuery extends Match`, `Match` is still executed, meaning it does fail on PHP 8.0. And that's kinda fine, as I did it for forward compat purposes (so people could switch to `MatchQuery` earlier, and depending on their code, be able to support later 6.x and also 7.x).

As per https://github.com/ruflin/Elastica/blob/6.1.5/composer.json#L15, 6.1.x doesn't support PHP 8 either (at least officially, you can obviously ignore certain requirements).

We're contemplating how to do PHP 8 migration for Wikipedia (et al), and as our ES upgrade might not have happened by then (and hence, also upgrading ruflin/elastica and elasticsearch/elasticsearch-php).

Would a patch to make MatchQuery the canonical (ie find and replace Match with MatchQuery, and then make the `class Match extend MatchQuery` instead) be accepted?

This is related to https://github.com/ruflin/Elastica/issues/1913... And would be needed to support that. As per the backtrace below (from the task/comment linked above)

```
[99a80ba4319b973cca1c54a7] /index.php?search=%C3%A9v%C3%A9nement+%3A+c&title=Sp%C3%A9cial%3ARecherche&go=Lire&ns0=1 ParseError: syntax error, unexpected token "match"

Backtrace:

from /var/www/vhosts/fallout-wiki.com/httpdocs/extensions/Elastica/vendor/ruflin/elastica/lib/Elastica/Query/MatchQuery.php(10)
```

And that line is https://github.com/ruflin/Elastica/blob/6.1.5/lib/Elastica/Query/MatchQuery.php#L10, so we would need to stop that being executed.

I think I did mumble about this in one of the prior tasks, expecting something like this might eventually happen.

So I guess the TLDR... Is would a patch swapping `MatchQuery` and `Match` be welcomed, in support of https://github.com/ruflin/Elastica/issues/1913 and potentially being able to make 6.x support PHP 8 (obviously might be other issues down the line).

Hikariii:
+1 for php8 support on the 6.x branch.

reedy:
Seems https://github.com/ruflin/Elastica/pull/1977 was created before I created this issue...

ruflin:
I definitively support this. The important part for me is that we don't break existing installations but it seems with this approach this is possible.

reedy:
[...]

PR #1981, opened 2021-09-23
[6.x] PHP8 support

reedy:
* Swap Match and MatchQuery around in heirarchy
* Start testing on PHP 8
* Add yoast/phpunit-polyfills for PHPUnit support
* PHPUnit version for PHP 8 support (8.5.12 is lowest version with PHP >= 7.2 support; newer 6.x is `^7.0` and 7.x is `^7.1`).
** Inspiration and help from https://github.com/ruflin/Elastica/commit/e701617059208fbd83276040677317f9ddc5dd40


reedy:
Doesn't at least break PHP 7.0-7.4 as is... :)

reedy:
Ok, and this is where the fun starts.

```
PHP Fatal error:  Declaration of Elastica\Test\Base::setUp() must be compatible with PHPUnit\Framework\TestCase::setUp(): void in /home/runner/work/Elastica/Elastica/test/Elastica/Base.php on line 13
```

It's not just `setUp()`, but `tearDown()`. The `void` return type wasn't introduced until PHP 7.1, so it would break 7.0 (which we need to support and run tests on).

Time for some polyfills, I guess :)

reedy:
Sweet. And that's tests passing on all versions before PHP 8.0... Progress!

reedy:
Only real outstanding problem I see (ignoring the 7.3 issues that are randomly intermittant), is one test failure on PHP 8.0

```

**Comment by reedy** [comment]
1) Elastica\Test\Transport\NullTransportTest::testGenerateDefaultResponse
Failed asserting that an array contains 'params'.

/home/runner/work/Elastica/Elastica/test/Elastica/Transport/NullTransportTest.php:110
```

`var_dump`-ing the array...

```
{
  ["took"]=>
  int(0)
  ["timed_out"]=>
  bool(false)
  ["_shards"]=>
  array(3) {
    ["total"]=>
    int(0)
    ["successful"]=>
    int(0)
    ["failed"]=>
    int(0)
  }
  ["hits"]=>
  array(3) {
    ["total"]=>
    int(0)
    ["max_score"]=>
    NULL
    ["hits"]=>
    array(0) {
    }
  }
  ["params"]=>
  array(1) {
    ["blah"]=>
    int(123)
  }
}
```

reedy:
Think CI is getting a little upset with me... `Elastica\Exception\ResponseException: blocked by: [FORBIDDEN/6/cluster read-only (api)];`

I think all that's really left is a handful of `assertContains` that need to become `assertArrayHasKey`

I'll have a further look when I've had some sleep

reedy:
```
There were 14 warnings:

1) Elastica\Test\Aggregation\DateRangeTest::testDateRangeSetFormatAccordingToFormatTargetField
Using assertContains() with string haystacks is deprecated and will not be supported in PHPUnit 9. Refactor your test to use assertStringContainsString() or assertStringContainsStringIgnoringCase() instead.

2) Elastica\Test\BulkTest::testErrorRequest
Using assertContains() with string haystacks is deprecated and will not be supported in PHPUnit 9. Refactor your test to use assertStringContainsString() or assertStringContainsStringIgnoringCase() instead.

3) Elastica\Test\Cluster\SettingsTest::testSetReadOnly
Using assertContains() with string haystacks is deprecated and will not be supported in PHPUnit 9. Refactor your test to use assertStringContainsString() or assertStringContainsStringIgnoringCase() instead.

4) Elastica\Test\Index\SettingsTest::testDeleteAliasWithException
Using assertContains() with string haystacks is deprecated and will not be supported in PHPUnit 9. Refactor your test to use assertStringContainsString() or assertStringContainsStringIgnoringCase() instead.

5) Elastica\Test\Index\SettingsTest::testSetReadOnly
Using assertContains() with string haystacks is deprecated and will not be supported in PHPUnit 9. Refactor your test to use assertStringContainsString() or assertStringContainsStringIgnoringCase() instead.

6) Elastica\Test\Index\SettingsTest::testNotFoundIndex
Using assertContains() with string haystacks is deprecated and will not be supported in PHPUnit 9. Refactor your test to use assertStringContainsString() or assertStringContainsStringIgnoringCase() instead.

7) Elastica\Test\PipelineTest::testPipelineCreate
Using assertContains() with string haystacks is deprecated and will not be supported in PHPUnit 9. Refactor your test to use assertStringContainsString() or assertStringContainsStringIgnoringCase() instead.

8) Elastica\Test\Processor\FailTest::testFailField
Using assertContains() with string haystacks is deprecated and will not be supported in PHPUnit 9. Refactor your test to use assertStringContainsString() or assertStringContainsStringIgnoringCase() instead.

9) Elastica\Test\Processor\KvTest::testKVIncludeExludeKeys
Using assertContains() with string haystacks is deprecated and will not be supported in PHPUnit 9. Refactor your test to use assertStringContainsString() or assertStringContainsStringIgnoringCase() instead.

10) Elastica\Test\Query\QueryStringTest::testSearchFieldsValidationException
Using assertContains() with string haystacks is deprecated and will not be supported in PHPUnit 9. Refactor your test to use assertStringContainsString() or assertStringContainsStringIgnoringCase() instead.

11) Elastica\Test\ResponseTest::testGetDataEmpty
Using assertContains() with string haystacks is deprecated and will not be supported in PHPUnit 9. Refactor your test to use assertStringContainsString() or assertStringContainsStringIgnoringCase() instead.

12) Elastica\Test\Type\MappingTest::testDynamicTemplate
assertArraySubset() is deprecated and will be removed in PHPUnit 9.

13) Elastica\Test\TypeTest::testUpdateDocumentWithParameter
Using assertContains() with string haystacks is deprecated and will not be supported in PHPUnit 9. Refactor your test to use assertStringContainsString() or assertStringContainsStringIgnoringCase() instead.

14) Elastica\Test\TypeTest::testUpdateDocumentWithoutSource
Using assertContains() with string haystacks is deprecated and will not be supported in PHPUnit 9. Refactor your test to use assertStringContainsString() or assertStringContainsStringIgnoringCase() instead.
```

ruflin:
I did not get to fully review this yet but supportive of this change.

@deguif Would be great to get your take on this.

reedy:
I think I am finished now (at least as tweaking commit summary and making tests pass)

reedy:
Accidentally deleted the branch in GH, restored it... And just rebased to fix the conflicts with my other merged PR

test/Elastica/DeprecatedClassBase.php
```diff
@@ -9,15 +9,15 @@
...
 {
-    protected function setUp()
+    protected function set_up()
     {
-        parent::setUp();
+        parent::set_up();
         $this->hideDeprecated();
     }
 
-    protected function tearDown()
```
thePanz, line 18: Not sure why we need this change, could you elaborate?
Is it to avoid using the `tearDown(): void`, which needs the `: void` return signature?

reedy: Yup, exactly. We can't use that on PHP 7.0, as it was added in 7.1.

Dropping 7.0 support would be a breaking change.

thePanz: Dropping php 7.0 would be still a good more IMHO, as it is not supported/maintained from Jan 2019 :)
See: https://www.php.net/supported-versions.php

reedy: Well, neither is PHP 7.1 or 7.2... ;)

But that's not necessarily my call to make

From https://github.com/ruflin/Elastica/issues/1978#issuecomment-925543910

>I definitively support this. The important part for me is that we don't break existing installations but it seems with this approach this is possible.

So I'm trying to follow that as much as possible.

Now, of course if we change the php version in composer.json, it wouldn't install on the older PHP (and keep using the older version of the 6.x branch), but there's always someone that uses `--ignore-platform-reqs`.

I don't think doing this as a one time change is too much of a maintenance burden.

There is a bigger question of when this branch becomes EOL/also unmaintained, but again, that's not really my call.

reedy: Oh, and it's not just for 7.0...

As per https://phpunit.de/supported-versions.html PHPUnit 8 supports PHP >= 7.2, meaning that composer will install PHP 8.5 on >= 7.2. And the differene between 6 and 8 brings in those return types.

So yeah, you could make the argument to drop 7.0 and 7.1 based on that (and then remove the old PHPUnit too), but again, breaking changes :).

ruflin:
Seems like we went from supporting 8.0 to dropping 7.0 / 7.1. Nobody likes breaking changes but lets see what we can do here.

* From a pure user perspective, can we keep compatibility with 7.0 so if someone upgrades to 6.x and runs php 7.0, things keep working?
* It seems many changes are related to the testing which require >7.1? Is it an option to require for testing a newer version? It has the downside we stop testing that 7.0 actually works but yes, it is EOL 2019
* We want to support PHP 8 in for 6.8 because ES 6.8 is still supported and will be around for a while. 
* Is there an option that we can build a new 6.x minor release where users that use 6.x and are on PHP 7.0 do not automatically upgrade?

reedy:
> Seems like we went from supporting 8.0 to dropping 7.0 / 7.1. Nobody likes breaking changes but lets see what we can do here.

Sorry, I'm confused. At what point have I dropped 7.0 and 7.1 support? I've not changed the PHP 7 requirement in composer.json, and I've not stopped the testing on 7.0 and 7.1 either.

And then all the tests are running, and they're passing on all the branches:

Screenshot 2021-09-30 at 12 39 18

>     * From a pure user perspective, can we keep compatibility with 7.0 so if someone upgrades to 6.x and runs php 7.0, things keep working?

Isn't that what I'm doing? See the above and below response

And as the tests/ directory is in `.gitattributes` as `export-ignore`, the changes to the person downloading and installing this library would be minimal.

>     * It seems many changes are related to the testing which require >7.1? Is it an option to require for testing a newer version? It has the downside we stop testing that 7.0 actually works but yes, it is EOL 2019

PHPUnit has fairly strict version compatability. There's no one version (without hacks like ignoring platform requirements) that would work on all versions...

https://phpunit.de/supported-versions.html

Screenshot 2021-09-30 at 12 43 18

Yes, we could just stop testing on old versions if we wanted. But of course, even PHP 7.3 is nearly out of security support too - https://www.php.net/supported-versions.php

>     * Is there an option that we can build a new 6.x minor release where users that use 6.x and are on PHP 7.0 do not automatically upgrade?

If we changed the version requirement, composer wouldn't upgrade to a version that is incompatible (ie that wouldn't run on the running PHP version). Someone can force this with the ignore platform requirements flag, but then any breakages would be on them (not that would necessarily stop some bug reports).

reedy:
Aha, no problem. I wasn't sure if I was missing something obvious.

I do agree that the PHPUnit changes aren't the cleanest (bringing in the polyfill), but as we're not likely going to be backporting features (beyond this one supporting PHP 8), it should be fairly minimal maintenance overhead. They're also roughly familiar (if you squint at least) to anyone familiar with PHPUnit.

I could've probably brought in https://github.com/ruflin/Elastica/commit/e701617059208fbd83276040677317f9ddc5dd40 as a cherry pick more wholesale to do the deprecated PHPUnit function updates, but possibly not worth it.

It's nice that the only really PHP_VERSION related hack I needed to make was to `test/Elastica/QueryBuilder/DSL/AbstractDSLTest.php` too.

Then the other changes to `lib/` are mostly a `s/Match/MatchQuery/` in code (not comments!) and then swapping `Match` and `MatchQuery` around in the code files.

User facing, this should basically be zero change. If people are still using `Match`, it still exists, but is slightly further down in the class heirarchy, but functionally otherwise the same.

We already have the conflicts for `"elasticsearch/elasticsearch": "6.8.0 || 6.8.1"` (which we know are broken), resulting in 6.8.2 being used on PHP 8.0, and 6.7.2 being on PHP 7.0.

ruflin (approved):
Approving change from my end but waiting on @thePanz or @deguif for a second opinion. I'll be out the next days so please directly get it in case of agreement.

thePanz (approved):

thePanz:
PR Approved, thank you @reedy !

thePanz:
Just to drop the idea here: I am not sure how much work will be needed, but his is an interesting strategy:

https://getrector.org/blog/how-to-bump-minimal-version-without-leaving-anyone-behind

PR #1983, opened 2021-09-23
[6.x] Require ext-json in composer (#1673)

reedy:


ruflin:
In case someone also looks for the reason on why this is needed: https://github.com/ruflin/Elastica/pull/1673

reedy:
Indeed. It's already needed (stuff won't work without it), but helpful to make it explicit :)

PR #1984, opened 2021-09-27
Fix phpdoc

franmomu:
This PRs basically adds `AbstractQuery` as possible value in phpdoc when creating searches. `Query::create` supports it:

https://github.com/ruflin/Elastica/blob/e646d300ccbf0a7ec49e45a09ce30e917c97e309/src/Query.php#L57-L76

I've used `inheritdoc` (it could be removed as well) instead of duplicate the phpdoc for `Index`.

### DIFF
--- a/src/Index.php
+++ b/src/Index.php
@@ -163,9 +163,9 @@ class Index implements SearchableInterface
     /**
      * Update entries in the db based on a query.
      *
-     * @param array|Query|string $query   Query object or array
-     * @param AbstractScript     $script  Script
-     * @param array              $options Optional params
+     * @param AbstractQuery|array|Collapse|Query|string|Suggest $query   Query object or array
+     * @param AbstractScript                                    $script  Script
+     * @param array                                             $options Optional params
      *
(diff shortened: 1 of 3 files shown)


thePanz:
On a sidenote: we should revisit the pattern there, maybe introducing an interface? (but that's for another PR)

franmomu:
> The expected parameter is correct, thank you for spotting it! We could even add the `Collapse` and `Suggest`: both of them are accepted by the `setQuery()` method.
> 
> The resulting method would be:
> 
> ```
>     /**
>      * @param array|Query|Query\AbstractQuery|string|Suggest|Collapse $query
>      */
>     public function setQuery($query): self
>     {
> ```
> 
> WDYT?

Sure, I've added all of them.

> On a sidenote: we should revisit the pattern there, maybe introducing an interface? (but that's for another PR)

Yep, that would be helpful, maybe also limiting the options.

thePanz (approved):

ruflin:
Thanks for the contribution!

PR #1986, opened 2021-09-30
Add changelog entry for changes since 6.1.5

reedy:
As per https://semver.org/

>MINOR version when you add functionality in a backwards compatible manner

Is adding PHP 8 support functionality? Or should this just be 6.1.6?

>PATCH version when you make backwards compatible bug fixes.


kbond:
To me, _PHP 8 support_ is a _feature_ so 6.2.0 makes sense.

thePanz:
And version `6.2.0` released

PR #1989, opened 2021-10-08
Add 6.x (>= 6.2.0) supports PHP 8.0

reedy:
Not sure if anyone really minds how we want to denote the 6.x releases before 6.2. Happy to change to match.

Will cherry pick into 7.x and 6.x when it's decided (and this lands in master)

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,9 +30,10 @@ For details on how to contribute, check the [CONTRIBUTING](https://github.com/ru
 
 Versions & Dependencies
 ------------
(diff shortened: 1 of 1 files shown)


thePanz (changes requested):
What about mentioning from which version the support of PHP 8.x is provided?

ruflin:
We could go so far to just state 5.x and older are unmaintained. Like this we don't need to keep the list of all the old releases? 2 tables also sounds finde to me.

ruflin (commented):
Current changes LGTM and +1 on the propose changes to rename to branches.

thePanz (approved):

PR #1997, opened 2021-10-19
Avoid deprecations with PHP 8.1

franmomu:
Using PHP 8.1, it triggers deprecations for implementations of some core interfaces without a return type.

Adding the return type will break BC, so [using `#[\ReturnTypeWillChange]` attribute will suppress the deprecation](https://wiki.php.net/rfc/internal_method_return_types).

I've added a job in the CI to test PHP 8.1 and removed the `composer_flags: '--ignore-platform-reqs'` option since I've guessed it was for testing `8.0` when it wasn't supported.

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -39,7 +39,8 @@ jobs:
             elasticsearch: '7.14.1'
           - php: '8.0'
             elasticsearch: '7.14.1'
-            composer_flags: '--ignore-platform-reqs'
+          - php: '8.1'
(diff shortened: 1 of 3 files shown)


ruflin:
I opened https://github.com/ruflin/Elastica/issues/2011 as a follow. Anyone interested to tackle this? :-)

AlexeyKosov:
Now that PHP 8.1 is live, is it possible to tag a new release including this update?

deguif:
@AlexeyKosov I will release a new version soon

PR #2002, opened 2021-10-21
Fix guzzlehttp/psr7 2.x compatibility

deguif:


thePanz:
We can toast to that next time we all meet :wink: 
Thank you @deguif for the PR!

PR #2006, opened 2021-10-25
fix Match class name for php8 compatibility.

egorbwork:
Change, that is fixing main compatibility problem for this old version.


egorbwork:
@ruflin our Elastic search version is 2.3.2, not advanced usage, but still vital for the application.

@reedy Ok, will try to apply this changes on this branch on weekend, will see if I will be able to finish with continuous-integration.

egorbwork:
@reedy @ruflin  I added back the Match class. And adjusted a bit composer.json for php versions. Regarding continuous integration file, I'm not sure that it will work with this version, there is local docker, which took a time for me to find adhoc solution and run tests, I received some failing with `indexExists` method from Status on php 7 version for which it was prepared prior. As the composer.json does not contains require-dev dependency for phpunit, would this code runnable for CI?

ruflin:
To set the right expectations, even if we get this working with CI and merge it into the branch, I don't plan to do any future releases on 3.x. So it will mean you will have to be using the branch and not a release tag.

On 3.x we were still running travis which is disabled by now. You have to dig into the .travis.yml to see how the testing was done there without having the dependency in the required-dev.

@egorbwork I really appreciate all the effort you put into this. I'm a bit worried that you go down a rabbit hole and keep finding issues. As pointed out by @reedy you might not only have to fix PHP 8 issues but also PHP 7 issues + CI + dependencies.

The benefit of having this PR merges is that likely some other users with PHP8 and Elastic Agent 3.x can use the change too. But how many of these are out there? One alternative for you to not have to go through all the hoops could be to just fork Elastica with the branch, apply the minimal fix you need there and use it until you upgrade your Elasticsearch version. In this case, you can even introduce breaking changes, no CI needed etc. As soon as you are able to upgrade Elasticsearch, you can switch back to Elastica.

egorbwork:
@ruflin yes, I understand. I was not planning to open pr from my fork, but it was request from colleagues, so I tried. In general I agree that for next half of year till we will migrate to last version of elastica search,  the option with fork is viable.

PR #2023, opened 2021-11-22
Extract setting gap policy to trait

stchr:
The purpose of the introduced interface is to avoid typos by using constants for options.

That said, it also fixed the typo on Serial differencing aggregation, which is/was also a typo in the official docs. See https://github.com/elastic/elasticsearch/pull/80893


ruflin:
Merged 🎉

PR #2026, opened 2021-11-24
Add processor target_field trait

deguif:


ruflin (approved):
Thanks for all this cleanup. I especially like that it reduces the duplication and lines of code we have without impacting the users in any way.

Screenshot 2021-11-24 at 11 13 06

PR #2029, opened 2021-12-01
Use array_key_first() when judicious

deguif:


src/Bulk.php
```diff
@@ -202,31 +202,21 @@ public function addData($data, ?string $opType = null)
...
-            } else {
+            if (!\is_array($row) || null === ($opType = \array_key_first($row))) {
                 throw new InvalidException('Invalid bulk data, should be array of array, Document or Bulk/Action');
             }
-        }
 
-        // add last action if available
-        if (isset($action)) {
-            $this->addAction($action);
+            if (Action::isValidOpType($opType)) {
```
ruflin, line 212: Nice rewrite. This old code was way too complicated!

src/Node/Info.php
```diff
@@ -221,9 +221,11 @@ public function refresh(array $params = []): Response
         $this->_response = $this->getNode()->getClient()->requestEndpoint($endpoint);
         $data = $this->getResponse()->getData();
 
-        $this->_data = \reset($data['nodes']);
-        $this->_id = \key($data['nodes']);
-        $this->getNode()->setId($this->getId());
+        if (null !== $nodeId = \array_key_first($data['nodes'])) {
```
ruflin, line 224: Trying to get my head around on what happened before when we didn't check furl null. 

If there are no nodes, it is expected that we keep the old values or should these be overwritten with empty values?

deguif: You're right, I didn't think this case, when nodes can be empty

ruflin:
Closing for now as didn't have any changes recently. Happy to reopen if this is still needed.

PR #2030, opened 2021-12-01
Allow to configure a subkey when adding a param

deguif:


src/Aggregation/AdjacencyMatrix.php
```diff
@@ -26,25 +23,4 @@ public function setSeparator(string $separator = '&'): self
...
-    {
-        $array = [];
-        $filters = $this->getParam('filters');
-
-        foreach ($filters as $filter) {
-            $key = \key($filter);
-            $array['adjacency_matrix']['filters'][$key] = \current($filter)->toArray();
-        }
-
-        if ($this->hasParam('separator')) {
```
ruflin, line 40: Why can we get rid of of the separator and aggs stuff too?

src/Param.php
```diff
@@ -84,9 +84,13 @@ public function setParams(array $params)
      *
      * @return $this
      */
-    public function addParam($key, $value)
+    public function addParam($key, $value, ?string $subKey = null)
```
ruflin, line 87: Nice addition!

src/Aggregation/AdjacencyMatrix.php
deguif: Yes as it will be handled by the inherited `toArray` method.
There are some tests that cover this in `AdjacencyMatrixTest`, behavior is same than before

src/Param.php
deguif: It simplifies existing code as it avoids to redefine `toArray` methods to handle this case

ruflin: It simplifies it a LOT!

ruflin (approved):

deguif:
I will add a changelog before merging this one

PR #2032, opened 2021-12-02
Use params to configure terms query

### LINKED ISSUE #2031: the post_filter is unset in Query if Terms is used post_filter
$query = new \Elastica\Query();

// set Terms as post_filter
$terms = new \Elastica\Query\Terms("key", ['term']);
$query->setPostFilter($terms);

// Query::toArray() will unset post_filter because count($terms) is 0 now, thus always ignoring the post_filter
$search->setQuery($query);
$search->search();

the _params should be set correctly in the construct
Terms:__construct()
{
    ...
    if ($terms)
        $this->setParam($this->field, $this->terms)
}

Or, the temporary workaround is to call $terms->toArray() right after the Terms is initialized.

deguif:
Fix #2031


ruflin (approved):
Nice cleanup. It not only fixes a bug but also simplifies the code!

I have the suspicion the same could be done to other classes to. We used a lot of private variables before the introduce of setParams.

PR #2033, opened 2021-12-02
Deprecate toString() in favor of __toString()

deguif:


tests/BulkTest.php
```diff
@@ -374,19 +374,22 @@ public function testRawDocumentDataRequest(): void
...
-        $bulk = new Bulk($client);
-        $bulk->addDocuments($documents);
-        $bulk->setIndex($index);
+        $bulk = (new Bulk($client))
+            ->setIndex($index)
+            ->addDocuments($documents)
+        ;
 
-        $expectedJson = '{"index":{}}
+        $expectedJson = <<<'JSON'
```
ruflin, line 382: What is the `JSON` used here for? Haven't seen this before.

ruflin (approved):
++ on getting rid of the toString() methods and instead use the magic methods.

tests/BulkTest.php
deguif: That's just HEREDOC syntax with `JSON` identifier ;)

ruflin: Ah, got it 👍

PR #2038, opened 2021-12-05
Deprecate wrong buckets_path initialization on bucket aggregations

deguif:


src/Aggregation/NormalizeAggregation.php
```diff
@@ -19,6 +19,10 @@ public function __construct(string $name, ?string $bucketsPath = null, ?string $
 
         if (null !== $bucketsPath) {
```
ruflin, line 20: It seems the top part of the constructor is always shared. Made me wonder if we should extract it.

ruflin (approved):
Change LGTM.

The comment I left can easily be a follow up PR or ignored.

src/Aggregation/NormalizeAggregation.php
deguif: Yes it's shared by a lot of pipeline aggregation, but there're some exceptions like `bucket_script` or `bucket_selector` which accepts an array of buckets paths.
A topic I have in mind for 8.0 version is to review the aggregations classes to separate them in three categories (as described in https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations.html) `pipeline`, `metric` and `bucket` as they have some difference (for example a pipeline aggregation can't have sub aggregations, ...)

PR #2043, opened 2021-12-08
Review docker-compose configuration

deguif:


docker/docker-compose.yml
```diff
@@ -1,12 +1,11 @@
 version: '3.4'
 
 # This is a docker-compose template file for running docker's containers locally.
-# It must be combined with `docker-compose.es.yml` and one of the ES specific templates like `docker-compose.es71.yml`
+# It must be combined with `docker-compose.es.yml`
 # See: docker-start unit in Makefile
 
 services:
     php:
-        container_name: elastica_php
```
ruflin, line 9: Why did you remove this one?

docker/docker-compose.es.yml
```diff
@@ -2,22 +2,21 @@ version: '3.4'
 
 services:
     es01:
-        image: docker.elastic.co/elasticsearch/elasticsearch:${ES_VERSION:-7.15.2}
-        command: >
+        image: &image docker.elastic.co/elasticsearch/elasticsearch:${ES_VERSION:-7.15.2}
```
ruflin, line 5: Yay, YAML magic :-) Have not thought of using this in docker-compose files so far but I see how it removes duplication.

docker/docker-compose.yml
thePanz: IIRC the "container name" will be inherited from the service's name
@deguif

thePanz (approved):

docker/docker-compose.yml
ruflin: Trying to remember what the reason was I added it ... If I remember correctly it had something to do with pushing the image also to dockerhub so CI could rely on it: https://hub.docker.com/r/ruflin/elastica But this is long gone and not needed anymore.

ruflin (approved):

docker/docker-compose.yml
deguif: Yes that's what I configurer project-name for docker-compose with `--project-name=elastica` in the Makefile
`container_name` is just a configuration for the runtime when the container is launched, but by setting the project name with docker-compose the container name is not related anymore to the folder where the docker-compose file is located (behavior by default of docker-compose).
That's also why I changed the `docker exec` to `docker-compose exec` to use the service name and not the underlying container name directly

Issue #2049, opened 2022-01-15
Internal versioning error when adding a document from one index to another 


pidera:
When adding a document from one index to another index, the `version` and `version_type` options are still added, which causes a validation error. 

```
Elastica\Exception\ResponseException: Validation Failed: 1: internal versioning can not be used for optimistic concurrency control. Please use `if_seq_no` and `if_primary_term` instead;
```

In #1803 these options have been removed for the `updateDocument` method, but these are still added in `addDocument`. 

I've created a separate repository to simulate the testcase:

Failing version: https://github.com/pidera/elastica-version-issue/tree/failing
Passing version: https://github.com/pidera/elastica-version-issue/tree/passing

I'll open a pull request with the fix I've come up with, let me know if this is adequate or not!

PR #2050, opened 2022-01-15
Remove version/version_type options from Index::addDocument()

pidera:
As mentioned in #2049, the version/version_type has been removed in 7.x as concurrency control.

These options were still always added in the `addDocument` method which causes an issue when adding a document from one index to another.

This PR removes those options, and sorts the other options alphabetically cfr #1803.


ruflin:
Thanks for the cleanup @pidera Any chance you could add an entry to the changelog?

Is my understanding correct that the test you added would not have passed without the change?

pidera:
@ruflin I have added the changelog entry. And indeed, the added test did not pass without the change!

ruflin:
Thanks for the contribution @pidera

PR #2056, opened 2022-02-21
Add explicit return annotation

franmomu:
When using Symfony >= 5.4, you get this kind of messages:

```
User Deprecated: Method "ArrayAccess::offsetGet()" might add "mixed" as a native return type declaration in the future. Do the same in implementation "Elastica\Multi\ResultSet" now to avoid errors or add an explicit @return annotation to suppress this message.
```

```
User Deprecated: Method "Iterator::current()" might add "mixed" as a native return type declaration in the future. Do the same in implementation "Elastica\Multi\ResultSet" now to avoid errors or add an explicit @return annotation to suppress this message.
```

This PR avoids these messages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/7.1.4...master)
 ### Backward Compatibility Breaks
 ### Added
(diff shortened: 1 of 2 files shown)


ruflin (commented):
Thanks for the cleanup.

It might seem overkill for this change as it is "only docs" but can you add an entry to the changelog? It will make it easier for others to find this change that have the same issue.

franmomu:
> It might seem overkill for this change as it is "only docs" but can you add an entry to the changelog? It will make it easier for others to find this change that have the same issue.

Yeah, no problem 👍

ruflin:
Thank you!

PR #2057, opened 2022-03-15
[6.x] No php 8.1 deprecation warnings

Hikariii:
Using PHP 8.1, it triggers deprecations for implementations of some core interfaces without a return type.

Adding the return type will break BC, so [using #[\ReturnTypeWillChange] attribute will suppress the deprecation](https://wiki.php.net/rfc/internal_method_return_types).

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -39,6 +39,8 @@ jobs:
             elasticsearch: '6.5.2'
           - php: '8.0'
             elasticsearch: '6.5.2'
+          - php: '8.1'
+            elasticsearch: '6.5.2'
       fail-fast: false
     steps:
       - name: 'Checkout'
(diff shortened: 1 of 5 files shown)


ruflin:
Thanks for the contribution. My understanding is that we only need this fix for 6.x as in master we already made the updates to the types (breaking change). It seems CI is complaining on the styling side, could you have a look?

Could you also add an entry to the changelog?

VincentLanglet:
@Hikariii the issue is with the php-cs-fixer version: https://github.com/ruflin/Elastica/blob/6.x/.github/workflows/continuous-integration.yaml#L17

It need to be bumped.

VincentLanglet:
Hi @ruflin, I fixed the cs issue in https://github.com/ruflin/Elastica/pull/2061

ruflin:
Thanks @VincentLanglet for taking over. It seems as soon as https://github.com/ruflin/Elastica/pull/2061 is in, we can close this one here.

ruflin:
Closing this one as https://github.com/ruflin/Elastica/pull/2061 was merged.

PR #2061, opened 2022-04-12
[6.x] No php 8.1 deprecation warnings

VincentLanglet:


ruflin:
Thanks @VincentLanglet for taking over the change. Any chance you could add a changelog entry?

VincentLanglet:
> Thanks @VincentLanglet for taking over the change. Any chance you could add a changelog entry?

Sure, I added one ; is it ok like this ?

ruflin:
Perfect, thanks!

VincentLanglet:
Thanks, if you have time for a release it would be perfect :)

ruflin:
Recently @thePanz has become the master of releases. Interested to take this one? 😇

ruflin:
I think I have commented too fast. There is one issue here: Elasticsearch 6.x is EOL so what does this mean for Elastica 6.x? If it is also EOL, should we still do releases? There is always a risk that we break existing deployments in an unexpected way and now have to do one more release and so on. So far it seems the last 6.x release is pretty stable :-) 

@VincentLanglet I wonder if we could also ask users that want these changes to depend on the commit hash or 6.x branch?

Lets have a quick conversation on the state of 6.x.

ruflin:
> risk to break something is really really low

I made this assumption on a 5.x release which blew up 🤦‍♂️ But I agree, I think we can take the risk and in the worst case roll it back. But lets wait on some feedback from @thePanz and others on their view. If we don't hear back, lets do it.

@VincentLanglet One thing you could do to get this already started is opening a pull request to prepare the release, something like https://github.com/ruflin/Elastica/pull/2060

VincentLanglet:
> @VincentLanglet One thing you could do to get this already started is opening a pull request to prepare the release, something like #2060

Sure https://github.com/ruflin/Elastica/pull/2062

thePanz:
> @ruflin But I agree, I think we can take the risk and in the worst case roll it back.
Yes, I agree too. Thank you for tagging and releasing it :)

Issue #1943, opened 2021-03-23
Use static code analysis


deguif:
I started a PR to use PHPStan to detect potential wrong code.
Many PRs I opened recently were related to this analysis.
This can be tracked here: https://github.com/deguif/Elastica/actions?query=branch%3Aphpstan+

Here's the current remaining errors:
```shell
------ ------------------------------------------------------------------- 
  Line   Aggregation/GeohashGrid.php                                        
 ------ ------------------------------------------------------------------- 
  48     Result of && is always false.                                      
         💡 Because the type is coming from a PHPDoc, you can turn off this  
         check by setting treatPhpDocTypesAsCertain: false in your          
         phpstan.neon.dist.                                                 
  49     Else branch is unreachable because ternary operator condition is   
         always true.                                                       
 ------ ------------------------------------------------------------------- 

 ------ ------------------------------------------------------------------- 
  Line   Bulk/Action/AbstractDocument.php                                   
 ------ ------------------------------------------------------------------- 
  124    Result of && is always false.                                      
         💡 Because the type is coming from a PHPDoc, you can turn off this  
         check by setting treatPhpDocTypesAsCertain: false in your          
         phpstan.neon.dist.                                                 
 ------ ------------------------------------------------------------------- 

 ------ ----------------------------------------------------------------------- 
  Line   Client.php                                                             
 ------ ----------------------------------------------------------------------- 
  363    Result of method Elastica\Client::_initConnections() (void) is used.   
  531    Ternary operator condition is always true.                             
  532    Ternary operator condition is always true.                             
  622    Strict comparison using === between true and array|string will always  
         evaluate to false.                                                     
 ------ ----------------------------------------------------------------------- 

 ------ ----------------------------------------------------------------------- 
  Line   ClientConfiguration.php                                                
 ------ ----------------------------------------------------------------------- 
  53     Unsafe usage of new static().                                          
         💡 See:                                                                 
         https://phpstan.org/blog/solving-phpstan-error-unsafe-usage-of-new-st  
         atic                                                                   
  83     Unsafe usage of new static().                                          
         💡 See:                                                                 
         https://phpstan.org/blog/solving-phpstan-error-unsafe-usage-of-new-st  
         atic                                                                   
  85     Parameter #1 $dsn of static method                                     
         Elastica\ClientConfiguration::parseDsn() expects                       
         Nyholm\Dsn\Configuration\Url,                                          
[...]

PR #2064, opened 2022-05-18
Add PHPStan

franmomu:
This PR tries to add PHPStan at level 3 that doesn't require many changes and I think it's a good start. The issues ignored are mostly about accessing dynamic fields (from `Document`) and through `Facade`.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,0 +1,102 @@
+parameters:
+	ignoreErrors:
+		-
+			message: "#^Method Elastica\\\\QueryBuilder\\:\\:aggregation\\(\\) should return Elastica\\\\QueryBuilder\\\\DSL\\\\Aggregation but returns Elastica\\\\QueryBuilder\\\\Facade\\.$#"
+			count: 1
+			path: src/QueryBuilder.php
+
+		-
+			message: "#^Method Elastica\\\\QueryBuilder\\:\\:collapse\\(\\) should return Elastica\\\\QueryBuilder\\\\DSL\\\\Collapse but returns Elastica\\\\QueryBuilder\\\\Facade\\.$#"
+			count: 1
+			path: src/QueryBuilder.php
+
+		-
+			message: "#^Method Elastica\\\\QueryBuilder\\:\\:query\\(\\) should return Elastica\\\\QueryBuilder\\\\DSL\\\\Query but returns Elastica\\\\QueryBuilder\\\\Facade\\.$#"
+			count: 1
+			path: src/QueryBuilder.php
+
+		-
+			message: "#^Method Elastica\\\\QueryBuilder\\:\\:suggest\\(\\) should return Elastica\\\\QueryBuilder\\\\DSL\\\\Suggest but returns Elastica\\\\QueryBuilder\\\\Facade\\.$#"
+			count: 1
+			path: src/QueryBuilder.php
+
+		-
+			message: "#^Function GuzzleHttp\\\\Psr7\\\\stream_for not found\\.$#"
+			count: 1
+			path: src/Transport/Guzzle.php
+
+		-
+			message: "#^Access to an undefined property Elastica\\\\Document\\:\\:\\$counter\\.$#"
+			count: 1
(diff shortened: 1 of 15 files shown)


ruflin:
I'm definitively in support of adding these kind of automation to improve the code quality. Don't really have any experience with PHPStan but maybe @deguif or @thePanz can have a more detailed look?

phpstan-baseline.neon
```diff
@@ -0,0 +1,102 @@
...
+			count: 1
+			path: src/QueryBuilder.php
+
+		-
+			message: "#^Method Elastica\\\\QueryBuilder\\:\\:suggest\\(\\) should return Elastica\\\\QueryBuilder\\\\DSL\\\\Suggest but returns Elastica\\\\QueryBuilder\\\\Facade\\.$#"
+			count: 1
+			path: src/QueryBuilder.php
+
+		-
+			message: "#^Function GuzzleHttp\\\\Psr7\\\\stream_for not found\\.$#"
```
thePanz, line 24: Is this caused by a missing library, or a Guzzle version mismatch?

phpstan.neon
```diff
@@ -0,0 +1,15 @@
...
+    - phpstan-baseline.neon
+
+parameters:
+    level: 3
+    paths:
+        - src
+        - tests
+    excludePaths:
+        - src/Transport/HttpAdapter.php
+        - src/Query/Match.php
```
thePanz, line 12: Are those too complex or hard to fix to pass the level=3 of PhpStan?

thePanz (approved):

thePanz:
Nice addition, gracias @franmomu!
There are a couple of exceptions, that would need a clarification, but LGTM!
This is a good first step to have Phpstan, I still fear that increasing to higher levels will be quite hard due to the dynamic fields and the mixed parameters we have for some properties

phpstan-baseline.neon
franmomu: because version mismatch, `stream_for` does not exists in the installed version (I haven't check if the call could be removed, I'll try later)

phpstan.neon
franmomu: `Match` is because of the name of the class is a reserved keyword with PHP 8+, so there is a syntax error for PHPStan.
And [`src/Transport/HttpAdapter.php`](https://github.com/ruflin/Elastica/blob/72a4598544e3f99b5dd8cacb05d009ee75c2a701/src/Transport/HttpAdapter.php) uses `egeloen/http-adapter` which is not installed so there are quite errors about unknown classes.

(both of them are deprecated)

franmomu:
> Nice addition, gracias @franmomu! There are a couple of exceptions, that would need a clarification, but LGTM! This is a good first step to have Phpstan, I still fear that increasing to higher levels will be quite hard due to the dynamic fields and the mixed parameters we have for some properties

de nada! 😉 Yeah, higher levels will need more work, we'll see when we get there

phpstan-baseline.neon
franmomu: Checking this, we only have `guzzlehttp/guzzle` in `require-dev`, we don't have any `conflict` restricting versions used, so right now I think it should be fine ignoring the error, I'll open an issue about this.

deguif:
Related issue: https://github.com/ruflin/Elastica/issues/1943

ruflin:
@thePanz Feel free to merge it in if you think it is ready.

@franmomu It would be nice to add an entry in the changelog around this addition even though it is not affecting the lib itself.

thePanz (approved):

thePanz:
@deguif I lost track of your PR, sorry for that.
Should we merge this one, and rebase yours? wdyt?

deguif:
@thePanz I didn't open a PR ;)
I just do some small PRs to fix some PHPStan warnings a while ago.
Yes, we can merge this one.

thePanz:
And merged @franmomu (\cc @ruflin @deguif )

phpstan-baseline.neon
thePanz: For reference: #2066

PR #2066, opened 2022-05-20
Add conflict with some guzzle versions

### LINKED ISSUE #2065: Add `conflict` section in composer
IMO there should be a `conflict` section in sync with `require-dev` (with `aws/aws-sdk-php` and `guzzlehttp/guzzle`):

https://github.com/ruflin/Elastica/blob/72a4598544e3f99b5dd8cacb05d009ee75c2a701/composer.json#L26-L31

So we only allow to installed tested versions.

franmomu:
Closes https://github.com/ruflin/Elastica/issues/2065

The thing is that we are testing `"guzzlehttp/guzzle": "^6.3 || ^7.2"`, but a user right now could for example use version `4` and will probably get some errors.

Adding a conflict allow us to be able to restrict which versions of the library we are supporting.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,6 @@
     },
     "suggest": {
         "aws/aws-sdk-php": "Allow using IAM authentication with Amazon ElasticSearch Service",
-        "egeloen/http-adapter": "Allow using httpadapter as transport",
         "guzzlehttp/guzzle": "Allow using guzzle as transport",
         "monolog/monolog": "Logging request"
     },
--- a/composer.json
+++ b/composer.json
@@ -23,6 +23,9 @@
         "symfony/deprecation-contracts": "^2.2 || ^3.0",
         "symfony/polyfill-php73": "^1.19"
     },
+    "conflict": {
+        "guzzlehttp/guzzle": "<6.3 || >=7.0,<7.2"
+    },
     "require-dev": {
         "aws/aws-sdk-php": "^3.155",
         "guzzlehttp/guzzle": "^6.3 || ^7.2",


composer.json
```diff
@@ -31,7 +34,6 @@
     },
     "suggest": {
         "aws/aws-sdk-php": "Allow using IAM authentication with Amazon ElasticSearch Service",
-        "egeloen/http-adapter": "Allow using httpadapter as transport",
```
ruflin, line 34: Was this removed on purpose?

ruflin (commented):
Thanks for the PR, this makes it much simpler to discuss.

One thing I'm curious: Lets assume someone uses Elastica and a guzzle version 4.0. But all the features this users is using in Elastica don't use guzzle. Now a user upgrading Elastica would suddenly have a theoretical conflict but none in practice?

franmomu:
> One thing I'm curious: Lets assume someone uses Elastica and a guzzle version 4.0. But all the features this users is using in Elastica don't use guzzle. Now a user upgrading Elastica would suddenly have a theoretical conflict but none in practice?

Yeah, that's the drawback of this approach 😞 I guess that is somehow reasonable if the range of versions makes sense, for example guzzle 6.3 was released 5 years ago, so I guess that having a lower bound of 6.3 should be fine for most people.

That's the thing with optional dependencies, I guess the best would be to have another package like `elastica-guzzle` or `elastica-aws`, but that comes with more maintenance work... which is not always desirable.

composer.json
franmomu: yeah, I did it in a separated commit https://github.com/ruflin/Elastica/pull/2066/commits/278ce0dcd628a411caade584c39cabffc8ebb55b (the project is abandoned), but it could be done in another PR.

ruflin: Removal SGTM, can we move it to a separate PR? Sounds like it is not related to the guzzle discussion.

ruflin:
Thanks for the additional details. The conflict sounds too intrusive to me. We create pain for some users that would be fine using it the way it is by trying to protect some other users from shooting themself into the foot, at least this is my understanding.

The runtime check proposed by @deguif sounds less intrusive.

But then I still stumble over:

> for example guzzle 6.3 was released 5 years ago, so I guess that having a lower bound of 6.3 should be fine for most people.

Why do we need this check in the first place? Are we trying to fix an actual issue or are we introducing additional checks for a theoretical edge case? @franmomu Can you share a bit more background here?

franmomu:
The idea behind this (it started in https://github.com/ruflin/Elastica/pull/2064#pullrequestreview-979619281) was to be able to add and remove compatibility with guzzle versions without creating BC breaks (in minor versions).

At some point we would want to remove support of guzzle 6, to do so I think the only way right now is to create a major version. 

Adding a conflict at composer level allow us to prevent/allow users to only use the supported versions of guzzle (with the drawback commented in https://github.com/ruflin/Elastica/pull/2066#pullrequestreview-981272873) and change those versions in a minor release.

A runtime check it's also valid in optional dependencies with the drawback that we should keep compatibility with old versions of guzzle until we release a major version.

At the end is about trade-offs.

ruflin:
As you stated @franmomu , its all about tradeoffs. 

`guzzle` is not a required but a suggested dependency. Breaking some users setup for a suggested dependency does not sound great even if the lib the users are using is very outdated. 

What holds me back from doing any change at all is that it is not an issue that was reported so far or at least I have not seen it. Even the runtime check is nice, but do we need the additional code? 

There is still the question around updating the dependencies only for majors. Good news is that we have a major release upcoming :-) At the same time I wonder if might be able to get rid of the suggested dependency to get rid of the problem?

franmomu:
I don't have enough knowledge of this package, but maybe we could use the`Client` from `elastic/elasticsearch-php` in next major version (I guess it would require quite work and BC breaks)

ruflin:
There are some different opinions around using elastic/elasticsearch-php as the client. I'm personally all in favour of using it and internally there are more and more places where it is used.

ruflin:
Closing this PR as there hasn't been any change recently but happy to reopen and continue the conversation.

PR #932, opened 2015-09-16
It's me again :) http.compression more effective !

yvann:
Moved the "accept-encoding" part to be always effective, the compression will depend of the ES configuration, we only say that we support it.

PR #2068, opened 2022-05-21
Remove 0 as request data

franmomu:
Checking errors found by PHPStan, there is this one, here the thing is that `Request::getData()` is supposed to return an `array`:

https://github.com/ruflin/Elastica/blob/72a4598544e3f99b5dd8cacb05d009ee75c2a701/src/Request.php#L84-L92

This was added in https://github.com/ruflin/Elastica/pull/640 (8 years ago) and the test was removed in https://github.com/ruflin/Elastica/pull/932.

So I don't know if the `0` is a possible value so we should modify the return type of `Request::getData()` or this can be removed.


ruflin:
@Tobion @leabaertschi I know https://github.com/ruflin/Elastica/pull/640 is years ago but do you remember by chance why we needed this and if it still applies today?

leabaertschi:
Sorry, I don't remember, I haven't worked with elasticseach in years.

ruflin:
I did a quick test. I think the following was working in old versions of Elasticsearch for analyze: https://www.elastic.co/guide/en/elasticsearch/reference/1.7/indices-analyze.html

```
curl -X GET "http://elastic:changeme@localhost:9200/_analyze?pretty" -H 'Content-Type: application/json' -d'0'
```

Looking at the current docs, it seems it is always an object that is expected: https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-analyze.html

Lets move forward with this. 

@franmomu Can you add a changelog entry? Thanks for the cleanup.

franmomu:
done, thanks for checking into this!

thePanz:
Merged @franmomu , thanks!

PR #2070, opened 2022-05-24
Add `null` to some PHPDoc types

franmomu:
This PR fixes some issues found by PHPStan raising level to 4, this include the fixes of @deguif in https://github.com/deguif/Elastica/commit/c74d4f0d226d9f14c64a2c068dd286f55d07af67

This first commit is only about extracting the variable assignment out of the logical statement.

### DIFF
--- a/src/AbstractUpdateAction.php
+++ b/src/AbstractUpdateAction.php
@@ -307,9 +307,9 @@ class AbstractUpdateAction extends Param
      */
     public function setRefresh($refresh = true)
     {
-        \is_bool($refresh) && $refresh = $refresh
-            ? Reindex::REFRESH_TRUE
-            : Reindex::REFRESH_FALSE;
+        if (\is_bool($refresh)) {
+            $refresh = $refresh ? Reindex::REFRESH_TRUE : Reindex::REFRESH_FALSE;
+        }
 
         return $this->setParam(Reindex::REFRESH, $refresh);
     }
--- a/src/AbstractUpdateAction.php
+++ b/src/AbstractUpdateAction.php
@@ -303,6 +303,8 @@ class AbstractUpdateAction extends Param
     /**
      * @param bool|string $refresh
      *
+     * @phpstan-param bool|Reindex::REFRESH_* $refresh
+     *
      * @return $this
      */
     public function setRefresh($refresh = true)
--- a/src/Reindex.php
+++ b/src/Reindex.php
@@ -75,7 +75,9 @@ class Reindex extends Param
 
     public function setWaitForCompletion($value): void
     {
-        \is_bool($value) && $value = $value ? 'true' : 'false';
+        if (\is_bool($value)) {
+            $value = $value ? 'true' : 'false';
+        }
 
         $this->setParam(self::WAIT_FOR_COMPLETION, $value);
     }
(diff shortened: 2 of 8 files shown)


src/AbstractUpdateAction.php
```diff
@@ -307,9 +307,9 @@ public function hasRetryOnConflict()
      */
     public function setRefresh($refresh = true)
     {
-        \is_bool($refresh) && $refresh = $refresh
-            ? Reindex::REFRESH_TRUE
-            : Reindex::REFRESH_FALSE;
+        if (\is_bool($refresh)) {
+            $refresh = $refresh ? Reindex::REFRESH_TRUE : Reindex::REFRESH_FALSE;
```
thePanz, line 311: We should try with something like `@param Reindex::REFRESH_*|bool $refresh`, as the method signature.
can you check it it works? :)

franmomu: yeah sure, I've used `@phpstan-param` because maybe some tools/IDEs do not understand that param.

src/Reindex.php
```diff
@@ -73,9 +73,16 @@ public function run(): Response
...
+    /**
+     * @param bool|string $value
+     *
+     * @phpstan-param bool|self::WAIT_FOR_COMPLETION_* $value
+     */
     public function setWaitForCompletion($value): void
     {
-        \is_bool($value) && $value = $value ? 'true' : 'false';
+        if (\is_bool($value)) {
+            $value = $value ? 'true' : 'false';
```
thePanz, line 84: From the ES documentation, this parameter is a boolean, no need  to use strings as parameters.
(while we still need to use "true" or "false" as strings in `->setParam()`
Would you adjust it? in that case the  `self::WAIT_FOR_COMPLETION_FALSE` can be removed (if not used)

thePanz (changes requested):

src/Reindex.php
thePanz: Should we trigger an user warning in case non-boolean are passed here?
Like an user_error or something like that?

src/Reindex.php
```diff
@@ -73,9 +77,14 @@ public function run(): Response
...
 
+    /**
+     * @param bool $value
+     */
     public function setWaitForCompletion($value): void
     {
-        \is_bool($value) && $value = $value ? 'true' : 'false';
+        if (\is_bool($value)) {
+            $value = $value ? 'true' : 'false';
+        }
```
deguif, line 87:
```suggestion
        } else {
            \trigger_deprecation('ruflin/elastica', '7.1.6', 'Passing anything else than a boolean as 1st argument to "%s()" is deprecated, pass a boolean instead. It will be removed in 8.0.', __METHOD__);
        }
```
As @thePanz suggested a deprecation like this would allow to type-hint more strictly the parameter in next major version.

franmomu: Sure! thanks for the suggestion. It would be nice to be able to add missing parameter type declarations in PHPDoc, so for the next major release we can move them to the method definition.

thePanz (approved):

thePanz:
Merged @franmomu , thank you for your contribution! :tada: 
and @deguif too :+1:

PR #2072, opened 2022-06-03
Add coverage to CI

franmomu:


ruflin:
+1 on adding this. The concern I have is on what diff it will show for each PR? Will it just compare the lines changed or will now each PR have a very lengthy list like this PR does? We do not have 100% coverage and I don't think it is the goal, it is more about good coverage (whatever that means).

ruflin:
That is what I'm hoping for. The odd part in this PR is that I see all these:

Screenshot 2022-06-07 at 15 18 08

I'm now wondering if this is a beta feature I have enabled or everyone is seeing this under "Files changed"?

ruflin:
Lets try it out, we can always disable it if needed. We had coverage reporting in the past, not sure when I got lost. There is even still the badge in the README.md file linked to https://codecov.io/github/ruflin/Elastica?branch=master

PR #2073, opened 2022-06-03
Fix some PHPStan findings in tests

franmomu:
`GuzzleException::getGuzzleException()` always returns a `TransferException`:

https://github.com/ruflin/Elastica/blob/4087569dfe7e0c8a24470fa22532c7a69bb86343/src/Exception/Connection/GuzzleException.php#L18-L41

PR #2075, opened 2022-06-07
Exclude Match.php from coverage

franmomu:
Follow-up of https://github.com/ruflin/Elastica/pull/2072

See https://github.com/ruflin/Elastica/pull/2073#issuecomment-1148717638

Let's see if it is generated properly

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -12,6 +12,9 @@
   <filter>
     <whitelist>
       <directory>src/</directory>
+      <exclude>
+        <file>src/Query/Match.php</file>
+      </exclude>
     </whitelist>
   </filter>
   <listeners>


ruflin:
Looks like this fixes the coverage reporting? It is still a bit odd that we have to exclude `Match.php` from the testing for this, I would rather just exclude it from the reporting :-(

In case someone else is looking for the error that was happening on CI before:

```
Generating code coverage report in Clover XML format ... failed [00:00.072]
Cannot parse /home/runner/work/Elastica/Elastica/src/Query/Match.php: Syntax error, unexpected T_PAAMAYIM_NEKUDOTAYIM, expecting ':' on line 5
```

I guess it is not happy with this line: https://github.com/ruflin/Elastica/blob/master/src/Query/Match.php#L5

franmomu:
> Looks like this fixes the coverage reporting? It is still a bit odd that we have to exclude `Match.php` from the testing for this, I would rather just exclude it from the reporting :-(

Yeah it excludes the file only from the reporting, see https://phpunit.readthedocs.io/en/8.5/configuration.html#the-filter-element

> In case someone else is looking for the error that was happening on CI before:
> 
> ```
> Generating code coverage report in Clover XML format ... failed [00:00.072]
> Cannot parse /home/runner/work/Elastica/Elastica/src/Query/Match.php: Syntax error, unexpected T_PAAMAYIM_NEKUDOTAYIM, expecting ':' on line 5
> ```
> 
> I guess it is not happy with this line: https://github.com/ruflin/Elastica/blob/master/src/Query/Match.php#L5

`Match` is a reserved word since PHP 8, so the parser is not able to understand `Match::`, we also had to exclude it in PHPStan

https://github.com/ruflin/Elastica/blob/583bc09dcc08e1b783076ea91ec73656e02db9da/phpstan.neon#L12

ruflin:
Perfect, thanks. I didn't realise it only is the filter for the coverage! Great to have coverage reports back again.

ruflin:
https://codecov.io/github/ruflin/Elastica?branch=master !

PR #2080, opened 2022-06-20
PHPStan level 4

franmomu:
So this PR increases PHPstan level and sets `treatPhpDocTypesAsCertain` as `false`, this is because there are several errors like:

```
 ------ --------------------------------------------------------------------------------------------------------------------------------------------
  Line   src/Aggregation/GeohashGrid.php
 ------ --------------------------------------------------------------------------------------------------------------------------------------------
  48     Result of && is always false.
         💡 Because the type is coming from a PHPDoc, you can turn off this check by setting treatPhpDocTypesAsCertain: false in your phpstan.neon.
  49     Else branch is unreachable because ternary operator condition is always true.
 ------ --------------------------------------------------------------------------------------------------------------------------------------------
```

but the code is fine:

https://github.com/ruflin/Elastica/blob/ea9fbd6740139d6045c201a6b75d061289009b80/src/Aggregation/GeohashGrid.php#L42-L50

About the ignored errors (all in `tests`), most of them are because of `TestCase::markTestSkipped()` call, so it warns about the code after that call not being executed.

### DIFF
--- a/tests/Collapse/CollapseTest.php
+++ b/tests/Collapse/CollapseTest.php
@@ -112,10 +112,12 @@ class CollapseTest extends BaseTest
         $collapse->setInnerHits($innerHits1);
         $collapse->addInnerHits($innerHits2);
 
-        $this->assertCount(2, $collapse->getParam('inner_hits'));
-        $this->assertIsArray($collapse->getParam('inner_hits'));
-        $this->assertEquals($innerHits1, $collapse->getParam('inner_hits')[0]);
-        $this->assertEquals($innerHits2, $collapse->getParam('inner_hits')[1]);
+        $innerHits = $collapse->getParam('inner_hits');
+
+        $this->assertCount(2, $innerHits);
+        $this->assertIsArray($innerHits);
+        $this->assertEquals($innerHits1, $innerHits[0]);
+        $this->assertEquals($innerHits2, $innerHits[1]);
 
         $innerHitsOverride = new InnerHits();
         $innerHitsOverride->setName('override');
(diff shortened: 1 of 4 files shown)


tests/Collapse/CollapseTest.php
```diff
@@ -112,10 +112,12 @@ public function testSetInnerHitsOverridesExistingValue(): void
...
-        $this->assertCount(2, $collapse->getParam('inner_hits'));
-        $this->assertIsArray($collapse->getParam('inner_hits'));
-        $this->assertEquals($innerHits1, $collapse->getParam('inner_hits')[0]);
-        $this->assertEquals($innerHits2, $collapse->getParam('inner_hits')[1]);
+        $innerHits = $collapse->getParam('inner_hits');
+
+        $this->assertCount(2, $innerHits);
+        $this->assertIsArray($innerHits);
+        $this->assertEquals($innerHits1, $innerHits[0]);
+        $this->assertEquals($innerHits2, $innerHits[1]);
```
franmomu, line 120: This has been changed to not confuse PHPStan, because below there is this call:

```php
$this->assertInstanceOf(InnerHits::class, $collapse->getParam('inner_hits'));
```
So without this change, this was the error:

```
 ------ -------------------------------------------------------------------------------------------------------------------------------------------
  Line   tests/Collapse/CollapseTest.php
 ------ -------------------------------------------------------------------------------------------------------------------------------------------
  125    Call to method PHPUnit\Framework\Assert::assertInstanceOf() with 'Elastica\\Collapse\\InnerHits' and array will always evaluate to false.
 ------ -------------------------------------------------------------------------------------------------------------------------------------------
```

ruflin (approved):

ruflin:
There is one build failing with:

```
Run vendor/bin/phpstan analyse --no-progress --error-format=github
Note: Using configuration file /home/runner/work/Elastica/Elastica/phpstan.neon.
Error: Expression on left side of ?? is not nullable.
Error: Process completed with exit code 1.
```

It seems it is also marked in the review. Not sure why?


Excited to see that we are climbing the phpstan levels :-) Thanks for getting us here.

franmomu:
Oops I missed that one, that is because in:

https://github.com/elastic/elasticsearch-php/blob/24c51c05103bf43fb9847acab039cba1d6e43173/src/Elasticsearch/Endpoints/AbstractEndpoint.php#L54-L57

there is no `null` as possible value. Since they released `8.x` and there is no `body` property anymore, not sure if they would create a release for this change.

PR #2082, opened 2022-06-21
Relax `value` type in `SetProcessor` to allow `mixed`

franmomu:
It is defined as `string`, but in `tests` there are a couple of places where another type is used:

https://github.com/ruflin/Elastica/blob/732ccf3874c4b77861c88ce0f1e92bc40cf4e119/tests/Processor/SetProcessorTest.php#L20

https://github.com/ruflin/Elastica/blob/8ecc5cdadd5bf0b69c1eb20e56246a940b43bdb7/tests/PipelineTest.php#L26

Looking at the documentation: https://www.elastic.co/guide/en/elasticsearch/reference/7.17/set-processor.html#set-processor looks like any type can be used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/7.1.5...master)
 ### Backward Compatibility Breaks
+* Changed `SetProcessor::setValue` signature to allow to pass any type, if you are overriding this method you must update the signature removing the `string` type-hint.
 ### Added
 * Added `PHPStan` at level 3 by @franmomu [#2064](https://github.com/ruflin/Elastica/pull/2064)
 * Added coverage check to CI by @franmomu [#2071](https://github.com/ruflin/Elastica/pull/2071)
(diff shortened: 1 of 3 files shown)


src/Processor/SetProcessor.php
```diff
@@ -25,9 +28,11 @@ public function __construct(string $field, string $value)
     /**
      * Set field value.
      *
+     * @param mixed $value
+     *
      * @return $this
      */
-    public function setValue(string $value): self
+    public function setValue($value): self
```
deguif, line 35: That's would be a BC break for anyone who reimplemented this method and kept the `string $value` signature.
But it should be very limited as extending this processor is probably not widespread at all.

ruflin: Agree, but I also hope this is very rare. I suggest we mark it as a breaking change in the changelog with a note on when it would be breaking (someone inheriting SetProcessor).

It is one of these scenarios I think where in theory it is a breaking change and I'm pretty sure someone is going to bring up that because of that we (I) don't follow semver but working around it would not bring benefits to the large majority of users. I'm +1 on the change.

franmomu: I'll update the PR later adding it in the BC Break section, it wasn't intentional, I changed it in the constructor which is not BC break and then changed it in the `setValue` without thinking about that.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/7.1.5...master)
 ### Backward Compatibility Breaks
+* Removed `string` parameter type-hint to `value` in `SetProcessor::setValue`.
```
ruflin, line 9: Can you expand this a bit more and explain in which scenarios it is breaking? Basically only if someone inherits from SetProcessor

PR #2083, opened 2022-06-22
ExceptionInterface should extend Throwable

b1rdex:


deguif (approved):
Thanks @b1rdex 
Could you add a changelog entry?

PR #2085, opened 2022-06-22
Relax `Settings::setMergePolicy` value to allow int

franmomu:
Ref: https://github.com/ruflin/Elastica/pull/2082#issuecomment-1163082427

`Setting::setMergePolicy()` second parameter:

https://github.com/ruflin/Elastica/blob/ff1a6e0bb8c241834302d4028c070f6a781ae154/src/Index/Settings.php#L262-L266

It expects a `string`, but in `tests`:

https://github.com/ruflin/Elastica/blob/4321f37083349c73faf7c4c7d728d4944af47795/tests/Index/SettingsTest.php#L239-L243

and looking at https://github.com/elastic/elasticsearch/blob/56d89577a033c15b518e2d74e55d9605eb90b78a/server/src/main/java/org/elasticsearch/index/MergePolicyConfig.java

seems like it could accept at least `int` as well based on the examples: https://github.com/elastic/elasticsearch/blob/56d89577a033c15b518e2d74e55d9605eb90b78a/server/src/main/java/org/elasticsearch/index/MergePolicyConfig.java#L39-L80

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/7.1.5...master)
 ### Backward Compatibility Breaks
 * Changed `SetProcessor::setValue` signature to allow to pass any type, if you are overriding this method you must update the signature removing the `string` type-hint.
+* Changed `Settings::setMergePolicy` signature to allow to pass `int` and `string` as argument 2, if you are overriding this method you must update the signature removing the `string` type-hint.
 ### Added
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/7.1.5...master)
 ### Backward Compatibility Breaks
+* Changed `Settings::setMergePolicy` signature to allow to pass `int` and `string` as argument 2, if you are overriding this method you must update the signature removing the `string` type-hint.
```
ruflin, line 9: Just resolved the merge conflict but now looking at it again, I see the links to PR's are missing. Can you add it by chance?

franmomu: sure, I've also added some other I missed

PR #2087, opened 2022-06-24
Fix PHPStorm findings

franmomu:
This doesn't fix any PHPStan issue, but some PHPStorm findings.

The PHPDoc code that has been removed is because it didn't match the method signature and once that has been fixed, `php-cs-fixer` removes it.

I'll add some a couple of comments in the code about the PHPDoc changes.


src/Bulk/Action.php
```diff
@@ -34,7 +34,7 @@ class Action
     protected $_metadata = [];
 
     /**
-     * @var array
+     * @var array|string
```
franmomu, line 37: Based on https://github.com/ruflin/Elastica/blob/1e746078f4ee6a492f67d593c387ffaddd102df7/src/Bulk/Action.php#L107-L125

deguif (approved):

PR #2089, opened 2022-06-24
Allow `int` and `float` for `FunctionScore::addDecayFunction()` origin and scale

franmomu:
Here we have the same situation as before about types, in this case `origin` and `scale` are required as ´string´, but in `tests` there are some calls with `int` and `float`

https://github.com/ruflin/Elastica/blob/a0a2dc27f4d8ef633893de6edf088b3c922f3767/tests/Query/FunctionScoreTest.php#L26-L33

https://github.com/ruflin/Elastica/blob/a0a2dc27f4d8ef633893de6edf088b3c922f3767/tests/Query/FunctionScoreTest.php#L130

Apparently in the documentation those values are allowed: https://www.elastic.co/guide/en/elasticsearch/reference/7.17/query-dsl-function-score-query.html#function-decay

My concern here is if this class is more likely to be inherited and override that method than the previous ones (`Settings` and `SetProcessor`), so maybe we could update the PHPDoc (and add a reminder) and ignore the PHPStan issue and in next major version allow those types apart from `string`.


ruflin (approved):
Change LGTM, happy to get it merged. As you mentioned, it goes to our continuing discussion around what we should do with these and also around branching. As soon as we have a 7.x branch (or the other way around) we are able to not get these changes into 7.x but 8.x and then it should be less of a problem. 

Should we get this in as is, follow up with branching and then make the partially breaking changes?

PR #2090, opened 2022-06-24
Sanitize host name for AWS requests before signing in AWSAuthV4 trans…

jhuebner79:
...port

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Bumped `elasticsearch/elasticsearch` to `7.10` to be able to use `OpenPointInTime` class [#2113](https://github.com/ruflin/Elastica/pull/2113)
 * Updated `php-cs-fixer` to `3.9.5` [#2110](https://github.com/ruflin/Elastica/pull/2110)
 * Changed `Elastica\Index\Settings::get` adding ability to get default settings by @krasilnikovm [#2115](https://github.com/ruflin/Elastica/pull/2115)
+* Update `AWSAuthV4 transport` to sanitize host name for AWS requests before signing [#2090](https://github.com/ruflin/Elastica/pull/2090)
 ### Deprecated
 ### Removed
 * Removed `CallbackStrategyTestHelper` and `ErrorsCollector` from `tests` [#2111](https://github.com/ruflin/Elastica/pull/2111)
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -135,3 +135,7 @@ parameters:
 			count: 1
 			path: tests/SnapshotTest.php
 
+		-
+			message: "#^Function GuzzleHttp\\\\Psr7\\\\modify_request not found\\.$#"
+			count: 1
+			path: src/Transport/AwsAuthV4.php
--- a/src/Transport/AwsAuthV4.php
+++ b/src/Transport/AwsAuthV4.php
@@ -10,6 +10,7 @@ use GuzzleHttp;
 use GuzzleHttp\Client;
 use GuzzleHttp\HandlerStack;
 use GuzzleHttp\Middleware;
+use GuzzleHttp\Psr7;
 use Psr\Http\Message\RequestInterface;
 
 class AwsAuthV4 extends Guzzle
@@ -42,13 +43,31 @@ class AwsAuthV4 extends Guzzle
             : \getenv('AWS_REGION');
         $signer = new SignatureV4('es', $region);
         $credProvider = $this->getCredentialProvider();
+        $transport = $this;
 
         return Middleware::mapRequest(static function (RequestInterface $req) use (
             $signer,
-            $credProvider
+            $credProvider,
+            $transport
         ) {
-            return $signer->signRequest($req, $credProvider()->wait());
+            return $signer->signRequest($transport->sanitizeRequest($req), $credProvider()->wait());
         });
+    }
+
+    private function sanitizeRequest(RequestInterface $request): RequestInterface
+    {
+        // Trailing dots are valid parts of DNS host names (see RFC 1034),
+        // but interferes with header signing where AWS expects a stripped host name.
+        if ('.' === \substr($request->getHeader('host')[0], -1)) {
+            $changes = ['set_headers' => ['host' => \rtrim($request->getHeader('host')[0], '.')]];
+            if (\class_exists(Psr7\Utils::class)) {
+                $request = Psr7\Utils::modifyRequest($request, $changes);
+            } else {
+                $request = Psr7\modify_request($request, $changes);
+            }
+        }
+
+        return $request;
     }
 
     private function getCredentialProvider(): callable
--- a/tests/Transport/AwsAuthV4Test.php
+++ b/tests/Transport/AwsAuthV4Test.php
@@ -205,4 +205,35 @@ class AwsAuthV4Test extends GuzzleTest
             );
         }
     }
+
+    /**
+     * @group unit
+     * @depends testSignsWithProvidedCredentials
+     */
+    public function testStripsTrailingDotInHost(): void
+    {
+        $host = $this->_getHost();
+        $hostWithTrailingDot = $host.'.';
+
+        $config = [
+            'persistent' => false,
+            'transport' => 'AwsAuthV4',
+            'aws_access_key_id' => 'foo',
+            'aws_secret_access_key' => 'bar',
+            'aws_session_token' => 'baz',
+            'aws_region' => 'us-east-1',
+            'host' => $hostWithTrailingDot,
+        ];
+        $client = $this->_getClient($config);
+
+        try {
+            $client->request('_stats');
+        } catch (GuzzleException $e) {
+            $guzzleException = $e->getGuzzleException();
+            $this->assertInstanceOf(ConnectException::class, $guzzleException);
+            $request = $guzzleException->getRequest();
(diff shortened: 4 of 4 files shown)


ruflin:
The change LGTM. Thanks @jhuebner79 . As I don't have any AWS setup and it is not part of the test suite, I don't really have a way to test this. Did you by chance run this code with one of your setups?

jhuebner79:
> The change LGTM. Thanks @jhuebner79 . As I don't have any AWS setup and it is not part of the test suite, I don't really have a way to test this. Did you by chance run this code with one of your setups?

Yes, we were able to reference my branch in our `composer.json` - it resolves our issue 👍

src/Transport/AwsAuthV4.php
```diff
@@ -42,15 +43,28 @@ private function getSigningMiddleware(): callable
...
+            return $signer->signRequest($transport->sanitizeRequest($req), $credProvider()->wait());
         });
     }
 
+    private function sanitizeRequest(RequestInterface $request): RequestInterface
+    {
+        // Trailing dots are valid parts of DNS host names (see RFC 1034),
+        // but interferes with header signing where AWS expects a stripped host name.
+        if ('.' === \substr($request->getHeader('host')[0], -1)) {
+            $request = Utils::modifyRequest($request, ['set_headers' => ['host' => \rtrim($request->getHeader('host')[0], '.')]]);
```
franmomu, line 62: Based on 

https://github.com/ruflin/Elastica/blob/7e4de1eebdcbc5a4731fda61e098aaf40a8d0258/src/Transport/Guzzle.php#L207-L210

Would this work with guzzle 6? I guess the same check should be one here.

jhuebner79: It would work with Guzzle 6.3, but not with earlier versions as far as I can see.

Would something like

```
private function sanitizeRequest(RequestInterface $request): RequestInterface
    {
        // Trailing dots are valid parts of DNS host names (see RFC 1034),
        // but interferes with header signing where AWS expects a stripped host name.
        if ('.' === \substr($request->getHeader('host')[0], -1)) {
            $changes = ['set_headers' => ['host' => \rtrim($request->getHeader('host')[0], '.')]];
            if (\class_exists(Psr7\Utils::class)) {
                $request = Psr7\Utils::modifyRequest($request, $changes);
            } else {
                $request = Psr7\modify_request($request, $changes);
            }
        }

        return $request;
    }
```
work for you/in this case, @franmomu ? If yes, I'll commit that.

franmomu: yeah, my concern was about the existing code that apparently works for both Guzzle 6 and 7, this seems like a proper way to deal with it 👍

tests/Transport/AwsAuthV4Test.php
```diff
@@ -198,4 +198,35 @@ public function testSignsWithEnvironmentalCredentials(): void
...
+            'aws_region' => 'us-east-1',
+            'host' => $hostWithTrailingDot,
+        ];
+        $client = $this->_getClient($config);
+
+        try {
+            $client->request('_stats');
+        } catch (GuzzleException $e) {
+            $guzzleException = $e->getGuzzleException();
+            /** @var \Psr\Http\Message\RequestInterface $request */
```
franmomu, line 226: For the PHPStan error we can ignore it like I did (adding 1 to the count)

https://github.com/ruflin/Elastica/blob/7e4de1eebdcbc5a4731fda61e098aaf40a8d0258/phpstan-baseline.neon#L138-L141

or now I see it I think I would add:

```php
use use GuzzleHttp\Exception\RequestException;

// ...
$request = $guzzleException->getRequest();
$this->assertInstanceOf(RequestException::class, $guzzleException);
$request = $guzzleException->getRequest();
```
with the `assertInstanceOf` call PHPStan understands the code and also we assert that it is the right type, I'll create a PR changing the rest of them to try.

jhuebner79:
@franmomu I force pushed the code I suggested earlier, and your suggestion as well

jhuebner79:
Found it - looks good locally now (did another force push)

jhuebner79:
@ruflin @franmomu do you need to give approval for another workflow run? Is there anything else I can do?

phpstan-baseline.neon
```diff
@@ -135,3 +135,7 @@ parameters:
 			count: 1
 			path: tests/SnapshotTest.php
 
+		-
```
thePanz, line 138: Not sure if this is the right indentation level

jhuebner79: Thanks, I'll check

jhuebner79: hmm, basically I copied lines 28-31 and did not change indentation at all 🤔 

It's the same like in the whole file, four spaces everywhere..

jhuebner79: @thePanz anything I can do here?

thePanz:
@jhuebner79 I will check the indentation later, would you be able to provide some tests covering the added lines?

jhuebner79:
> @jhuebner79 I will check the indentation later, would you be able to provide some tests covering the added lines? Looks like the code-coverage is not picking them up 🤔

We honestly have no idea why the added lines are not picked up by codecov, I'd be very happy if someone who's more familiar with the project could help us a bit here.

franmomu:
I've created https://github.com/ruflin/Elastica/pull/2113 that should help with this (it adds a job to test with lowest dependencies)

ruflin:
Lets get this moving again. I resolved the Changelog conflict which triggered another CI run and got the PR from @franmomu merged.

@jhuebner79 @thePanz What are the other things we still need?

jhuebner79:
@ruflin thanks for that!

I still don't know why codecov is not happy or what I can do about it, but I think I would need to rebase to current master?

tests/Transport/AwsAuthV4Test.php
```diff
@@ -205,4 +205,35 @@ public function testSignsWithEnvironmentalCredentials(): void
...
+            'aws_access_key_id' => 'foo',
+            'aws_secret_access_key' => 'bar',
+            'aws_session_token' => 'baz',
+            'aws_region' => 'us-east-1',
+            'host' => $hostWithTrailingDot,
+        ];
+        $client = $this->_getClient($config);
+
+        try {
+            $client->request('_stats');
```
ruflin, line 230: Is this executing all the new code? Maybe that is the issue?

ruflin (commented):
Could you try to rebase on master? But not sure if this is the issue. 

The part I was not sure about is if the test you added executes all these lines of code but based on your exception check around host, it should.

tests/Transport/AwsAuthV4Test.php
jhuebner79: We're able to have a look later on today or tomorrow.

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Updated `Query::create` PHPDoc to include supported types and propagate it to callers by @franmomu [#2088](https://github.com/ruflin/Elastica/pull/2088)
 * Update some iterable types in PHPDoc to be more specific by @franmomu [#2092](https://github.com/ruflin/Elastica/pull/2092)
 * Updated `AwsAuthV4Test` adding assertions for exception type by @franmomu [#2094](https://github.com/ruflin/Elastica/pull/2094)
+* Update `AWSAuthV4 transport` to sanitize host name for AWS requests before signing [#2090](https://github.com/ruflin/Elastica/pull/2090)
```
krasilnikovm, line 40: Pay attention, You've added a changelog in 7.2.0 release, please pull changes from master branch and move the changelog entry to [Unreleased](https://github.com/ruflin/Elastica/compare/7.2.0...master)

jhuebner79: Thanks, did that

tests/Transport/AwsAuthV4Test.php
jhuebner79: So, we hacked the make targets to be able to generate phpunit code coverage reports, and it doesn't look too bad there.. see attached screenshot. 🤔

Screenshot 2022-08-25 at 12-23-31 Code Coverage for var www html src Transport AwsAuthV4 php

tests/Transport/AwsAuthV4Test.php
```diff
@@ -205,4 +205,35 @@ public function testSignsWithEnvironmentalCredentials(): void
...
+        } catch (GuzzleException $e) {
+            $guzzleException = $e->getGuzzleException();
```
thePanz, line 232: Can we move this out of the `catch`?

We should make sure that the exception is thrown and according to our expectations.
This test case is not covering the possibility that no exceptions are thrown from `$client->request('_stats');`

WDYT?

jhuebner79: To be honest we just replicated what we already found in the file (see all other tests in the file above). Somehow those changes must've made their way to master as well. But don't get me wrong, you have a point there - it's just maybe not in the scope of our change to rework this and make it better within our PR?

ruflin:
I suggest we just leave the bots to themself and agree that codecov is fine even thought it shows it isn't? Any objections?

jhuebner79:
By reacting with the heart smiley I meant "not from my side", of course :)

ruflin:
I resolved a changelog conflict to get this mergeable. @thePanz As you were also involved in this PR, ok for you if we get this in?

ruflin:
I merged this PR now. If we see issues, we can always revert. Anyone using / testing this, please report back.

PR #2091, opened 2022-06-24
Fix `MoreLikeThis::setLike()` allowing `Document`

franmomu:
`MoreLikeThis::setLike()` allows `Document` as showed in https://github.com/ruflin/Elastica/blob/cd552e1ab6ec7ded4ee59bf693d8e1aba8980be6/tests/Query/MoreLikeThisTest.php#L96-L102

### DIFF
--- a/tests/Query/QueryStringTest.php
+++ b/tests/Query/QueryStringTest.php
@@ -17,7 +17,7 @@ class QueryStringTest extends BaseTest
      */
     public function testSearchMultipleFields(): void
     {
-        $str = \md5(\mt_rand());
+        $str = \uniqid();
         $query = new QueryString($str);
 
         $expected = [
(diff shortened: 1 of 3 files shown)


tests/Query/QueryStringTest.php
```diff
@@ -17,7 +17,7 @@ class QueryStringTest extends BaseTest
      */
     public function testSearchMultipleFields(): void
     {
-        $str = \md5(\mt_rand());
+        $str = uniqid();
```
franmomu, line 20: This change is because `\md5()` expects a `string` and `\mt_rand()` returns an `int`. The return value of `\mt_rand()` could also be casted to `string`, but instead I've used `uniqid()`, if cast is preferred, let me know.

ruflin (approved):

PR #2092, opened 2022-06-25
Specify some iterable types

franmomu:
There are quite `array` types in the code, the idea is to be more specific about what types are holding, but I guess most of them are `array<string, mixed>`, we could maybe be more specific in some cases like `AbstractSuggestion:: setRegexOptions()`.

### DIFF
--- a/src/Suggest/AbstractSuggest.php
+++ b/src/Suggest/AbstractSuggest.php
@@ -56,6 +56,11 @@ abstract class AbstractSuggest extends Param implements NameableInterface
      * Expects one of the next params: max_determinized_states - defaults to 10000,
      * flags are ALL (default), ANYSTRING, COMPLEMENT, EMPTY, INTERSECTION, INTERVAL, or NONE.
      *
+     * @phpstan-param array{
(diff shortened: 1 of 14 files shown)


src/Suggest/AbstractSuggest.php
```diff
@@ -56,6 +56,11 @@ public function setRegex(string $regex): self
      * Expects one of the next params: max_determinized_states - defaults to 10000,
      * flags are ALL (default), ANYSTRING, COMPLEMENT, EMPTY, INTERSECTION, INTERVAL, or NONE.
      *
+     * @phpstan-param array{
+     *     max_determinized_states?: int,
+     *     flags: 'ALL'|'ANYSTRING'|'COMPLEMENT'|'EMPTY'|'INTERSECTION'|'INTERVAL'|'NONE'
```
ruflin, line 61: Nice!

ruflin (approved):
I really appreciate that you put all this into small digestible PR's. Makes reviewing and getting these in much easier!

Issue #2079, opened 2022-06-19
Strict types in next major version?


franmomu:
Analyzing the code with PHPStan level to 5, there are about 656 errors and most of them (~95%) are like this:

```
 ------ -----------------------------------------------------------------------------------------
  38     Parameter #1 $id of class Elastica\Document constructor expects string|null, int given.
  73     Parameter #1 $id of class Elastica\Document constructor expects string|null, int given.
  157    Parameter #1 $id of class Elastica\Document constructor expects string|null, int given.
 ------ -----------------------------------------------------------------------------------------
```

because in `tests` when creating `Document`, it is used an `int` like:

```php
$id = 1;
$data = ['id' => $id, 'name' => 'Item 1'];
$doc = new Document($id, $data);
```
and `Document` constructor has `string` has parameter type:

https://github.com/ruflin/Elastica/blob/682b78d775a2147eda66e8d12e56aa4db3b85809/src/Document.php#L36-L43

So should we ignore those errors? or are we going to add `declare(strict_types = 1);` in next major version?

IMHO adding strict types is the way to go (and changing all `int` ids to `string` in `tests`).

ruflin:
I think the following should stay valid:

```
$doc = new Document(1, $data);
```

I know it is not strict but it is one of these nice convenience things about script languages. What we send in the end to Elasticsearch is a string. But my biggest concern is breaking all the users that might use it today and force them to change the code. Is this really worth it? What if we modify to allow `int`?

franmomu:
> I think the following should stay valid:
> 
> ```
> $doc = new Document(1, $data);
> ```

yep, I agree with that.

> I know it is not strict but it is one of these nice convenience things about script languages. What we send in the end to Elasticsearch is a string. But my biggest concern is breaking all the users that might use it today and force them to change the code. Is this really worth it? What if we modify to allow `int`?

Yeah, adding `int` is ok to me, I'll create a PR after reaching level 4 👍

franmomu:
yeah, there are some other inconsistencies, we'll see them

thePanz:
Should we close this issue as PR #2093 has been merged?

PR #2093, opened 2022-06-25
Use `string` type for Document id in tests

franmomu:
Ref: https://github.com/ruflin/Elastica/issues/2079#issuecomment-1160163215

~This PR allows `int` as possible type for `id` when creating `Document` and also in `AbstractScript`, `Script` and `ParentId`.~

~The idea of this PR is to be BC, but then we should change `AbstractUpdateAction::setId()`, `Index::createDocument()` and maybe some other that would mean BC break, for them we could maybe leave it as it is, ignore the error in PHPStan and add some kind of reminder to change it in `8.x`.~

~BTW with this change errors in PHPStan level 5 go from 624 to 81.~

This is just a proposal, we could also ignore those errors in PHPStan. I'm open to any alternative.

**Update**: This PR updates our test suite to use `string` type for `Document` ids.

### DIFF
--- a/tests/Aggregation/AdjacencyMatrixTest.php
+++ b/tests/Aggregation/AdjacencyMatrixTest.php
@@ -128,9 +128,9 @@ class AdjacencyMatrixTest extends BaseAggregationTest
     {
         $index = $this->_createIndex();
         $index->addDocuments([
-            new Document(1, ['accounts' => ['hillary', 'sidney']]),
-            new Document(2, ['accounts' => ['hillary', 'donald']]),
-            new Document(3, ['accounts' => ['vladimir', 'donald']]),
+            new Document('1', ['accounts' => ['hillary', 'sidney']]),
+            new Document('2', ['accounts' => ['hillary', 'donald']]),
+            new Document('3', ['accounts' => ['vladimir', 'donald']]),
         ]);
 
         $index->refresh();
--- a/tests/Aggregation/AvgTest.php
+++ b/tests/Aggregation/AvgTest.php
@@ -54,11 +54,11 @@ class AvgTest extends BaseAggregationTest
         $index = $this->_createIndex();
 
         $index->addDocuments([
-            new Document(1, ['price' => 5]),
-            new Document(2, ['price' => 8]),
-            new Document(3, ['price' => 1]),
-            new Document(4, ['price' => 3]),
-            new Document(5, ['anything' => 'anything']),
+            new Document('1', ['price' => 5]),
+            new Document('2', ['price' => 8]),
+            new Document('3', ['price' => 1]),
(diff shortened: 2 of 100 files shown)


src/AbstractUpdateAction.php
```diff
@@ -26,6 +28,8 @@ public function setId(?string $id = null): self
     }
 
     /**
+     * TODO: Allow also "int" as return type.
```
ruflin, line 31: I think here it should always be an `string`. Even if you set is a `int` it is converted to a string. That you can set it as an int is more a convenience for the dev but not for internal representation.

ruflin (commented):
When I made the comment in https://github.com/ruflin/Elastica/issues/2079#issuecomment-1160163215 it all sounded so simple but now looking at the PR things are a bit more complicated. For example, should we really "loosen" strict mappings that we have? It seems it worked so far and users were happy with it. 

Would the following be possible?

* Keep strict mappings where we already had it
* Update support for `int` in the docs where it was missing
* Update tests (if needed)?

Not sure I get to this PR again before I'm out, @thePanz would be great if you could have a look.

franmomu:
I changed it as draft until we see it is fine.

I've been checking the dependent packages in https://packagist.org/packages/ruflin/elastica/dependents?order_by=downloads (the most used ones) and what I've seen so far is that the ones using `Document`, they use a proper `string` as id.

> It seems it worked so far and users were happy with it.

I agree with this, since it has been typed since version 7 (and in the PHPDoc before), I think it is safeish to say that users have updated their code.

> Would the following be possible?
> 
> * Keep strict mappings where we already had it

Done.

> * Update support for `int` in the docs where it was missing

do you mean add `int` so it looks like:

```php
/**
 * Creates a new document.
 *
 * @param int|string|null  $id    The document ID, if null it will be created
 * @param array|string $data  Data array
 * @param Index|string $index Index name
 */
public function __construct(?string $id = null, $data = [], $index = '')
```
and when support for php < 8 is dropped, change it to `string|int $id`? and internally cast it always to `string`
> * Update tests (if needed)?
> 

I've updated most of the test (using regular expressions) to see how it would look like if we just keep `string`.

thePanz:
I would opt to keep our library accept and return the type used in ES.
It would feel weird to index a product with `int 123` and later load it, to have `string "123"` in the response.

I am more keen to keep the strict `string` type.

For this PR it means:
- code wise it is good, all calls have been casted to `string`
- remove the new lines from the cangelog :)

WDYT?

franmomu:
Yeah, the initial issue here was about [tests not being in sync with the types defined in `src` because of not using strict types](https://github.com/ruflin/Elastica/issues/2079#issue-1276007828).

This change should be fine and we'll see about adding strict types in the next major version.

PR #2096, opened 2022-06-30
Fix some PHPDoc

franmomu:
`Term::setTerm()` has `array` as possible type because `Boost can be set with array`, but I think that is legacy PHPDoc when there was no `$boost` parameter and `$value` was used for everything:

https://github.com/ruflin/Elastica/blob/7a35358c52e9b863253652c49e2888803cee5986/lib/Elastica/Filter/Term.php#L37-L48

and now is:

https://github.com/ruflin/Elastica/blob/f72262d1cb68668d7a370fb113e04b4809f92e7a/src/Query/Term.php#L35-L47

Apart from `string` other values can be used like:

https://github.com/ruflin/Elastica/blob/50346d813e7c0e764bafc88a0a79cffe8d039b35/tests/Query/MoreLikeThisTest.php#L114-L115

~For `Search::addIndex()` there is a `is_scalar` check below:~

~https://github.com/ruflin/Elastica/blob/50346d813e7c0e764bafc88a0a79cffe8d039b35/src/Search.php#L80-L95~

### DIFF
--- a/src/AbstractUpdateAction.php
+++ b/src/AbstractUpdateAction.php
@@ -19,6 +19,8 @@ class AbstractUpdateAction extends Param
 
     /**
      * Sets the id of the document.
+     *
+     * @return $this
      */
     public function setId(?string $id = null): self
     {
--- a/tests/BulkTest.php
+++ b/tests/BulkTest.php
@@ -711,7 +711,7 @@ JSON;
     {
         $bulk = new Bulk($this->_getClient());
 
-        $this->assertSame($bulk, $bulk->setShardTimeout(10));
+        $this->assertSame($bulk, $bulk->setShardTimeout('10s'));
     }
 
     /**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fixed `Http` PHPDoc adding `\CurlHandle` type for Curl connection by @franmomu [#2086](https://github.com/ruflin/Elastica/pull/2086)
 * Fixed case mismatch in method calls by @franmomu [#2087](https://github.com/ruflin/Elastica/pull/2087)
 * Fixed `MoreLikeThis::setLike()` PHPDoc allowing `Document` by @franmomu [#2091](https://github.com/ruflin/Elastica/pull/2091)
+* Fixed `Term::setTerm()` PHPDoc allowing scalar values for `$value` parameter by @franmomu [#2094](https://github.com/ruflin/Elastica/pull/2094)
 
 ### Security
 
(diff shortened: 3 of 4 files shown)


tests/BulkTest.php
```diff
@@ -711,7 +711,7 @@ public function testSetShardTimeout(): void
     {
         $bulk = new Bulk($this->_getClient());
 
-        $this->assertSame($bulk, $bulk->setShardTimeout(10));
+        $this->assertSame($bulk, $bulk->setShardTimeout('10s'));
```
franmomu, line 714: `Bulk:: setShardTimeout()` expects an string and seems right:

https://www.elastic.co/guide/en/elasticsearch/reference/7.17/docs-bulk.html#docs-bulk-api-query-params

franmomu:
friendly ping @thePanz and @deguif if you have some time to review this, thanks!

src/AbstractUpdateAction.php
```diff
@@ -19,6 +19,8 @@ class AbstractUpdateAction extends Param
 
     /**
      * Sets the id of the document.
+     *
+     * @return static
```
deguif, line 23:
```suggestion
     * @return $this
```

That's the way this kind of return is type-hinted in the codebase.

franmomu: 👍 changed

thePanz (changes requested):

tests/BulkTest.php
thePanz: good catch!

src/Search.php
```diff
@@ -80,7 +80,7 @@ public function __construct(Client $client, ?BuilderInterface $builder = null)
     /**
      * Adds a index to the list.
      *
-     * @param Index|string $index Index object or string
+     * @param bool|float|Index|int|string $index Index object or string
```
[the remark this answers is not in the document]
thePanz: Thanks! what about changing the handling into something like:
```
if ($index !instanceOf Index || !is_string($index)) {
  // Throw the exception
}
```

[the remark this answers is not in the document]
thePanz: but yes, this should be another PR

src/AbstractUpdateAction.php
```diff
@@ -19,6 +19,8 @@ class AbstractUpdateAction extends Param
 
     /**
      * Sets the id of the document.
+     *
+     * @return $this
```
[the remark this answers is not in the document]
deguif: On child class, self is resolved as the abstract class not the child, which can be annoying when chaining methods.
Starting from PHP 8.0, we will be able to use `static` which will solve this issue ;)

thePanz:
Thanks @franmomu , merged!

Issue #2097, opened 2022-07-01
[DateHistogram] `interval` field is deprecated


kissifrot:
Hello,

`interval` field is deprecated in [7.x](https://www.elastic.co/guide/en/elasticsearch/reference/7.14/search-aggregations-bucket-datehistogram-aggregation.html#calendar_and_fixed_intervals) (in favor of `fixed_interval)` and was deleted in 8.x version of ElasticsSearch, would it be possible to replace it?

thePanz:
Fixed in d4849e2

PR #2099, opened 2022-07-11
Avoid deprecation with interval for DateHistogram

VincentLanglet:
This could be a way to solve https://github.com/ruflin/Elastica/issues/2097

### DIFF
--- a/src/Aggregation/DateHistogram.php
+++ b/src/Aggregation/DateHistogram.php
@@ -38,10 +38,10 @@ class DateHistogram extends Histogram
     public function setInterval($interval): Histogram
     {
         if (\in_array($interval, self::CALENDAR_INTERVAL)) {
-            return $this->setCalendarInterface($interval);
+            return $this->setCalendarInterval($interval);
         }
 
-        return $this->setFixedInterface($interval);
+        return $this->setFixedInterval($interval);
     }
 
     /**
@@ -49,7 +49,7 @@ class DateHistogram extends Histogram
      *
      * @return $this
      */
-    public function setCalendarInterface($interval): self
+    public function setCalendarInterval($interval): self
     {
         return $this->setParam('calendar_interval', $interval);
     }
@@ -59,7 +59,7 @@ class DateHistogram extends Histogram
      *
      * @return $this
      */
-    public function setFixedInterface($interval): self
+    public function setFixedInterval($interval): self
     {
         return $this->setParam('fixed_interval', $interval);
     }
--- a/src/Aggregation/DateHistogram.php
+++ b/src/Aggregation/DateHistogram.php
@@ -38,9 +38,29 @@ class DateHistogram extends Histogram
     public function setInterval($interval): Histogram
     {
         if (\in_array($interval, self::CALENDAR_INTERVAL)) {
-            return $this->setParam('calendar_interval', $interval);
+            return $this->setCalendarInterface($interval);
         }
 
+        return $this->setFixedInterface($interval);
+    }
+
+    /**
+     * @param int|string $interval
+     *
+     * @return $this
+     */
+    public function setCalendarInterface($interval): self
+    {
+        return $this->setParam('calendar_interval', $interval);
+    }
+
+    /**
+     * @param int|string $interval
+     *
+     * @return $this
+     */
+    public function setFixedInterface($interval): self
+    {
         return $this->setParam('fixed_interval', $interval);
     }
 
--- a/src/Aggregation/DateHistogram.php
+++ b/src/Aggregation/DateHistogram.php
@@ -37,7 +37,7 @@ class DateHistogram extends Histogram
      */
     public function setInterval($interval): Histogram
     {
-        if (\in_array($interval, self::CALENDAR_INTERVAL)) {
+        if (\in_array($interval, self::CALENDAR_INTERVAL, true)) {
             return $this->setCalendarInterval($interval);
         }
 
--- a/src/Aggregation/DateHistogram.php
+++ b/src/Aggregation/DateHistogram.php
@@ -35,7 +35,7 @@ class DateHistogram extends Histogram
      *
      * @return $this
      */
-    public function setInterval($interval): Histogram
+    public function setInterval($interval): self
     {
         if (\in_array($interval, self::CALENDAR_INTERVAL_VALUES, true)) {
             return $this->setCalendarInterval($interval);
@@ -45,21 +45,17 @@ class DateHistogram extends Histogram
     }
 
     /**
-     * @param int|string $interval
-     *
      * @return $this
      */
-    public function setCalendarInterval($interval): self
+    public function setCalendarInterval(string $interval): self
     {
         return $this->setParam('calendar_interval', $interval);
     }
 
     /**
-     * @param int|string $interval
-     *
      * @return $this
      */
-    public function setFixedInterval($interval): self
+    public function setFixedInterval(string $interval): self
     {
         return $this->setParam('fixed_interval', $interval);
     }
--- a/src/Aggregation/DateHistogram.php
+++ b/src/Aggregation/DateHistogram.php
@@ -37,6 +37,8 @@ class DateHistogram extends Histogram
      */
     public function setInterval($interval): self
     {
+        $interval = (string) $interval;
+
         if (\in_array($interval, self::CALENDAR_INTERVAL_VALUES, true)) {
             return $this->setCalendarInterval($interval);
         }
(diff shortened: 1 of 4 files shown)


franmomu:
what about to create `setCalendarInterval`, `setFixedInterval` and deprecate `setInterval`? to be in sync with elasticsearch

VincentLanglet:
> what about creating `setCalendarInterval`, `setFixedInterval` and deprecate `setInterval`? to be in sync with elasticsearch

DateHistogram extends Histogram, and the Histogram constructor expects a `interval` value as third parameter (and then call setInterval in the constructor).
I thought about adding a 4th parameter to set if it's calendar or fixed, but this can automatically be determined like I did, so it's kinda useless.
Also, the setInterval method is coming from the Histogram class. So if you want to remove it, you need to stop extending Histogram.

An other strategy started in https://github.com/ruflin/Elastica/pull/1874 but never ended...

thePanz:
I like this approach, later on we could deprecate the `setInterval`,

We should also add the `setCalendarInterval` and `setFixedInterval` methods as suggested by @franmomu, to allow users to properly set the intervals following the ES documentation.

WDYT?

Later we could even have 2 different classes implementing the `CalendarIntervaDateHistograml` and `FixedIntervalDateHistogram`, so that we could enforce the creation of valid objects.
This would be a topic for another PR, imho.

VincentLanglet:
> I like this approach, later we could deprecate the `setInterval`.
> 
> We should also add the `setCalendarInterval` and `setFixedInterval` methods as suggested by @franmomu, to allow users to properly set the intervals following the ES documentation.
> 
> WDYT?

Done @thePanz

src/Aggregation/DateHistogram.php
```diff
@@ -11,6 +11,59 @@ class DateHistogram extends Histogram
...
+
+        return $this->setFixedInterface($interval);
+    }
+
+    /**
+     * @param int|string $interval
+     *
+     * @return $this
+     */
+    public function setCalendarInterface($interval): self
```
thePanz, line 52: do you mean 'setCalendarInterval()`?

src/Aggregation/DateHistogram.php
```diff
@@ -11,6 +11,59 @@ class DateHistogram extends Histogram
...
+    {
+        return $this->setParam('calendar_interval', $interval);
+    }
+
+    /**
+     * @param int|string $interval
+     *
+     * @return $this
+     */
+    public function setFixedInterface($interval): self
```
thePanz, line 62: do you mean `setFixedInterval()`?

tests/Aggregation/DateHistogramTest.php
```diff
@@ -99,7 +99,7 @@ public function testSetOffset(): void
         $expected = [
             'date_histogram' => [
                 'field' => 'created',
-                'interval' => '1h',
```
thePanz, line 102: Do we have test covering the fixed_interval alternative?

thePanz (changes requested):

src/Aggregation/DateHistogram.php
VincentLanglet: Indeed, fixed

src/Aggregation/DateHistogram.php
VincentLanglet: Indeed, fixed

tests/Aggregation/DateHistogramTest.php
VincentLanglet: I added two different test of the construct method

tests/Aggregation/DateHistogramTest.php
```diff
@@ -13,6 +13,40 @@
  */
 class DateHistogramTest extends BaseAggregationTest
 {
+    /**
+     * @group unit
+     */
+    public function testConstructForCalendarInterval(): void
+    {
+        $agg = (new DateHistogram('hist', 'created', '1h'));
```
thePanz, line 21:
```suggestion
        $agg = new DateHistogram('hist', 'created', '1h');
```

tests/Aggregation/DateHistogramTest.php
```diff
@@ -13,6 +13,40 @@
...
+
+        $this->assertSame($expected, $agg->toArray());
+    }
+
+    /**
+     * @group unit
+     */
+    public function testConstructForFixedInterval(): void
+    {
+        $agg = (new DateHistogram('hist', 'created', '2h'));
```
thePanz, line 38:
```suggestion
        $agg = new DateHistogram('hist', 'created', '2h');
```

thePanz (changes requested):

src/Aggregation/DateHistogram.php
```diff
@@ -11,6 +11,59 @@ class DateHistogram extends Histogram
...
+    ];
+
+    /**
+     * Set the interval by which documents will be bucketed.
+     *
+     * @param int|string $interval
+     *
+     * @return $this
+     */
+    public function setInterval($interval): Histogram
```
thePanz, line 38: This returns an instance of the current clas, not the parent one.
WDYT?

```suggestion
    public function setInterval($interval): DateHistogram
```

VincentLanglet: I have to respect the parent signature, which is `self` and not `static`.

thePanz: I see, and we can not use that until php v8
What about altering the parent method (the docblock should be added to this method too, btw), to use
```/**
    * @return static
    */```

and removing the `: self` from the method signature?

File: src/Aggregation/DateHistogram.php [hunk 7]
```diff
@@ -11,6 +11,59 @@ class DateHistogram extends Histogram
...
+
+        return $this->setFixedInterval($interval);
+    }
+
+    /**
+     * @param int|string $interval
+     *
+     * @return $this
+     */
+    public function setCalendarInterval($interval): self
```
franmomu, line 52:
```suggestion
    /**
     * @return $this
     */
    public function setCalendarInterval(string $interval): self
```
What about creating these methods with the right type? `setInterval` allows `int|string`, but in [the ES documentation](https://www.elastic.co/guide/en/elasticsearch/reference/7.14/search-aggregations-bucket-datehistogram-aggregation.html#calendar_and_fixed_intervals) seems like it only accepts `string`.

It should be fine since we don't have `strict_types` so the value from `setInterval` will be casted to `string`.

src/Aggregation/DateHistogram.php
VincentLanglet: I removed `:self` to add the typehint in DateHistogram.

`@return $this` seems better than `@return static` I think.

src/Aggregation/DateHistogram.php
thePanz: The problem now is that `setInterval()` accepts an `int|string`, while the two methods it calls, they expect a `string`.

I guess we can not force `setInterval(string $internal)` as signature, as it will not be compatible with the parent class

src/Aggregation/DateHistogram.php
```diff
@@ -11,6 +11,55 @@ class DateHistogram extends Histogram
...
+     *
+     * @return $this
+     */
+    public function setInterval($interval): self
+    {
+        if (\in_array($interval, self::CALENDAR_INTERVAL_VALUES, true)) {
+            return $this->setCalendarInterval($interval);
+        }
+
+        return $this->setFixedInterval($interval);
```
thePanz, line 44: We must cast the interval to string, as the two functions accepts a string only, not an integer

```suggestion
            return $this->setCalendarInterval((string) $interval);
        }

        return $this->setFixedInterval((string) $interval);
```

thePanz (changes requested):

src/Aggregation/DateHistogram.php
VincentLanglet: This is not needed because
- We don't use `strict_types` so it's casted automatically
- No real `int` value are supported. It's always a number and a unit, like `2y`. 2 is not a supported value.

thePanz: but in the spirit to prepare to move to strict_types and PHP8, I see how
- adding casting to String
- and having the proper phpdoc
 
could help 3rd party users to properly migrate or prepare to migrate.
wdyt?

VincentLanglet: I've trouble to understand.

If someone is calling
```
->setInterval(3);
```
the code already doesn't work.

Casting the value to string won't solve anything, since `'3'` is not a valid value either for DateHistogram interval.

IMHO, when creating CalendarIntervaDateHistograml or FixedIntervalDateHistogram, int support should be removed since it's not a valid value.

Anyway, I added a cast if it's a blocker for you.

thePanz (approved):

thePanz:
LGTM, thanks @VincentLanglet!
Would you add the changes to the Changelog?

thePanz:
Merged in d4849e2
Thanks @VincentLanglet !

PR #2102, opened 2022-07-20
Improve types in Query

b1rdex:


src/Index.php
```diff
@@ -164,9 +166,11 @@ public function updateDocuments(array $docs, array $options = []): ResponseSet
     /**
      * Update entries in the db based on a query.
      *
-     * @param AbstractQuery|array|Collapse|Query|string|Suggest $query   Query object or array
-     * @param AbstractScript                                    $script  Script
-     * @param array                                             $options Optional params
+     * @param AbstractQuery|AbstractSuggest|array|Collapse|Query|string|Suggest|null $query Query object or array
```
franmomu, line 169: This PR looks nice! I was updating these lines recently to add `AbstractSuggest`, but in this case I remember I didn't add it because I wasn't sure if that makes sense.

`Query::create()` accepts all these types, but in this particular case, `updateByQuery` (and `deleteByQuery`) I guess `AbstractSuggest`, `Collapse` and `Suggest` doesn't make sense here and probably `null` as well.

b1rdex: Yeah, that makes sense. I removed abstract/suggest and collapse. I left null in place because it's used to create a match all query.

src/Query/HasChild.php
```diff
@@ -28,7 +33,8 @@ public function __construct($query, ?string $type = null)
     /**
      * Sets query object.
      *
-     * @param AbstractQuery|BaseQuery|string $query
+     * @param AbstractQuery|AbstractSuggest|array|Collapse|Query|string|Suggest|null $query
+     * @phpstan-param TCreateQueryArgs $query
```
franmomu, line 37: I think here `TCreateQueryArgsForModifyingQueries` is enough

src/Query/HasParent.php
```diff
@@ -26,7 +31,8 @@ public function __construct($query, string $type)
     /**
      * Sets query object.
      *
-     * @param AbstractQuery|BaseQuery|string $query
+     * @param AbstractQuery|AbstractSuggest|array|Collapse|Query|string|Suggest|null $query
+     * @phpstan-param TCreateQueryArgs $query
```
franmomu, line 35: here `TCreateQueryArgsForModifyingQueries` too

src/Rescore/Query.php
```diff
@@ -50,7 +55,8 @@ public function toArray(): array
     /**
      * Sets rescoreQuery object.
      *
-     * @param BaseQuery|\Elastica\Query\AbstractQuery|string $rescoreQuery
+     * @param AbstractQuery|AbstractSuggest|array|BaseQuery|Collapse|string|Suggest|null $rescoreQuery
+     * @phpstan-param TCreateQueryArgs $rescoreQuery
```
franmomu, line 59: and here `TCreateQueryArgsForModifyingQueries` too.

src/Query.php
```diff
@@ -19,18 +20,62 @@
  * @author Nicolas Ruflin <spam@ruflin.com>
  *
  * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html
+ * @phpstan-type TCreateQueryArgs = TCreateQueryArgsForModifyingQueries|AbstractSuggest|Collapse|Suggest
+ * @phpstan-type TCreateQueryArgsForModifyingQueries = AbstractQuery|TRawQuery|self|string|null
+ * @phpstan-type TRawQuery = array{
```
ruflin, line 25: Haven't really used phpstan on my end yet. How will we make sure this stays up-to-date? Is this automated / validated by phpstan? If it is automated, do we actually need it in the source code or could the IDE do this directly? 

My more fundamental question is: What problem are we solving with this addition / how will it make the developer experience better.

b1rdex: > How will we make sure this stays up-to-date? Is this automated / validated by phpstan?

It's not automated and must be edited by hand.

Yes, it's validated:
1. you can't write a piece of code that violate annotations
2. you can't write an annotation that violate existing code

And the validation is already in the CI: https://github.com/ruflin/Elastica/runs/7645330172?check_suite_focus=true#step:7:1

> What problem are we solving with this addition / how will it make the developer experience better.

The problem is that `array` doesn't have/give any information about its contents. Adding enhanced type annotations help to understand what a piece of code does, what it accepts and what it produces.

src/Rescore/Query.php
b1rdex: Fixed all of the above + changed `\Elastica\SearchableInterface::count()` too. Also, I renamed `TCreateQueryArgsForModifyingQueries` to `TCreateQueryArgsMatching`

ruflin:
Overall change LGTM. Would be good to add a changelog.

@franmomu Could you have a look again as you have more experience with phpstan. Please approve if all looks good so we can get thi sin.

franmomu (approved):
Sorry I've been away some time, this LGTM! I've tried to come up with another name for `TCreateQueryArgsMatching`, but with no luck 😞

src/Rescore/Query.php
```diff
@@ -10,13 +11,12 @@
...
  * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-rescore.html
+ * @phpstan-import-type TCreateQueryArgsMatching from BaseQuery
  */
 class Query extends AbstractRescore
 {
     /**
-     * Constructor.
-     *
-     * @param \Elastica\Query\AbstractQuery|string $query
+     * @param AbstractQuery|array|BaseQuery|string|null $query
```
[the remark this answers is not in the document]
b1rdex: Fixed

src/Search.php
```diff
@@ -288,7 +291,7 @@ public function hasIndexByName(string $index): bool
     public function getQuery(): Query
     {
         if (null === $this->_query) {
-            $this->_query = Query::create('');
+            $this->_query = new Query(new MatchAll());
```
ruflin, line 294: I assume this should not change anything for users?

ruflin (approved):
I'm getting this in. My general assumption is that if we adjust things in the comments or get something wrong on the php-stan comments I don't consider it a breaking change.

ruflin:
@b1rdex Thank you very much for your contribution.

src/Search.php
b1rdex: Yes, that's the same thing - I just made it explicit.

PR #2103, opened 2022-07-21
Deprecate passing `string` to Search::addIndex, `hasIndex` and `addIndices`

franmomu:
Ref: https://github.com/ruflin/Elastica/pull/2100#issuecomment-1189189514

Right now `Search::addIndex()` and `Search::hasIndex()` methods allow `string|Index` as parameter and `Search::addIndices()` allows an array of `Index` or `string`.

This PR creates `Search::addIndexByName()` and `Search::hasIndexByName()` which only allows `string` as parameter and  `Search::addIndicesByName()` which only accepts an array of `string`.

This PR also deprecates passing `string` to `Search::addIndex()`, `Search::hasIndex()` and `Search::addIndices()`.

### DIFF
--- a/tests/Query/MatchQueryTest.php
+++ b/tests/Query/MatchQueryTest.php
@@ -127,39 +127,6 @@ class MatchQueryTest extends BaseTest
     /**
      * @group functional
      */
-    public function testMatchSetFieldBoostWithString(): void
-    {
-        $client = $this->_getClient();
-        $index = $client->getIndex('test');
-        $index->create([], [
-            'recreate' => true,
-        ]);
-
-        $index->addDocuments([
-            new Document('1', ['name' => 'Basel-Stadt']),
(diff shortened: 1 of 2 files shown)


src/Search.php
```diff
@@ -80,34 +80,74 @@ public function __construct(Client $client, ?BuilderInterface $builder = null)
...
+        return $this;
+    }
+
+    /**
+     * @param string[] $indices
+     */
+    public function addIndicesByName(array $indices = []): self
+    {
+        foreach ($indices as $index) {
+            $this->addIndexByName($index);
```
thePanz, line 150: we should check if `$index` is *really* a string

thePanz (changes requested):
When adding indexesByName we should make sure that each item is a non-empty string

thePanz:
Moved to #2106

PR #2104, opened 2022-07-22
Add inferred return types

thePanz:
Applied those by using Rector
See: https://getrector.org/blog/how-to-automatically-add-return-type-declarations-without-breaking-your-code

### DIFF
--- a/src/AbstractUpdateAction.php
+++ b/src/AbstractUpdateAction.php
@@ -461,20 +461,15 @@ class AbstractUpdateAction extends Param
         return $this->_upsert;
     }
 
-    /**
-     * @return bool
-     */
-    public function hasUpsert()
+    public function hasUpsert(): bool
     {
         return null !== $this->_upsert;
     }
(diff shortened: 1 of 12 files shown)


deguif:
That should target the next major release as that would break backward compatibility. WDYT?

thePanz:
> That should target the next major release as that would break backward compatibility. WDYT?

The return types were inferred by the actual code's usage and return values.
But I see the point, it could break 3rd party codes if they extend and overwrite one of the changed methods.
Is that what you mean?

deguif:
> But I see the point: it could break 3rd party codes if they extend and overwrite one of the changed methods.

Yes absolutely

thePanz:
right, I'll close this PR.
Let's hope to include Rector when we'll move to the v8.x version in the future

thePanz:
@deguif on a second thought, wouldn't this PR introduce the same BC-break as #2085?

PR #2105, opened 2022-07-23
Prepare release 7.2.0

thePanz:
I am not 100% sure about the BC breaks, as they would call for a major release

Any inputs @deguif ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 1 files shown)


franmomu:
For the BC breaks introduced see https://github.com/ruflin/Elastica/pull/2082#discussion_r903405419 and https://github.com/ruflin/Elastica/pull/2082#issuecomment-1163082427, they are in classes not likely to be extended and it was to fix some behaviour.

deguif (approved):
There is a BC break as pointed in the comment, but its impact should be limited.

deguif:
@thePanz there are some deprecations to update. For example in this PR (https://github.com/ruflin/Elastica/pull/2070/files) it was marked as introduced in 7.1.6 but should be 7.2.0.

thePanz:
@deguif I rebased this PR to the latest changes

Thanks @franmomu for the PR fixing the deprecation messages with the correct version

thePanz:
Merged, and v7.2.0 released

PR #2107, opened 2022-07-25
Fix deprecation version

franmomu:
Adding a deprecation means to release a minor version, in https://github.com/ruflin/Elastica/pull/2070 I used a patch version. This PR fixes it as pointed out in https://github.com/ruflin/Elastica/pull/2105#issuecomment-1193629646.


thePanz:
Thanks @franmomu

PR #2108, opened 2022-08-04
Bump PHPStan to level 5

franmomu:
Most of the ignored errors are expected because they are testing exceptions using wrong types and PHPStan complains about that. Like:

https://github.com/ruflin/Elastica/blob/8fe61767b604a10e40b0b59c9511c4317cae3cb8/tests/Aggregation/GeohashGridTest.php#L51-L57

```
 ------ ---------------------------------------------------------------------------------------------------------------------
  Line   tests/Aggregation/GeohashGridTest.php
 ------ ---------------------------------------------------------------------------------------------------------------------
  56     Parameter #1 $precision of method Elastica\Aggregation\GeohashGrid::setPrecision() expects int|string, float given.
 ------ ---------------------------------------------------------------------------------------------------------------------
```

The different ones are:
- The ones in `tests/Query/FunctionScoreTest.php`, which I think it's fine to  add `TODO` just in case someone has overridden `FunctionScore::addDecayFunction()` method, if this is fine we can close https://github.com/ruflin/Elastica/pull/2089
- In `ClientFunctionTest`, there is a call to `Document::setIndex()` passing `null` which is not allowed, that is done to make sure that the index is properly set when deleting the documents. Instead of that a new document can be passed as we do lines above to update them:
https://github.com/ruflin/Elastica/blob/8fe61767b604a10e40b0b59c9511c4317cae3cb8/tests/ClientFunctionalTest.php#L161-L182

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -26,9 +26,24 @@ parameters:
 			path: src/QueryBuilder.php
 
 		-
+			message: "#^Function GuzzleHttp\\\\Psr7\\\\modify_request not found\\.$#"
+			count: 1
+			path: src/Transport/AwsAuthV4.php
+
+		-
 			message: "#^Function GuzzleHttp\\\\Psr7\\\\stream_for not found\\.$#"
 			count: 1
 			path: src/Transport/Guzzle.php
+
+		-
+			message: "#^Parameter \\#1 \\$precision of method Elastica\\\\Aggregation\\\\GeohashGrid\\:\\:setPrecision\\(\\) expects int\\|string, float given\\.$#"
+			count: 1
+			path: tests/Aggregation/GeohashGridTest.php
+
+		-
+			message: "#^Parameter \\#1 \\$data of static method Elastica\\\\Bulk\\\\Action\\\\AbstractDocument\\:\\:create\\(\\) expects Elastica\\\\Document\\|Elastica\\\\Script\\\\AbstractScript, stdClass given\\.$#"
+			count: 1
+			path: tests/Bulk/Action/AbstractDocumentTest.php
 
 		-
 			message: "#^Access to an undefined property Elastica\\\\Document\\:\\:\\$counter\\.$#"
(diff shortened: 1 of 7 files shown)


tests/ClientFunctionalTest.php
```diff
@@ -174,8 +174,10 @@ public function testBulkIndex(): void
         $this->assertEquals('AnonCoin', $index->getDocument(1)->get('name'));
         $this->assertEquals('iXcoin', $index->getDocument(2)->get('name'));
 
-        $ixCoin->setIndex(null);  // Make sure the index gets set properly if missing
-        $index->deleteDocuments([$anonCoin, $ixCoin]);
+        $index->deleteDocuments([
+            new Document('1', ['name' => 'AnonCoin']),
```
ruflin, line 178: Why do you need to recreate the doc here for this change? I also wander why we do it onl line 170/171?

phpstan-baseline.neon
```diff
@@ -30,6 +30,16 @@ parameters:
 			count: 1
 			path: src/Transport/Guzzle.php
 
+		-
```
ruflin, line 33: Trying to get my head around why we don't fix these in the test files directly? Related to breaking changes?

franmomu: This is what it was explained in the PR description https://github.com/ruflin/Elastica/pull/2108#issue-1328708352, a test like this: 

https://github.com/ruflin/Elastica/blob/8fe61767b604a10e40b0b59c9511c4317cae3cb8/tests/Aggregation/GeohashGridTest.php#L51-L57

makes PHPStan fails because `setPrecision` only accepts `int|string`:

https://github.com/ruflin/Elastica/blob/6d51cdbff2fb37aceac8b70adf9042ccb9be0aed/src/Aggregation/GeohashGrid.php#L42-L50

We can get rid of these ignored errors and checks in the next major if we add type parameter declarations.

tests/ClientFunctionalTest.php
franmomu: I've updated the test to reuse the documents and added a couple of tests to check that the index is set if missing (to remove the `$ixCoin->setIndex(null);` part)

thePanz (approved):

ruflin:
@franmomu Thanks for getting us to phpstan level 5

PR #2109, opened 2022-08-05
Improve PHPDoc types in `Cluster`

franmomu:
In order to reach PHPStan level 6, we need to specify iterable types, in this case this is done in the `Cluster` namespace.

### DIFF
--- a/src/Cluster/Health.php
+++ b/src/Cluster/Health.php
@@ -11,6 +11,26 @@ use Elastica\Cluster\Health\Index;
  * @author Ray Ward <ray.ward@bigcommerce.com>
  *
  * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-health.html
+ *
+ * @phpstan-type HealthData = array{
+ *   cluster_name: string,
+ *   status: HealthStatus,
+ *   timed_out: bool,
+ *   number_of_nodes: int,
+ *   number_of_data_nodes: int,
+ *   active_primary_shards: int,
+ *   active_shards: int,
+ *   relocating_shards: int,
+ *   initializing_shards: int,
+ *   unassigned_shards: int,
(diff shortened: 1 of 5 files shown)


src/Cluster/Health.php
```diff
@@ -11,6 +11,26 @@
  * @author Ray Ward <ray.ward@bigcommerce.com>
  *
  * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-health.html
+ *
+ * @phpstan-type HealthData = array{
```
ruflin, line 15: See my comment in the other PR related to these. The question is likely mostly for my understanding.

src/Cluster/Health.php
```diff
@@ -60,6 +84,7 @@ public function getClusterName(): string
      * Gets the status of the cluster.
      *
      * @return string green, yellow or red
+     * @phpstan-return HealthStatus
```
thePanz, line 87: We could use `@return 'green'|'yellow'|'red'` instead of introducing virtual class ;-)

see: https://phpstan.org/writing-php-code/phpdoc-types#literals-and-constants

src/Cluster/Health/Index.php
```diff
@@ -8,6 +8,20 @@
  * @author Ray Ward <ray.ward@bigcommerce.com>
  *
  * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-health.html
+ *
+ * @phpstan-import-type HealthStatus from \Elastica\Cluster\Health
+ * @phpstan-import-type ShardData from Shard
```
thePanz, line 13: should we prefer a fully-qualified-class name here?

ruflin:
Closing as no recent updates. But in general in favor of these updates. Please reopen if needed.

PR #2111, opened 2022-08-06
Removed unused files in tests

franmomu:
I was trying to add missing types and bumped into these files that apparently are not used.

`CallbackStrategyTestHelper` was added in https://github.com/ruflin/Elastica/pull/871 and looks like it has been replaced by a callback:

https://github.com/ruflin/Elastica/blob/8fe61767b604a10e40b0b59c9511c4317cae3cb8/tests/Connection/Strategy/CallbackStrategyTest.php#L25-L33 

`ErrorsCollector` was added in https://github.com/ruflin/Elastica/pull/1031 and its usage was removed in https://github.com/ruflin/Elastica/pull/1742.


ruflin (approved):
Thx for the cleanup.

PR #2112, opened 2022-08-06
Add missing types

franmomu:
Raising PHPStan level to 6 shows missing types, this PR tries to add missing types and generics (we still need to add iterable types).

### DIFF
--- a/src/Scroll.php
+++ b/src/Scroll.php
@@ -10,6 +10,8 @@ use Elastica\Exception\InvalidException;
  * @author Manuel Andreo Garcia <andreo.garcia@gmail.com>
  *
  * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-scroll.html
+ *
+ * @template-implements \Iterator<string|null, ResultSet>
  */
 class Scroll implements \Iterator
 {
@@ -38,11 +40,18 @@ class Scroll implements \Iterator
      * 1: scroll id.
      * 2: ignore_unavailable.
      *
-     * @var array
+     * @var array{mixed, mixed, mixed}
      */
     protected $_options = [null, null, null];
 
+    /**
+     * @var int
+     */
     private $totalPages = 0;
+
+    /**
+     * @var int
+     */
     private $currentPage = 0;
 
     public function __construct(Search $search, string $expiryTime = '1m')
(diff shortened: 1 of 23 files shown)


tests/Transport/TransportBenchmarkTest.php
```diff
@@ -18,8 +18,24 @@
...
+     */
     protected $_max = 1000;
+
+    /**
+     * @var int
+     */
     protected $_maxData = 20;
+
+    /**
+     * @var array<string, array<string, array{
```
ruflin, line 32: Is this auto generated or manual? Same question for the other more complex blocks?

thePanz: This is manual, as phpstan is not that "smart" :)

tests/Transport/TransportBenchmarkTest.php
```diff
@@ -18,8 +18,24 @@
...
+     *    count: int,
+     *    max: int,
+     *    min: int,
+     *    mean: float,
+     * }>>
```
thePanz, line 37:
```suggestion
     * }
```

stof: this suggestion is wrong. We need to close the generic types for the 2 levels of arrays after closing the array shape for the inner array

thePanz: Thanks for resurrecting this PR, indeed the `}>>` is correct as the `array<key, array<key, { ... }>>` is used

thePanz:
@franmomu would you have time to rebase this PR?

src/Scroll.php
```diff
@@ -10,6 +10,8 @@
  * @author Manuel Andreo Garcia <andreo.garcia@gmail.com>
  *
  * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-scroll.html
+ *
+ * @template-implements \Iterator<string|null, ResultSet>
```
stof, line 14:
```suggestion
 * @template-implements \Iterator<string, ResultSet>
```

The iterator key is never `null` when the iterator is in a state where the `valid` method returns `true` (which is the state you care about during iteration)

ruflin:
Closing as no recent changes to the PR. But as mentioned in the other PR, in general open for these kind of contributions. I also wonder how much we could use some of the LLM to automate some of these updates?

PR #2113, opened 2022-08-09
Test lowest dependencies in CI

franmomu:
Using [`ramsey/composer-install`](https://github.com/ramsey/composer-install#about) allows us to simplify the CI jobs removing the cache steps for composer. 

This PR adds a job to run our tests using the lowest set of dependencies, while doing so, some tests failed because [the `OpenPointInTime` class](https://github.com/ruflin/Elastica/blob/8fe61767b604a10e40b0b59c9511c4317cae3cb8/src/Index.php#L325-L336) did not exist, that is because it was added in [`elasticsearch/elasticsearch` `7.10`](https://github.com/elastic/elasticsearch-php/releases/tag/v7.10.0) so I bumped the library to that version.


ruflin (approved):
I quite like this addition as we keep knowing what the minimal version of ES is that has all the features also in Elastica. It doesn't mean you can't use Elastica with Elasticsearch 7.0 but there are functions that only work with never version (which I think should be expected). 

One thought I have around versioning is if we add features only supported in 7.15, should we also increment the Elastica version to this number?

PR #2117, opened 2022-08-25
Terms::setMissingBucket()

andriinavrotskii:


tests/Aggregation/TermsTest.php
```diff
@@ -162,4 +162,15 @@ private function getIndex(): Index
 
         return $index;
     }
+
+    /**
+     * @group functional
```
ruflin, line 167: This does not look like a functional test. As far as I can see, no call is made to Elasticsearch. Any chance you could expand the test in a way that a call to Elasticsearch is made to confirm the functionality.

ruflin (commented):
Thanks for the contribution. Left one comment around the tests.

tests/Aggregation/TermsTest.php
andriinavrotskii: You are right. My mistake. Updated.

ruflin (approved):

ruflin:
Thanks for the change!

PR #2118, opened 2022-09-02
Restrict terms list param type

VincentLanglet:
I'm getting 
```
[1:51] [terms_lookup] unknown field [0] [reason: [1:55] [bool] failed to parse field [must]]
```
error when I passed something like
```
[0 => 3, 2 => 4]
```
to the Query\Terms.

As soon as I use `array_values` on the array I passed, the error is getting away.

I see two solutions for this:
- Restrict the phpdoc to `list<bool|float|int|string>`
- Calling `array_values` inside the setTerms` method.

What do you prefer ? @ruflin

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 * Removed `CallbackStrategyTestHelper` and `ErrorsCollector` from `tests` [#2111](https://github.com/ruflin/Elastica/pull/2111)
 ### Fixed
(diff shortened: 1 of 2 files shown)


ruflin:
I like the phpdocs change as it seems mainly we got the docs wrong. Adding `array_values` would mainly expand the magic the function does and I'm normally not a big fan. Lets have just one mode of adding the data.

@VincentLanglet Thanks for the update, any chance you could also add a note to the changelog file?

VincentLanglet:
> @VincentLanglet Thanks for the update, any chance you could also add a note to the changelog file?

done

ruflin:
Thanks!

PR #2125, opened 2022-11-26
Prevent throwing JSONParseException when the response is empty.

aribon-cs:
I saw this issue when I wanted to check whether an index exists in the elastic.
Elastica uses HEAD requests to check whether the index exists or not, so the response body is empty. 
This issue is handled inside the code by try-catch (getData function). But there were two reasons for this PR:
1-  by this change, we can check if the response is empty, don't call the JSON parse function. So no exception will occur, and there isn't any need to catch it.
2- we use newrelic for distributed tracing and see many errors. ( newrelic agent collects all exceptions even though they are handled, and we want to keep this ability for better tracing )


ruflin (approved):
Thanks for the cleanup! I wonder why we didn't write it this way in the first place 🤔 I'll wait on CI to complete and then get it in.

PR #2127, opened 2022-11-30
Update(CI) Update actions/checkout to v3

thePanz:


ruflin (approved):
LGTM to me. Can you add a note why we need to switch to v3. Any feature missing or just to make sure we are up-to-date?

thePanz:
> LGTM to me. Can you add a note why we need to switch to v3. Any feature missing or just to make sure we are up-to-date?

The pipelines running with v2 were flagged with a notice, due to the outdated NodeJS 12 (which is used to run actions/checkout@v2)

ruflin (approved):
Second approval :-D

PR #2128, opened 2022-12-02
PHP 7.3 is EOL

frederikbosch:


ruflin:
Change LGTM. Can you add an entry in the changelog file?

deguif:
This package should be dropped when PHP version in requirements is at least 7.3 or greater, currently it's still `7.2`.
Would you mind updating the requirements too and drop tests configuration accordingly?

frederikbosch:
Maybe drop every EOL version?

sidz:
Looks like this PR can be closed. as we merged https://github.com/ruflin/Elastica/pull/2131

PR #2131, opened 2022-12-05
Drop PHP support <8.0

sidz:
This PR drops support for PHP <8.0 as discussed in https://github.com/ruflin/Elastica/discussions/1961


ruflin (approved):
I'm curious how much push back we will get on this as we are basically only 8 days after the EOL of 7.4. But I think it is the right decision. If we have many users with lots of issues, it is easier to add it later again as removing is the breaking change.

Screenshot 2022-12-06 at 09 26 33

Does anyone know if there are some numbers out there on how well users migrated to 8?

sidz:
@ruflin I suppose we could rely on packagist statistics: https://packagist.org/php-statistics.
There are more users with PHP 8.1 rather than 8.0 . Migration from 8.0 to 8.1 is smth like from 7.0 to 7.1

.github/workflows/continuous-integration.yaml
```diff
@@ -31,20 +31,12 @@ jobs:
...
           - '8.1'
         dependencies:
           - 'highest'
         elasticsearch:
           - '7.15.2'
-        include:
-          # Test with the lowest set of dependencies
-          - dependencies: 'lowest'
-            php: '7.2'
-            elasticsearch: '7.15.2'
```
franmomu: I think this should be modified instead of removed, the idea of this job is to run the tests with the lowest set of dependencies.

Let's say we require `elasticsearch/elasticsearch: ^8.0` package, this job would use `8.0` to run the tests, otherwise we would only test with the latest release (right now `8.5.2`)

sidz: Yeah the idea is clear but currently (in this PR) we don't have lower dependencies (except probably psr/log). We could reintroduce testing against lower dependencies soon when `elasticsearch/elasticsearch` will be updated and we decide which minimum version of it we will use

ruflin:
Oh, 7.4 is still at around 28% but good to see it is decreasing ... 8.* is about 60% which is promising.

.github/workflows/continuous-integration.yaml
ruflin: Maybe a compromise here is to comment it out for now and leave a comment. I'm a bit worried we might forget to readd it.

sidz: reverted

ruflin: thx

ruflin:
@franmomu Can you have another look?

deguif (approved):

ruflin:
@sidz One thing I only realised now: Changelog is missing. I would argue this is worth to be mentioned :-D 

@deguif We should also update the README page: https://github.com/ruflin/Elastica

sidz:
@ruflin Changelog amended

ruflin:
Thanks @sidz

PR #2132, opened 2022-12-07
Remove deprecated classes

franmomu:
This PR removes all deprecated classes in `7.x` branch, there is still deprecated code to be removed/changed but I think it will be easier to review in another PR.


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/7.3.0...master)
 ### Backward Compatibility Breaks
 * Dropped support for PHP <8.0 [#2131](https://github.com/ruflin/Elastica/pull/2131)
+* Removed deprecated classes in 7.x [#2132](https://github.com/ruflin/Elastica/pull/2132)
```
ruflin, line 10: I know the list might be long but I think it is worth for our users to list the classes in the changelog. Same for methods and variables that were removed. Like this, users don't have to look at the PR but only changelog to see what was removed.

If we want to go even one step further, we could add use now. Something like:

* Elastica\Processor\Convert removed -> use Elastica\Processor\ConvertProcessor instead

franmomu: 👍 done @ruflin

ruflin:
Thanks!

PR #2133, opened 2022-12-07
Remove json utils methods and json custom exception

deguif:


sidz:
my 2 cents: what about to no remove JSON class and improve it instead to smth like:

```
class JsonEncoder
{
    private function __construct()
    {
    }

    public static function encode(mixed $value, int $flags = null, int $maxDepth = 512): string
    {
        $flags ??= \JSON_HEX_TAG | \JSON_HEX_APOS | \JSON_HEX_AMP | \JSON_HEX_QUOT | \JSON_PRESERVE_ZERO_FRACTION;

        try {
            return json_encode($value, $flags | \JSON_THROW_ON_ERROR, $maxDepth);
        } catch (\JsonException $e) {
            throw new InvalidArgumentException('Invalid value for "json" option: '.$e->getMessage());
        }
    }
```

and the same for `json_decode`.

The idea is pretty simple - tons of `json_encode(decode)` around the codebase

ruflin:
@sidz @deguif Thoughts on a path forward here for 9.x?

ruflin:
Going to close this as too many conflicts by now. Lets reopen fresh if still needed.

PR #2134, opened 2022-12-07
Remove deprecated index creation options

deguif:


sidz:
I suppose that changelog need to be amended in this case as well

ruflin:
Yep, changelog please. And lets be verbose in the changelog so anyone migrating to 8 will not only know what is removed but if possible also how to migrate to it without having to jump to the PR.

franmomu:
@ruflin this can be closed, it was merged in https://github.com/ruflin/Elastica/pull/2147

PR #2136, opened 2022-12-14
Allow PHP 8.2

franmomu:
Adds support for PHP 8.2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   * `Elastica\QueryBuilder\DSL\Query::common_terms()` -> use `Elastica\QueryBuilder\DSL\Query::match()` instead
   * `Elastica\Transport\HttpAdapter` -> use `Elastica\Transport\Http` instead
 ### Added
+* Added support for PHP 8.2 [#2136](https://github.com/ruflin/Elastica/pull/2136)
 ### Changed
 ### Deprecated
 ### Removed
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -32,9 +31,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   * `Elastica\QueryBuilder\DSL\Query::common_terms()` -> use `Elastica\QueryBuilder\DSL\Query::match()` instead
   * `Elastica\Transport\HttpAdapter` -> use `Elastica\Transport\Http` instead
 ### Added
+* Added support for PHP 8.2 [#2136](https://github.com/ruflin/Elastica/pull/2136)
 ### Changed
 ### Deprecated
 ### Removed
+* Dropped support for PHP <8.0 [#2131](https://github.com/ruflin/Elastica/pull/2131)
```
franmomu: I've moved this entry since it is not really BC Break, if the user is using PHP < 8.0 it won't be able to use this version since composer won't allow it.

ruflin: I would still keep it under breaking change. Imagine a user goes to the changelog for 8.0 and wants to see what the breaking changes are. Nothing around 8.0 in it and then tries to deploy it but can't. Agree, the technical bits are in place but I hope users read the changelog before.

franmomu: I've reverted this change since it's not related to this PR, I hope users try their projects with the PHP version used in their servers, but well you'll never know 🙈

ruflin:
Thx

PR #2138, opened 2022-12-17
Remove aggregation deprecations

franmomu:
I would go a step further (in another PR) and remove setters or make them private for mandatory data passed in the constructor like `setBucketsPath()` or `setScript()`.

### DIFF
--- a/src/Aggregation/BucketScript.php
+++ b/src/Aggregation/BucketScript.php
@@ -2,8 +2,6 @@
 
 namespace Elastica\Aggregation;
 
-use Elastica\Exception\InvalidException;
-
 /**
  * Class BucketScript.
  *
@@ -13,17 +11,12 @@ class BucketScript extends AbstractAggregation implements GapPolicyInterface
 {
     use Traits\GapPolicyTrait;
 
-    public function __construct(string $name, ?array $bucketsPath = null, ?string $script = null)
+    public function __construct(string $name, array $bucketsPath, string $script)
     {
         parent::__construct($name);
 
-        if (null !== $bucketsPath) {
-            $this->setBucketsPath($bucketsPath);
-        }
-
-        if (null !== $script) {
-            $this->setScript($script);
-        }
+        $this->setBucketsPath($bucketsPath);
+        $this->setScript($script);
     }
 
     /**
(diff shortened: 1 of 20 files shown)


src/Aggregation/BucketScript.php
```diff
@@ -13,17 +11,12 @@ class BucketScript extends AbstractAggregation implements GapPolicyInterface
 {
     use Traits\GapPolicyTrait;
 
-    public function __construct(string $name, ?array $bucketsPath = null, ?string $script = null)
+    public function __construct(string $name, array $bucketsPath, string $script)
     {
         parent::__construct($name);
 
-        if (null !== $bucketsPath) {
```
ruflin, line 20: This one we did not deprecate. 

One thing I've seen in the past is that an object was created when the bucketPath did not exist yet and was set at a later stage (just an example). Is this a use case we want to support or on this scenario we except the object to be created when all info is available?

src/Aggregation/BucketSelector.php
```diff
@@ -11,17 +11,12 @@ class BucketSelector extends AbstractSimpleAggregation implements GapPolicyInter
 {
     use Traits\GapPolicyTrait;
 
-    public function __construct(string $name, ?array $bucketsPath = null, ?string $script = null)
+    public function __construct(string $name, array $bucketsPath, string $script)
     {
         parent::__construct($name);
 
-        if (null !== $bucketsPath) {
```
ruflin, line 18: Same as above. Do we need to deprecate it in 7.x branch if we go down this path?

ruflin (commented):
I like how this simplifies the code base. Left 2 high level comments to have some discussion.

src/Aggregation/BucketScript.php
franmomu: Yeah, I would deprecate them in `7.x`.

When working with ES, these parameters are required:

https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-pipeline-bucket-script-aggregation.html
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-pipeline-bucket-selector-aggregation.html

Since they are required, I think it makes sense to not be able to create a `BucketScript` instance with invalid data.

IMHO we should avoid supporting those scenarios where these classes can contain invalid data (in some of them we can even mark them as [immutable](https://phpstan.org/writing-php-code/phpdocs-basics#immutable-classes)), it simplifies the code and ease the maintenance since these objects would always be in a valid state.

ruflin: Lets call this a decision that all required params have to be set when creating an object.

Can you open a PR against 7.x with a deprecation?

ruflin (approved):

ruflin:
As discussed above, would be great to open PR's against 7.x with the deprecations.

PR #2140, opened 2023-01-04
Deprecate constructing BucketScript and BucketSelector with invalid data

franmomu:
Reference: https://github.com/ruflin/Elastica/pull/2138#discussion_r1060332200

These parameters will be mandatory in 8.0


ruflin (approved):
LGTM

PR #2142, opened 2023-01-15
PHP8: Use match expression

franmomu:
Since we are using PHP 8 we can modernize the code a bit.

This has been done with `rector` locally.


ruflin (approved):
LGTM

Not sure why codecov complains.

PR #2147, opened 2023-02-26
Remove deprecated index creation options

franmomu:
Replaces https://github.com/ruflin/Elastica/pull/2134

Using the work done in https://github.com/ruflin/Elastica/pull/2134 and fixed issues found by phpstan


ruflin (commented):
Change LGTM. I'm less concerned about this change here then in https://github.com/ruflin/Elastica/pull/2148 because I expect `create` to be used less but also lets wait on the conclusion in https://github.com/ruflin/Elastica/pull/2148 before merging this one to potentially have it aligned.

Code LGTM.

franmomu:
I think in this case is fine like this since it was deprecated to pass `null`, so the user has probably already changed its code.

ruflin (approved):
Lets get this in. If there is a lot of push back we can still enable it later again.

PR #2148, opened 2023-02-26
Remove deprecations passing $options as int

franmomu:
The deprecation was only in `Search::setOptionsAndQuery()` method, but it affected the other ones.


ruflin (commented):
This is one of the changes I wonder how big the impact will be. I like the change but I wonder if we should still allow `null` to be passed in:

```
$index->search('hans', null, Request::GET);
```

I expect that is something quite a few will use instead of 

```
$index->search('hans', [], Request::GET);
```

Our code gets cleaner by removing it but `search` is likely one of the most used functions in the library. So changing it can have a big impact. Thoughts?

franmomu:
thanks for the review! In this case the only deprecation was to pass an `int`, so I think it makes sense to allow `null`. I guess I removed it because I did this one after https://github.com/ruflin/Elastica/pull/2147

PR #2152, opened 2023-03-22
Add missing @throws annotation to Client::request and related methods

VincentLanglet:
When using `PHPStan`, it can report exceptions which are not caught if and only if the library is correctly using the `@throws` annotations. Same for `PHPStorm`.

The `Client::request` method is throwing ConnectionException, ClientException and ResponseException, so it should be propagated to the phpdoc of method using `Client::request` in order to help the developper to catch such exceptions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 * Added support for PHP 8.2 [#2136](https://github.com/ruflin/Elastica/pull/2136)
+* Added missing `@throws` annotations [#2152](https://github.com/ruflin/Elastica/pull/2152)
 ### Changed
 * Updated `php-cs-fixer` to `3.13.2` [#2143](https://github.com/ruflin/Elastica/pull/2143)
 * Modernize code using `match` expression [#2141](https://github.com/ruflin/Elastica/pull/2141)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,7 +68,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 * Added support for PHP 8.2 [#2136](https://github.com/ruflin/Elastica/pull/2136)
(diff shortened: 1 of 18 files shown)


VincentLanglet:
@ruflin Hi, I opened the PR on the 7.x branch because this is the currently stable version. Not sure if it should be done only on 8.x or if it should be merged on 7.x and then cherry-pick/merged on 8.x.

ruflin:
I prefer to have PR's against 8.x and then backported to 7.x. The reason is that this ensures no feature is missing in 8.x. Lets say this PR goes into 8.x and for whatever reason gets stuck in 7.x, users would not have a degraded experience going to 8.x, the other way around they do.

I assume not really much code has changed around this in 8.x so I hope you can just retarget it to 8.x?

Could you also add a changelog entry?

VincentLanglet:
Thanks for your quick answer @ruflin, it targets 8.x now

ruflin:
@VincentLanglet Change LGTM. Can you add a changelog entry?

CHANGELOG.md
```diff
@@ -68,6 +68,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 * Added support for PHP 8.2 [#2136](https://github.com/ruflin/Elastica/pull/2136)
+* Added missing `@throws` annotations [#2152](https://github.com/ruflin/Elastica/pull/2152)
```
ruflin:
```suggestion
* Added missing `@throws` annotations to Client::request and related methods [#2152](https://github.com/ruflin/Elastica/pull/2152)
```

ruflin: Nit: Added a bit more details so reading only the changelog should users give an understanding on what changed.

VincentLanglet:
Thanks for the review @ruflin, changes applied

ruflin:
@VincentLanglet Thanks for the contribution. Please open a backport to 7.x so we have it there too.

VincentLanglet:
> @VincentLanglet Thanks for the contribution. Please open a backport to 7.x so we have it there too.

Done in https://github.com/ruflin/Elastica/pull/2153

PR #2155, opened 2023-04-04
added more versions of ES

oleg-andreyev:


oleg-andreyev:
@ruflin one more PR that's adding more versions to CI

.github/workflows/continuous-integration.yaml
```diff
@@ -48,7 +48,19 @@ jobs:
             elasticsearch: '7.15.2'
```
ruflin: We should also update this to 7.17.9

.github/workflows/continuous-integration.yaml
```diff
@@ -48,7 +48,19 @@ jobs:
...
+            elasticsearch: '8.1.3' # there are some bc in minor version https://www.elastic.co/guide/en/elasticsearch/reference/current/migrating-8.1.html#breaking-changes-8.1
+            experimental: false
+          - php: '8.1'
+            elasticsearch: '8.5.3' # there are some bc in minor version https://www.elastic.co/guide/en/elasticsearch/reference/current/migrating-8.5.html
+            experimental: false
+          - php: '8.1'
+            elasticsearch: '8.7.0'
+            experimental: false
+          - php: '8.1'
+            elasticsearch: '8.8.0-SNAPSHOT' # future version, allow to fail https://www.elastic.co/guide/en/elasticsearch///reference/master/release-notes-8.8.0.html
```
ruflin, line 63: One thing to keep in mind, 8.8. could contain a breaking change that is still fixed before it goes stable. I like the idea of testing against snapshots at the same time these can be very flaky so I would rather not do it for the lib.

ruflin (approved):
This looks great. Could you also mention this in the changelog. I know, it only changes tests but it is still good to know for everyone that the test matrix has been updated!

.github/workflows/continuous-integration.yaml
oleg-andreyev: This step is marked as experimental so it's allowed to fail, and that was my goal.

ruflin: I was actually more worried about the parts were tests had to be adjusted for 8.8 because some thing changes (which might get reverted)

ruflin:
@oleg-andreyev Even though I approved, I would still like the changelog to be added ;-)

.github/workflows/continuous-integration.yaml
oleg-andreyev: I can revert it, it's not a problem, as long as we collect and see them (issues)

.github/workflows/continuous-integration.yaml
oleg-andreyev: updated

.github/workflows/continuous-integration.yaml
ruflin: Nah, keep it in. We can always cleanup afterwards. The docs you added are really good.

ruflin:
Thanks @oleg-andreyev

Issue #1777, opened 2020-05-14
Trailing slash in host causes issues


BackEndTea:
When i use elastica, and configure my host to be `elastic.dev` everything is fine, but if i use `elastic.dev/` i'm unable to connect.

Maybe this normal for elastic, but it wasn't something i expected. My best guess would be that an `rtrim` on the host could fix this.

ruflin:
I remember some historical discussion around and there was a reason we did not trim. Unfortunately couldn't find it.

In general I'm in favor of no magic code but good errors. So I rather show an error the first time the config is not correct instead of fixing a potentially invalid config which might brake at a later stage.

fabianaromagnoli:
I experienced the opposite issue: when host is configured with trailing slash, everything works fine, while if the trailing slash is missing, I get a "Malformed URL" exception.

@ruflin What about trimmig and re-adding a slash, in order to be compliant with both host with and without trailing slash? Something like `$baseUri = rtrim($baseUri, '/') . '/';` in \Elastica\Transport\Http::exec

Sometimes you cannot simply change the host in your configuration, for instance when you use [symfony local webserver and you have a docker integration](https://symfony.com/doc/current/setup/symfony_server.html#docker-integration):

> When using the Symfony binary with php bin/console (symfony console ...), the binary will always use environment variables detected via Docker and will ignore local environment variables.[...]

In this case you have no power to change the value of the env variable.

ruflin:
@fabianaromagnoli Possible, but sounds like even more magic :-)

In the Docker case, who sets these environment variables? Who can change it?

jhuebner79:
Trailing dots in hostnames, very well a valid choice, also cause 403 from the remote end, when using signed requests towards AWS ElasticSearch domains.

Why is it a valid choice? To reduce DNS load in busy clusters as described here: https://medium.com/cloutive/production-ready-eks-coredns-configuration-6fea830606f8 - Solution 3

Should I create a separate issue? @ruflin

ruflin:
@jhuebner79 What is your preferred solution here?

jhuebner79:
I'd be the happiest if Elastica stripped the trailing dot during building the signed request, hoping it'll fix the issue. We strongly suspect this to be the reason for the 403s we see when using the trailing dot in the hostname given to Elastica.

jhuebner79:
[...]

PR #2161, opened 2023-05-16
Allow base url without ending slash

jaroslavtyc:
_Accept both with and without trailing slash URL as Elastic source._

I have spent few time to find out what is wrong with Elastic URL `https://foo:bar@elastic-dev.example.com`, getting error `Couldn't resolve host`
```
./bin/console fos:elastica:create

Creating bitbag_shop_product

In Http.php line 186:
                         
  Couldn't resolve host  
                         

fos:elastica:create [--index [INDEX]] [--no-alias]
```

Reason was missing trailing slash in URL, leading into invalid URL `https://foo:bar@elastic-dev.example.combitbag_shop_product` instead of correct `https://foo:bar@elastic-dev.example.com/bitbag_shop_product`

Suggested change ensures slash between base URL and request path (Elastic index name in this case).

### DIFF
--- a/src/Transport/Http.php
+++ b/src/Transport/Http.php
@@ -65,7 +65,10 @@ class Http extends AbstractTransport
             $requestPath = Util::escapeDateMath($requestPath);
         }
 
-        $baseUri .= $requestPath;
+        if ($requestPath !== '') {
+            $baseUri = rtrim('/', '');
+            $baseUri .= '/' . $requestPath;
+        }
 
         $query = $request->getQuery();
 


ruflin:
This reminds me of https://github.com/ruflin/Elastica/issues/1777 Can you check where we landed there? Need to read up on it again.

jaroslavtyc:
@ruflin In that issue I have found opposite problems

1. it is broken without trailing slash https://github.com/ruflin/Elastica/issues/1777#issuecomment-735692398
2. it is broken with trailing slash https://github.com/ruflin/Elastica/issues/1777#issue-617992054

My proposal is to make sure there is slash if and only if there will be path appended to host. It can be redundant if you solve somehow the https://github.com/ruflin/Elastica/issues/1777, but it will solve without any detect-server-and-client-expectations magic single problem now and will be harmless later.

ruflin:
Thanks for the clarification. I couldn't really think of a reason why there would ever be no `/` be needed in this situation except the path put in would start with a `/`. But then again, users don't put paths in but `indexNames` for example where `/` is not an allowed character. But now things become interesting: All tests fail so there seems to be a pretty strong side effect of this. Can you take a closer look on what happens?

Please also add a changelog entry.

PR #2170, opened 2023-05-31
Elasticsearch 8.8.0 stable

oleg-andreyev:


ruflin (approved):
Change LGTM.

We should discuss how we keep the README here up-to-date: https://github.com/ruflin/Elastica/blob/8.x/README.md?plain=1#L22 Should we just remove it and point to the workflow file?

PR #2175, opened 2023-06-16
Add more missing @throws tags

VincentLanglet:
Hi, @ruflin 

This is a follow up of https://github.com/ruflin/Elastica/pull/2152
I found some missing @throws tag.

get/set methods are throwing exceptions but I forgot to add the tag to related get/set methods.


ruflin:
thx!

PR #1971, opened 2021-08-02
Allow psr/log 2.0 + 3.0

Jean85:
`psr/log` 2.0 was tagged a couple of weeks ago: https://twitter.com/phpfig/status/1415413578191024130

3.0 has been tagged too, and it can be allowed since the interface us just used but not implemented anywhere.
This PR is also a blocker for https://github.com/Seldaek/monolog/pull/1564

More details are available in the new meta document: https://www.php-fig.org/psr/psr-3/meta/

PR #2183, opened 2023-12-06
[6.x] Allow psr/log v2 and v3

aistis-:
Seems [we allow psr/log v2 or v3](https://github.com/ruflin/Elastica/pull/1971) to elastica v7+. This is a backport support for elastica v6. No changes from the code itself is necessary.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     "require": {
         "php": "^7.0||^8.0",
         "ext-json": "*",
-        "psr/log": "~1.0",
+        "psr/log": "^1.0||^2.0||^3.0",
         "elasticsearch/elasticsearch": "^6.0"
     },
     "require-dev": {


ruflin:
Thanks for the contribution @aistis- I'm happy to get this in but please we aware that there are no other releases planned for 6.x at the moment.

ruflin:
@aistis- Some of the CI tests are failing. I don't think this is because of your change but because some parts in the geo lib have been updated. It needs an adjustment in at least on of the test for the geo values.

aistis-:
Hi @ruflin and thanks for feedback! Should I look into the failing tests and try to fix them? Indeed, they seems broken, but not because of added `psr/log v2/3` support.

ruflin:
> Should I look into the failing tests and try to fix them?

That would be great! I hope it is mostly about adjusting some values.

ruflin:
Going to close this as there hasn't been an update recently. But happy to discuss again and get it in.

Issue #2182, opened 2023-12-01
Does RetryOnConflict suppose to work with Bulk updates


csabavirag:
I have an application which updates documents in bulk through multiple workers. I'd like to use the `retry_on_conflict` feature, but even if I have the setting at the ClientConfiguration, `Bulk::addDocuments()` does not seem to add the necessary piece to the Document metadata in order to get the proper update action as it is described at https://www.elastic.co/guide/en/elasticsearch/reference/7.17/docs-bulk.html#bulk-update

They say, every update should contain the `retry_on_conflict` attribute like this:
```
POST _bulk
{ "update" : {"_id" : "1", "_index" : "index1", "retry_on_conflict" : 3} }
{ "doc" : {"field" : "value"} }
{ "update" : { "_id" : "0", "_index" : "index1", "retry_on_conflict" : 3} }
{ "script" : { "source": "ctx._source.counter += params.param1", "lang" : "painless", "params" : {"param1" : 1}}, "upsert" : {"counter" : 1}}
{ "update" : {"_id" : "2", "_index" : "index1", "retry_on_conflict" : 3} }
{ "doc" : {"field" : "value"}, "doc_as_upsert" : true }
{ "update" : {"_id" : "3", "_index" : "index1", "_source" : true} }
{ "doc" : {"field" : "value"} }
{ "update" : {"_id" : "4", "_index" : "index1"} }
{ "doc" : {"field" : "value"}, "_source": true}
```

I only get the `_id` and `_index` values added to the document metadata. What do I do wrong?

ruflin:
Can you share the exact code snippet you are using and the version of Elastica? There was in one version a change from `_retry_on_conflict` to `retry_on_conflict`. A quick search, i found there is a for a similar scenario you describe here: https://github.com/ruflin/Elastica/blob/0f6b04b5ebe6df51015fb54b4c455bd7af5b093f/tests/BulkTest.php#L694

csabavirag:
Actually I'm using FOSElasticaBundle and they have added support for RetryOnConflict parameter per connection (https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/965) long time ago.

My config looks like:
``` yaml
fos_elastica:
    clients:
        default: { url: '%env(ELASTICSEARCH_URL)%', retryOnConflict: 5 }
    indexes:
```

which sets 

[...]

PR #2184, opened 2024-01-16
Apply RetryOnConflict on Bulk documents

csabavirag:
This PR is created to solve the issue on the missing RetryOnConflict (retry_on_conflict) metadata attribute when adding documents in bulk (detailed at https://github.com/ruflin/Elastica/issues/2182)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Backward Compatibility Breaks
 ### Added
+* If not expicitly set, use `retry_on_conflict` from Client configuration in Bulk updates [#2184](https://github.com/ruflin/Elastica/pull/2184)
 ### Changed
 ### Deprecated
 ### Removed
--- a/src/Bulk.php
+++ b/src/Bulk.php
@@ -131,8 +131,8 @@ class Bulk
      */
     public function addDocument(Document $document, ?string $opType = null): self
     {
-        if ($this->_client->getConnection()->hasParam('retryOnConflict')) {
-            $document->setRetryOnConflict($this->_client->getConnection()->getParam('retryOnConflict'));
+        if (!$document->hasRetryOnConflict() && $this->_client->hasConnection() && ($retry = $this->_client->getConnection()->getParam('retryOnConflict')) > 0) {
+            $document->setRetryOnConflict($retry);
         }
 
         $action = AbstractDocumentAction::create($document, $opType);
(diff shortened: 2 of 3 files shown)


ruflin (commented):
Thanks for the PR. Change LGTM. Could you add an entry to the changelog? And can you check in the test suite if there is an easy way to add a test for this?

Normally we make changes to 8.x first and then backport to 7.x. But currently there is quite a large PR ongoing against 8.x to make it 8 compatible: https://github.com/ruflin/Elastica/pull/2181 What I propose is to get this in but in parallel follow up with a PR against the 8.x branch to ensure we don't forget about it. Otherwise I worry we will miss this feature in 8.x.

ruflin:
@csabavirag I did just trigger the test suite and it seems to break one of the tests. It seems that in some cases the object is `null` so you need to add a check for this.

csabavirag:
I pushed a new commit with does not set RetryOnConflict if it has already been set on the document/script. I know `retry_on_conflict` is stripped from the final message(s) if its value is `0`, but thought to exclude this case in order to a smaller memory footprint.
This allowed me to run BulkTest.php successfully. Only 1 error was reported, but I believe this is not related to these changes

Output of `make docker-run-phpunit PHPUNIT_OPTIONS="tests/BulkTest.php"`:
```
There was 1 failure:

1) Elastica\Test\BulkTest::testMemoryUsage
Failed asserting that 2.3990734932979025 is less than 1.31.
```

I also updated the `BulkTest` suite to check the parameter is set on the document/script properly when only the client has `retry_on_conflict` set ("global option").

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Backward Compatibility Breaks
 ### Added
+* If not expicitly set, use `retry_on_conflict` from Client configuration in Bulk updates [#2184](https://github.com/ruflin/Elastica/pull/2184)
```
ruflin, line 11: Should we put this under "fixed"? sounds more like a bug fix but also happy to keep it under Added.

ruflin (approved):
Change LGTM, waiting on CI to run through.

CHANGELOG.md
csabavirag: First I added to that section. Then I moved to Added. TBH, I would leave it up to you.

ruflin: Lets keep it, I can always clean it up later.

ruflin:
@csabavirag Thank you for the contribution!

Could you follow up with a PR against 8.x. The big 8.x change made it in and we should make sure this change also makes it into 8.x

PR #2188, opened 2024-01-21
Re-work connections and connection pool in favor of Node Pool

sidz:
This PR re-works Connection Pool, Connections and Strategies in favor of [Node Pool](https://github.com/elastic/elastic-transport-php#node-pool) from the official client.

As this package is built on top of official client we have to use Node Pool instead of our custom connections mechanism. What makes the whole package a bit simplier to maintain. In the future this could be re-worked again if needed. We would liek to release new version of this package as fast as possible.

### ConnectionPool is replaced with Node Pool

More details can be found via link: https://www.elastic.co/guide/en/elasticsearch/client/php-api/8.12/node_pool.html.

By default Official elasticsearch client builds Simple Node Pool with `RoundRobin` selector and `NoResurrect` resurrect. 

To override default Node Pool behavior do the following:
```php
use Elastic\Transport\NodePool\Resurrect\ElasticsearchResurrect;
use Elastic\Transport\NodePool\Selector\RoundRobin;
use Elastic\Transport\NodePool\SimpleNodePool;

$nodePool = new SimpleNodePool(
    new RoundRobin(),
    new ElasticsearchResurrect()
);

new Client([
    'hosts' => [
        'https://node1.com:9200',
    ],
    'transport_config' => [
        'node_pool' => $nodePool,
    ],
]);
```

### ClientConfiguration changes

`ClientConfiguration` class is a bit reworked and the following parameters have been removed:
- `port`
- `path`
- `url`
- `connections` in favor of `hosts`
- `servers` in favor of `hosts`
- `roundRobin`

`host` parameter has been renamed to `hosts` and it's should be an array of strings.

`elastic/transport` is updated to 8.8 as since that version each host could be configured with its own username/password credentials.

How to configure authentication per node.
```php
new Client([
    'hosts' => [
        'https://username1:password1@node1.com:9200',
        'https://username2:password2@node2.com:9200',
    ],
]);
```

### No longer needed dependecies

- `symfony/deprecation-contracts`
- `guzzlehttp/psr7`
- `nyholm/dsn`


tests/ClientTest.php
```diff
@@ -104,25 +104,6 @@ public function testConfigValue(): void
         $this->assertIsArray($client->getConfigValue(['level1', 'level2']));
     }
 
-    public function testAddHeader(): void
```
ruflin, line 107: Should we keep this test but adjust it to how users will add headers in the future?

ruflin:
I like the simplification / reduced code base this introduces. One thing that would be useful for the discussion is @sidz is some examples on how users would do it in 7.x and how users would do some of the connection / node pool changes in the future. This should also help guide users that need to migrate to the new behaviour.

sidz:
Yeah, I'll try to add some information as part of this request @ruflin. 

All what we called connections now is: `hosts` and as I can see user is not able to configure multiple connections with different username/password or API Keys in the official client or I didn't find how to do it yet so we are not able to do it either.

pawelkeska:
@sidz @ruflin i left `ConnectionPoll` from elastica becasue as @sidz writes in `elasticsearch-php ` library we are not able to set multiple connections with diffrent username/password. In the current version on branch 8.x we can do this by set connections like that:
```
['connections'] => [
    [
       'host' => 'host1',
       'username' => 'example1'
       'password' => 'password1',
       'transport_config' => and here we can set node poll for this connection
    ],
     [
       'host' => 'host2',
       'username' => 'example2'
       'password' => 'password2',
        'transport_config' => and here we can set node poll for this connection
    ],
]
```

In `transport_config` key we can set additionally configuration [Node Poll](https://github.com/elastic/elastic-transport-php#node-pool) for this connection. I think is more flexible for multiple connections with diffrent username/password.

sidz:
Hey @pawelkeska. Thanks for the explanation.

I've looked at what official transport provides and it seems like it is possible to use username and password as part of URI (sense 8.8.0 https://github.com/elastic/elastic-transport-php/pull/22) in the following way:
```php
$builder = TransportBuilder::create();
$build->setHosts([
    'http://user1:password1@node1.com',
    'http://user2:password2@node2.com',
]);
```

What is made life a bit more easier.


Current solution will work either but our **Connection** is not a fair single connection/node (or maybe I've missed smth) for a case where amounf of `hosts` more than one.

If you will open [Transport class](https://github.com/elastic/elastic-transport-php/blob/main/src/TransportBuilder.php#L108-L111) you will see that when we set an array of hosts to the Transport it will set all of them to the NodePool. So NodePool builds so many Nodes as we pass hosts to the Transport and each Node has it's own URI.

When we call `sendRequest` on our Client: https://github.com/ruflin/Elastica/blob/8.x/src/Client.php#L644 we will call Transports `sendRequest` method which works with its NodePool inside of the transport.

And Transport will try with all Nodes defined in the NodePool in the context of our one Connection. So here we have a case when our one Connection works with all Nodes/Hosts.

Ping me if it is not clear. I'll try to give more examples or re-phrase.

pawelkeska:
@sidz Thanks for finding a solution. I haven't seen this PR [elastic/elastic-transport-php#22](https://github.com/elastic/elastic-transport-php/pull/22) before and this is good information for us. In this case we could remove our connection pool.

Yes, I agree with you, now is not a fair single connection/node and within one our connection pool we could set many hosts which  build so many Nodes as we pass hosts to the Transport and each Node has it's own URI.

If we rebuild this and remove our connection poll is the best solution because we will rely on `elastic-transport-php` only.

Could you please make corrections in tests and prepare it for merging ?

sidz:
This PR requires a bit more refactoring. I'll try to do it ASAP @pawelkeska.

pawelkeska:
@sidz Thank you for your commitment, it will help us a lot.

ruflin:
Great conversation, not really much to add on my end. My only request is that in the end we have some notes / quicks on how Elastica users can migrate to the new mehtods.

composer.json
```diff
@@ -17,12 +17,9 @@
     "require": {
         "php": "~8.0.0 || ~8.1.0 || ~8.2.0",
         "ext-json": "*",
-        "elastic/transport": "^8.4",
+        "elastic/transport": "^8.8",
         "elasticsearch/elasticsearch": "^8.4.1",
-        "guzzlehttp/psr7": "^2.0",
-        "nyholm/dsn": "^2.0.0",
```
sidz, line 23: No longer needed as we rely on `elastic/transport`.

composer.json
```diff
@@ -17,12 +17,9 @@
     "require": {
         "php": "~8.0.0 || ~8.1.0 || ~8.2.0",
         "ext-json": "*",
-        "elastic/transport": "^8.4",
+        "elastic/transport": "^8.8",
         "elasticsearch/elasticsearch": "^8.4.1",
-        "guzzlehttp/psr7": "^2.0",
```
sidz, line 22: no longer used directly

composer.json
```diff
@@ -17,12 +17,9 @@
     "require": {
         "php": "~8.0.0 || ~8.1.0 || ~8.2.0",
         "ext-json": "*",
-        "elastic/transport": "^8.4",
+        "elastic/transport": "^8.8",
```
sidz, line 20: Since 8.8 elastic/transport supports basic authentication per node. see https://github.com/elastic/elastic-transport-php/pull/22

composer.json
```diff
@@ -31,9 +28,11 @@
         "phpunit/phpunit": "^9.5",
         "symfony/phpunit-bridge": "^6.0"
     },
+    "conflict": {
+        "guzzlehttp/psr7": "<2.0.0"
```
sidz, line 32: We don't have direct dependency to `guzzlehttp/psr7` but without this hack `--prefer-lowest` installs 1.4.0 which doesn't have `psr/http-factory` and official transport is not able to establish connection (see  https://github.com/elastic/elastic-transport-php/blob/main/src/NodePool/Node.php#L33)

tests/Transport/NodePool/TraceableSimpleNodePool.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Elastica\Test\Transport\NodePool;
+
+use Elastic\Transport\NodePool\Node;
+use Elastic\Transport\NodePool\SimpleNodePool;
+
+final class TraceableSimpleNodePool extends SimpleNodePool
```
sidz, line 10: For testing purposes as original node pool doesn't have `getNodes` method so we couldn't check different cases in our tests

sidz:
Ready for review @ruflin, @pawelkeska.

While I worked on this PR I started to think that probably we need our own `ClientBuilder` as a replacement for `ClientConfiguration`.

composer.json
ruflin: I assume even if we use 8.8, Elastica will still be compatible with older versions of Elasticsearch 8.x

src/Bulk.php
```diff
@@ -133,8 +133,12 @@ public function getActions(): array
      */
     public function addDocument(Document $document, ?string $opType = null): self
     {
-        if (!$document->hasRetryOnConflict() && $this->_client->hasConnection() && $this->_client->getConnection()->hasParam('retryOnConflict') && ($retry = $this->_client->getConnection()->getParam('retryOnConflict')) > 0) {
-            $document->setRetryOnConflict($retry);
+        if (!$document->hasRetryOnConflict()) {
+            $retry = $this->_client->getConfigValue('retryOnConflict', 0);
```
ruflin, line 137: Is this directly related to the change? Maybe make it a quick second PR if not?

ruflin (commented):
Overall change LGTM, it simplifies the code.

My main concern is around users migration from 7.x to 8.x. This PR touches the configs that likely every users uses which are the client configs. As you describe, port, path, connections etc. go away / are replaced. To simplify migration, could we keep some of these configs but convert these automatically for users? If services or connectiosn exists, we convert it to `hosts`. Same if a a port exist, we add it all hosts but log a deprecation warning and let the user know, they should adjust to hosts? This change could also be done as a follow up.

composer.json
sidz: Yes

src/Bulk.php
sidz: Yes. because Client class doesn't have getConnection method so I had to find another way to get configuration value

sidz:
> To simplify migration, could we keep some of these configs but convert these automatically for users? If services or connectiosn exists, we convert it to hosts. Same if a a port exist, we add it all hosts but log a deprecation warning and let the user know, they should adjust to hosts?

IMO we need to amend previous version an add deprecation message in the previous (7.x) version instead of in the newest 8. Does it make sense @ruflin ?

ruflin:
> IMO we need to amend previous version an add deprecation message in the previous (7.x) version instead of in the newest 8.

In theory I agree. But I think the changes fall into 2 categories:

* Changes that can be done in 7.x: We can add a deprecation message and users can already migrate to the new behaviour
* Changes that users can't apply in 7.x yet: To make it available in 7.x, we would have to make significant changes to 7.x

I don't have a good understanding on which changes fall where. What can we easily add deprecation in 7.x and make migrate users there, lets do it there. For the ones we can't, we force users to make the changes on upgrade or for the low hanging fruits, lets make changes in 8.x. We can also wait with changes in 8.x and see what stands out for users during the migration and make it easier over time. But especially for configs which 100% of the Elastica users likely have, I think we need to do something.

sidz:
> But especially for configs which 100% of the Elastica users likely have, I think we need to do something.

Migration guide should work fine I think.

ruflin:
Lets get it in and work from there. We can always add things. @pawelkeska would be great to get a second eye on this. @sidz From your end, ready to get in?

sidz:
@ruflin yes. ready for review

ruflin:
@sidz Did not forget about the review but was out / busy the last two weeks. Try to get to it this week. Would be great to get some additional reviews from others like @pawelkeska @thePanz etc.

composer.json
ruflin: I wonder if there is a way we could add a note around this to the code somewhere. I'm worried someone comes along eventually and removes it and we don't see the side effects immediately. The good news is, that now if someone goes back to where it was introduced, they see these comments.

tests/Base.php
```diff
@@ -46,38 +48,38 @@ protected static function showDeprecated(): void
...
+        );
 
-        return new Client($config, $callback, $logger);
+        return new Client($config, $logger);
     }
 
     protected function _getHost(): string
     {
-        return \getenv('ES_HOST') ?: Connection::DEFAULT_HOST;
+        return \getenv('ES_HOST') ?: 'localhost';
```
ruflin, line 65: It would be nice to keep some constants but not sure where to move it. Same below.

ruflin (approved):

ruflin:
Hi @sidz , it took me much longer to get to the review then it should have. Sorry about this.

I must confess, merging this PR was harder then I expected (not because of code reasons). I'm excited about the removal of all the connection code as I will simplify the maintenance of this library. Removing connection was always one of the goals why the dependency on the official client was added and finally we are here. The changes that keep bugging me are the ones where it is breaking for most users but the removal of connections requires it. How will users get to 8 in an easy way? 

The answer for me is, the faster we get this in, the more feedback we get and the fewer users already jumped into 8.x and have these changes from day 1. I expect there will be more changes coming to 8.x and I'm looking forward to it together with the community to finally fully support 8.x

@sidz Thank you so much for hanging in there and keep pushing this forward. Will merge this PR now and we will keep iterating on top of it.

sidz:
> How will users get to 8 in an easy way?

We could add [UPGRADE.MD](https://github.com/symfony/symfony/blob/7.1/UPGRADE-7.0.md) like Symfony does and provide all steps which users should do in order to migrate from 7 to 8

ruflin:
> We could add [UPGRADE.MD](https://github.com/symfony/symfony/blob/7.1/UPGRADE-7.0.md) like Symfony does and provide all steps which users should do in order to migrate from 7 to 8

@sidz ++

PR #2190, opened 2024-01-27
Enable declare_strict_types option and apply @PHP80Migration set with…

sidz:
This PR enables `declare_strict_types` option and applies `@PHP80Migration` set with risky rules.

### DIFF
--- a/src/AbstractUpdateAction.php
+++ b/src/AbstractUpdateAction.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Elastica;
 
 use Elastica\Exception\InvalidException;
--- a/tests/Cluster/SettingsTest.php
+++ b/tests/Cluster/SettingsTest.php
@@ -40,7 +40,7 @@ class SettingsTest extends BaseTest
             $index->addDocument($doc2);
             $this->fail('should throw read only exception');
         } catch (ClientResponseException $e) {
-            $error = json_decode((string) $e->getResponse()->getBody(), true)['error']['root_cause'][0] ?? null;
+            $error = \json_decode((string) $e->getResponse()->getBody(), true)['error']['root_cause'][0] ?? null;
 
             $this->assertSame('cluster_block_exception', $error['type']);
             $this->assertStringContainsString('cluster read-only', $error['reason']);
--- a/tests/Cluster/SettingsTest.php
+++ b/tests/Cluster/SettingsTest.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Elastica\Test\Cluster;
 
 use Elastic\Elasticsearch\Exception\ClientResponseException;
--- a/tests/Cluster/SettingsTest.php
+++ b/tests/Cluster/SettingsTest.php
@@ -40,7 +40,7 @@ class SettingsTest extends BaseTest
             $index->addDocument($doc2);
             $this->fail('should throw read only exception');
         } catch (ClientResponseException $e) {
-            $error = \json_decode($e->getResponse()->getBody(), true)['error'] ?? null;
+            $error = json_decode((string) $e->getResponse()->getBody(), true)['error']['root_cause'][0] ?? null;
 
             $this->assertSame('cluster_block_exception', $error['type']);
             $this->assertStringContainsString('cluster read-only', $error['reason']);
--- a/tests/Index/SettingsTest.php
+++ b/tests/Index/SettingsTest.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Elastica\Test\Index;
 
 use Elastic\Elasticsearch\Exception\ClientResponseException;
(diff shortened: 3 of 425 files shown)


src/AbstractUpdateAction.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
```
ruflin, line 3: Is this line automatically added to each file or is this something done manually?

ruflin (commented):
Looks like this triggered quite a list of CI failures :-)

sidz:
> Looks like this triggered quite a list of CI failures :-)

Yeah. Here is a lot what need to be done.

src/AbstractUpdateAction.php
sidz: With the help of PHP-CS-Fixer `declare_strict_types` rule included in `@PHP80Migration:risky` rule set.

ruflin: Great, I was worried there would a be manual step involved.

sidz:
ready for review @ruflin

tests/Cluster/SettingsTest.php
```diff
@@ -38,7 +40,7 @@ public function testSetReadOnly(): void
             $index->addDocument($doc2);
             $this->fail('should throw read only exception');
         } catch (ClientResponseException $e) {
-            $error = \json_decode($e->getResponse()->getBody(), true)['error'] ?? null;
+            $error = \json_decode((string) $e->getResponse()->getBody(), true)['error']['root_cause'][0] ?? null;
```
ruflin, line 43: Do you know why this changed?

tests/Index/SettingsTest.php
```diff
@@ -136,8 +137,7 @@ public function testDeleteAliasWithException(): void
             $indexAlias->delete();
             $this->fail('Should throw exception because you should delete the concrete index and not the alias');
         } catch (ClientResponseException $e) {
-            $response = ResponseConverter::toElastica($e->getResponse());
-            $error = $response->getFullError();
+            $error = \json_decode((string) $e->getResponse()->getBody(), true)['error']['root_cause'][0] ?? null;
```
ruflin, line 140: What is the reason you removed the ResponseCoverter here?

tests/Query/CombinedFieldsQueryTest.php
```diff
@@ -16,10 +18,10 @@
 class CombinedFieldsQueryTest extends BaseTest
 {
     private static $data = [
-        ['id' => 1, 'title' => 'Rodolfo', 'body' => 'Moraes',   'abstract' => 'Lorem'],
-        ['id' => 2, 'title' => 'Tristan', 'body' => 'Maindron', 'abstract' => 'Dolor'],
-        ['id' => 3, 'title' => 'Monique', 'body' => 'Maindron', 'abstract' => 'Ipsum'],
-        ['id' => 4, 'title' => 'John',    'body' => 'not Doe',  'abstract' => 'Consectetur'],
+        ['id' => '1', 'title' => 'Rodolfo', 'body' => 'Moraes',   'abstract' => 'Lorem'],
```
ruflin, line 21: I assume because of line 148, the old syntax would still work?

ruflin (approved):
Change LGTM. I'll get this in directly to make sure we don't have any future conflicts. I have one question around a change in the test but we can still follow up on this separately.

ruflin:
Thanks for all the cleanup @sidz One thing I just noticed after merging, no changelog 🤦

tests/Index/SettingsTest.php
sidz: `ResponseCoverter` in my opinion should work with code related to `src` folder  (production code). Official Elasticsearch client does not return `ResponseInterface`.

I'm planning to raise a new PR where we won't allow ResponseInterface as an option to `ResponseCoverter`

tests/Query/CombinedFieldsQueryTest.php
sidz: Document id must be string or null. see https://github.com/ruflin/Elastica/blob/8.x/src/Document.php#L45

And it was added long time ago. Clients code must respect this and pass string or null only.

sidz:
> Thanks for all the cleanup @sidz One thing I just noticed after merging, no changelog 🤦

I'll add in the separate PR

tests/Cluster/SettingsTest.php
sidz: that's a good catch:
Seems like elasticsearch response looks like:
```
{
  "error": {
    "type": "cluster_block_exception",
    "root_cause": [
      {
        "type": "cluster_block_exception",
        ...
      }
    ]
  }
}
```

ruflin: I wonder why the test passed before 🤔  If I remember correctly, there was a major ES version where the response format changed, but that was quite a while ago. Good to know we have it covered in the tests.

tests/Index/SettingsTest.php
ruflin: Does this make migration from 7.x potentially harder? Lets discuss on the upcoming PR, code will make it much easier to fully understand what you mean by this.

PR #2191, opened 2024-01-28
Add Elasticsearch 8.10 - 8.12 to pipeline

sidz:
This PR adds testing against Elasticsearch versions:
- 8.10
- 8.11
- 8.12

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -63,11 +63,20 @@ jobs:
             elasticsearch: '8.7.1' # there are no bc in minor version https://www.elastic.co/guide/en/elasticsearch/reference/current/migrating-8.7.html
             experimental: false
           - php: '8.1'
-            elasticsearch: '8.8.0' # there are no bc in minor version https://www.elastic.co/guide/en/elasticsearch/reference/current/migrating-8.8.html
+            elasticsearch: '8.8.2' # there are no bc in minor version https://www.elastic.co/guide/en/elasticsearch/reference/current/migrating-8.8.html
             experimental: false
           - php: '8.1'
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yaml
```diff
@@ -63,11 +63,20 @@ jobs:
             elasticsearch: '8.7.1' # there are no bc in minor version https://www.elastic.co/guide/en/elasticsearch/reference/current/migrating-8.7.html
             experimental: false
           - php: '8.1'
-            elasticsearch: '8.8.0' # there are no bc in minor version https://www.elastic.co/guide/en/elasticsearch/reference/current/migrating-8.8.html
+            elasticsearch: '8.8.2' # there are no bc in minor version https://www.elastic.co/guide/en/elasticsearch/reference/current/migrating-8.8.html
             experimental: false
           - php: '8.1'
-            elasticsearch: '8.9.0-SNAPSHOT' # future version, allow to fail https://www.elastic.co/guide/en/elasticsearch/reference/master/release-notes-8.9.0.html
```
ruflin, line 69: Did you remove SNAPSHOT on purpose? For example there is now `8.13.0-SNAPSHOT`. 

I'm ok with removing it at the moment as development is not as quick on Elastica and removing it means more stable builds.

ruflin (approved):
LGTM. Pains me a bit that we have to test against so many versions. It helps to debug issues quickly in which version it broken. Eventually we can thin it out to have oldest (8.0.0) and newest 2 tested or similar to speed up CI time.

.github/workflows/continuous-integration.yaml
sidz: I forgot to check is `8.13-SNAPSHOT` exists or not. That's why I didn't add it.

> Eventually we can thin it out to have oldest (8.0.0) and newest 2 tested or similar to speed up CI time.

Makes sense.

PR #2204, opened 2024-03-24
Drop ES 7 from the pipeline

sidz:
This PR drops testing 8.x branch with ES 7 as we don't support it anymore

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -41,14 +41,15 @@ jobs:
         dependencies:
           - 'highest'
         elasticsearch:
-          - '7.17.9'
+          - '8.0.1'
         include:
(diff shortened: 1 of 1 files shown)


ruflin:
Can we keep the test for 7.x around until we actually break it? For migration purpose, it is kind of nice to know Elastica still works with 7.last. I suggest to add a line to the code you removed and the testing, that this is here because of backward compatibility but not officially supported. As soon as we hit a change where there is extra work to keep compatibility, we remove the code. On the docs, we should always state that it is not officially supported.

sidz:
Fine with me @ruflin. I'll amend PR later today

ruflin (approved):

ruflin:
Thanks, I adjusted the commit message a bit to make it clear we didn't drop it.

PR #2206, opened 2024-03-25
Improve ResponseConverter to work with Elasticsearch Response class

sidz:
This PR a bit improves `ResponseConverter::toElasticaResponse` method and allows to pass `Elastic\Elasticsearch\Response\Elasticsearch` class only.

ResponseInterface was passed from tests only.


sidz:
Closed for now.

PR #2211, opened 2024-04-19
Send scroll_id in body for clearScroll

SosthenG:
Sending the scroll_id in the url is [deprecated since 7.0](https://www.elastic.co/guide/en/elasticsearch/reference/current/clear-scroll-api.html#clear-scroll-api-path-params) and it's what elasticsearch-php clearScroll does when passing a "scroll_id" parameter. It should be passed in the body instead to prevent errors with big ids.


ruflin:
Thanks for the contribution. Does this need any adjustements to the test suite?

SosthenG:
I don't think so, every scroll id passed in the url should work in the body, the opposite is not true however. The tests are already checking the number of active contexts after a scroll so it should be fine.

ruflin:
@SosthenG Thanks for the contribution!

PR #2220, opened 2024-07-12
Support multi match type 'bool_prefix'

wolfsblu:
Adds support for the multi match query type `bool_prefix` [1]

[1] https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-multi-match-query.html#multi-match-types


ruflin:
Thanks for the contribution. Could you add an entry to the Changelog file? And can you check the tests if there is a way to add one to make sure it works with different Elasticsearch versions?

wolfsblu:
Added changelog and test case

ruflin (approved):
Thanks for the contribution!

PR #2224, opened 2024-08-07
Upgrade PHPUnit from 9.5 to 10.5

sidz:
<del>This PR upgrades PHPUnit from 9.5 to the latest possible version 10.5.</del>

p.s. Seems like I was too optimistic. We can't use phpunit 10 as we support PHP 8.0.
Seems like I'll update to the latest 9.6 instead 10.5


composer.json
```diff
@@ -25,8 +25,7 @@
         "guzzlehttp/guzzle": "^7.2",
         "phpstan/phpstan": "^1.5",
         "phpstan/phpstan-phpunit": "^1.1",
-        "phpunit/phpunit": "^9.5",
-        "symfony/phpunit-bridge": "^6.0"
```
sidz, line 29: symfony/phpunit-bridge is not compatible with phpunit >= 10

tests/Aggregation/BaseAggregationTestCase.php
```diff
@@ -6,6 +6,6 @@
 
 use Elastica\Test\Base;
 
-abstract class BaseAggregationTest extends Base
+abstract class BaseAggregationTestCase extends Base
```
sidz, line 9: PHPUnit shows warning message now so abstract Test Class can't have `Test` prefix

```
There was 1 PHPUnit test runner warning:

1) Class Elastica\Test\Aggregation\BaseAggregationTest declared in /home/sid/projects/Elastica/tests/Aggregation/BaseAggregationTest.php is abstract
```

ruflin:
Should we stop testing against PHP 8.0 / stop supporting it? Main reason we keep testing against 8.0 is to know when we would break something there which I hope is very rare. 

There is the other issue around the symphony bride, I assume there is work ongoing to get this to 10?

What is the path forward here? Jump to 9 and have a follow up discussion how we dump 8.0 and also go to 10?

sidz:
> Should we stop testing against PHP 8.0 / stop supporting it? 
> What is the path forward here? Jump to 9 and have a follow up discussion how we dump 8.0 and also go to 10?

As for me we need keep support and test against minimum PHP version we defined in the composer.json @ruflin 

I'll downgrade phpunit to latest 9.6. 

> There is the other issue around the symphony bride, I assume there is work ongoing to get this to 10?

As far as I'm aware some of the logic from Bridge was implemented in the PHPUnit 11, including direct/indirect notifications.

More details in the https://github.com/symfony/symfony/issues/49069

sidz:
Going to close this PR for now. As we currently run tests on phpunit 9.6

PR #2227, opened 2024-09-26
Add support for custom range key and refactor addRange in a trait

SosthenG:
IpRange and GeoDistance aggregations were missing the ability to specify a custom key for a range.
As it was already done for the Range aggregation, I moved that to a trait that is now used everywhere we need `addRange`.
For the IpRange aggregation, I also added the key option to the `addMaskRange` method.

The only drawback with this refactoring is that the arguments of the trait have to allow string|int|float|null to be usable everywhere and we loose some phpdoc details on the arguments. In my opinion, people using this should already know what format is expected and can always check the doc so I personally prefer to have the code deduplicated, but if you disagree I can change it back to add the key everywhere without using a common trait.

### DIFF
--- a/src/Aggregation/IpRange.php
+++ b/src/Aggregation/IpRange.php
@@ -4,8 +4,6 @@ declare(strict_types=1);
 
 namespace Elastica\Aggregation;
 
-use Elastica\Exception\InvalidException;
-
 /**
  * Class IpRange.
  *
@@ -14,6 +12,7 @@ use Elastica\Exception\InvalidException;
 class IpRange extends AbstractAggregation
 {
     use Traits\KeyedTrait;
+    use Traits\RangeTrait;
 
     /**
      * @param string $name  the name of this aggregation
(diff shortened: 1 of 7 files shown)


ruflin:
Hi @SosthenG Sorry for the late reply. I'm currently on holidays but try to take a look at this rather soonish.

src/Aggregation/IpRange.php
```diff
@@ -38,42 +37,21 @@ public function setField(string $field): self
...
+     * Add an ip range in the form of a CIDR mask.
      *
-     * @throws InvalidException
+     * @param string      $mask a valid CIDR mask
+     * @param string|null $key  customized key value
      *
      * @return $this
      */
-    public function addRange(?string $fromValue = null, ?string $toValue = null): self
+    public function addMaskRange(string $mask, ?string $key = null): self
```
ruflin, line 47: This is an additional change not directly tied to the trait is my understanding?

ruflin (approved):
Change LGTM. I like that the new trait removes quite a bit of duplicated code and makes sure it stays in sync.

ruflin:
Thanks for the contribution!

src/Aggregation/IpRange.php
SosthenG: Yes I did that as well because my original goal was to add support for the key option and the mask range was missing it too.

PR #2240, opened 2025-01-12
Remove unused method argument in the Elastica\SearchableInterface methods

sidz:
This PR removes:
- 3rd `$method` argument from the `Elastica\SearchableInterface::search` and `Elastica\SearchableInterface::count` as they are not unused anymore. 
- `Elastica\Request class` as it is not needed anymore.

On one hand it's a BC break as signatures are changed but on other hand it is not as we basically didn't clean up code base well before releasing version 8.

`$method`  argument is not used inside of the methods.
`Elastica\Request` class is not needed anymore as right methods are set inside of official elasticsearch SDK


ruflin:
Late reply again, sorry about that. Change LGTM. The main concern I have is that these are methods which are likely used a lot at the same time, if some would have used the params and not the default ones, it would not have had an affect. 

Maybe we can add a quick explanation to the changelog that users should remove the additional param from their code before upgrading. At the same time, this is what the BC break is for. What I try to achieve is:

* Most users that upgrade and actually read the changelog, know what to change
* The ones that upgrade and don't read the changelog and are caught by surprise, will have an easy way to find what changed and why so basically preventing questions on Github ;-)

sidz:
@ruflin I could add a new `UPGRADE-8.2.md` where we could add **before/after** sections so users would be able to review it and do necessary changes.

WDYT?

ruflin:
@sidz ++ on the idea.

ruflin:
Remind me, did we do any follow up on this one?

sidz:
I've dropped the head repo as it was forked against 8.x only. PR has been re-created: https://github.com/ruflin/Elastica/pull/2255

Issue #1450, opened 2018-02-06
To array method in BoolQuery changes _params property to empty StdClass


JanCizmar:
Hello,
I found a serious problem in Query\BoolQuery class.
If method toArray is called, then the _params property is changed to StdClass, but the rest of the code uses the _params propetry like array. So if you want to add some attribute after you call toArray, it dies on error: Cannot use object of type stdClass as array.

```
    public function toArray()
    {
        if (empty($this->_params)) {
            $this->_params = new \stdClass();
        }

        return parent::toArray();
    }
```

Please change this behaviour.

Thanks! :)

ruflin:
Could you share some code snippets from your code? I wonder how / why you access _params directly? Or is this within Elastica?

xfoldvar:
Hey.
We are not accessing _params directly. We are just adding params to a query through method addParam($key, $value) (packages/ruflin/elastica/lib/Elastica/Param.php:137). But this fails when we params are empty.

Exception:
Cannot use object of type stdClass as array
Snippet:

`/** @var $query \Elastica\Query\BoolQuery */`
`$query = $this->elasticSearch->getQuery()->getQuery();`
`$query->addParam("filter", ['bool' => $condition->toElasticQuery()]);`

ruflin:
I tried to figure out what this check for the empty _params is there and the creation of the stdClass. Here is is: https://github.com/ruflin/Elastica/issues/817

@xfoldvar Not sure I can follow your code above. So you get the error when you run `addParam`? Assuming `$query` is a BoolQuery object and toArray is not run yet, I would expect this to work?

xfoldvar:
@ruflin Exactly, the exception is thrown while addParam(key,val) method is called on BoolQuery instance. The property _params is set to stdClass and addParam(key,val) is accessing _params as array.

ruflin:
I still don't understand why _params should be stdClass before `toArray` is called. In the above you seem to use a framework around Elastica. Does the `getQuery()` in there run `toArray()`?

enricobono:
For example here:
https://github.com/ruflin/Elastica/blob/master/src/Query/MatchNone.php
on the constructor we set `_params` to a `stdClass()`

ruflin:
@enricobono Thanks for the specific example, this helps. Lets take the MatchNone as an example. The Elasticsearch query in the end must look as following:

```
{
  "query": {
    "match_none": {}
  }
}
```

If we don't set it to stdClass, it would look as following:

```
[...]

PR #2241, opened 2025-02-12
fix: do not mutate BoolQuery instance in toArray method

MrkMrk00:
Change the returned value of the `BoolQuery::toArray` method instead of mutating the instance.

Issue: https://github.com/ruflin/Elastica/issues/1450

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Deprecated
 ### Removed
 ### Fixed
+* `Elastica\Query\BoolQuery::toArray` no longer changes `$this->_params` to \stdClass when empty [#2241](https://github.com/ruflin/Elastica/pull/2241)
 ### Security
 
 ## [8.1.0](https://github.com/ruflin/Elastica/compare/8.0.0...8.1.0)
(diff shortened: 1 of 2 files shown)


ruflin:
Thanks for opening the PR. This looks very promising as all tests pass -> nothing broke :-) Could you also add an entry to the changelog and convert this to ready for review? 

Do you have any concerns on getting this in? Some side effects I'm not thinking of?

MrkMrk00:
Hello, thanks. I wasn't sure whether to add the entry to `Fixes` or `Change`. Let me know if you want it somewhere else.

There's no side-effects that I can think of. Since `_params` is private, the only way I can see that someone would access it is through reflection.

Sorry for the commit from my different (work) account.

ruflin:
Changelog LGTM. No worries about the different users for the commits. I'll squash it all together on merge. Running CI now and if all green, I'll get it in.

Should we also backport this to 7.x?

ruflin:
Thanks for the contribution!

MrkMrk00:
Thank you too!

> Changelog LGTM. No worries about the different users for the commits. I'll squash it all together on merge. Running CI now and if all green, I'll get it in.
> 
> Should we also backport this to 7.x?

I don't see why not. Do you want me to also create a PR to the 7.x branch?

Also the issue (https://github.com/ruflin/Elastica/issues/1450) could probably be closed, since the original problem has been fixed.

ruflin:
> I don't see why not. Do you want me to also create a PR to the 7.x branch?

Would be great. Thank you!

PR #2242, opened 2025-04-18
Add rewrite param setters to all relevant queries

Sharptsa:
I recently had to play with the `rewrite` params on a few queries and only the `Wildcard` and `QueryString` classes have a setter for this param.

This PR adds setters for the `rewrite` params for all relevant the relevant Query classes where it is missing (`Prefix`, `Regexp` and `Fuzzy`), based on the Elasticsearch documentation https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-multi-term-rewrite. As well as adding constants for all the possible values for this param.

## Summary by CodeRabbit

* **New Features**
  * Fuzzy, Prefix, Regexp and Wildcard queries now support configurable multi-term rewrite strategies with dedicated rewrite-mode options.
* **Documentation**
  * Changelog restructured with an added unreleased section and documented backward-compatibility breaks and release notes sections.
* **Tests**
  * Query serialization tests updated to assert rewrite-mode serialization.


src/Query/Fuzzy.php
```diff
@@ -15,6 +15,16 @@
  */
 class Fuzzy extends AbstractQuery
 {
+    /**
+     * Rewrite methods: @see https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-multi-term-rewrite.html.
+     */
+    public const REWRITE_CONSTANT_SCORE = 'constant_score';
```
ruflin, line 21: Are these const always the same. I wonder if we should define these once in AbstractQuery. But then, they don't apply to all queries. What is the pattern of all the queries this applies to? Should we have an abstract query or similar that all these extend?

src/Query/Regexp.php
```diff
@@ -32,8 +49,28 @@ public function __construct(string $key = '', ?string $value = null, float $boos
...
         return $this->setParam($key, ['value' => $value, 'boost' => $boost]);
     }
+
+    /**
+     * Set the method used to rewrite the query.
+     * Use one of the Regexp::REWRITE_* constants, or provide your own.
+     *
+     * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-multi-term-rewrite.html
+     */
+    public function setRewrite(string $rewriteMode): self
```
ruflin, line 65: These methods are always identical?

src/Query/Wildcard.php
```diff
@@ -19,6 +19,9 @@ class Wildcard extends AbstractQuery
     public const REWRITE_CONSTANT_SCORE = 'constant_score';
     public const REWRITE_CONSTANT_SCORE_BOOLEAN = 'constant_score_boolean';
     public const REWRITE_SCORING_BOOLEAN = 'scoring_boolean';
+    public const REWRITE_TOP_TERMS_BLENDED_FREQS_N = 'top_terms_blended_freqs_N';
```
ruflin, line 22: Interesting that we had some of the const already

ruflin (commented):
Overall LGTM. I was wondering if we could abstract out parts but then I also come to the conclusion, that it might not worth the effort and we rather have some duplicated code.

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Backward Compatibility Breaks
 ### Added
+* `Elastica\Query\Fuzzy::setRewrite` rewrite param setter and possible values as constants
```
ruflin: Can you add the PR number here (2242)

ruflin:
For the CI runs, looks like we need do update Ubuntu first, we got the following error message:

> This is a scheduled Ubuntu 20.04 retirement. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. For more details, see https://github.com/actions/runner-images/issues/11101

src/Query/Fuzzy.php
Sharptsa: These constants are always the same. The only pattern between these, is that they are all Term-level queries that use wildcards in some shape or form.

When writing the PR I was also debating puttings the constants in AbstractQuery but reached the same conclusion.

I guess we could have a new level of abstract query for these. Not sure how to name it though.

src/Query/Regexp.php
Sharptsa: It's the same for Prefix and Regexp, also for Wildcard except the null check on `$this->field` since it's set in the constructor.
For Fuzzy I used the `setFieldOption` shortcut since it was already present but it can rewritten the same as the others.

IMHO, these queries could use a refactor in the long run, to harmonize them and have consistent constructors.
Right now, Prefix takes an array, the others a `field` and `value`, but some are nullable and some are not.
It would create breaking changes though.

ruflin:
I opened a PR with a fix here: https://github.com/ruflin/Elastica/pull/2243

ruflin:
I just merged https://github.com/ruflin/Elastica/pull/2243 which should fix the CI issue. Can rebase / update the branch?

ruflin:
This new push slipped through on my end. Any chance you could rebase this on 9.x so we can get this into 9.x and backport it from there to 8.x?

ruflin:
I changed the target of this PR to go against 9.x. We can always backport to 8.x afterwards.

ruflin:
@Sharptsa  I picked this up again and pushed some cleanup / fixes directly to your branch. hope this is ok for you.

PR #2246, opened 2025-05-06
Add elasticsearch 9 support

sidz:
This PR adds elasticsearch 9 support.

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -11,7 +11,7 @@ jobs:
       - name: 'Setup PHP'
         uses: 'shivammathur/setup-php@v2'
         with:
-          php-version: '8.0'
+          php-version: '8.1'
           coverage: 'none'
           extensions: 'json, mbstring, tokenizer'
           tools: 'composer-normalize:2.28.0, php-cs-fixer:3.59.3'
@@ -32,7 +32,6 @@ jobs:
     strategy:
       matrix:
         php:
-          - '8.0'
           - '8.1'
           - '8.2'
           - '8.3'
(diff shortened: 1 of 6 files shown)


ruflin:
Thanks for starting work on this one. Should we take this into a 9.x branch and then backport if possible?

For the error:

```
Elastic\Elasticsearch\Exception\ClientResponseException: 400 Bad Request: {"error":{"root_cause":[{"type":"media_type_header_exception","reason":"Invalid media-type value on headers [Content-Type, Accept]"}],"type":"media_type_header_exception","reason":"Invalid media-type value on headers [Content-Type, Accept]","caused_by":{"type":"status_exception","reason":"Accept version must be either version 8 or 7, but found 9. Accept=application/vnd.elasticsearch+json; compatible-with=9"}},"status":400}
```

It basically means if we use client version 9 we can't use it against version 7 / 8?

sidz:
> Should we take this into a 9.x branch and then backport if possible?

Yup.. Need to create 9.x branch which will support ES >= 9.

> It basically means if we use client version 9 we can't use it against version 7 / 8?

`compatible-with` header was added with ES 8. If you will take a look at the CI [tests for ES 7.17](https://github.com/ruflin/Elastica/actions/runs/14867956890/job/41749329536) you will see that all our tests are passed. So it sounds like elasticsearch-php client could work with ES < 9.
That's just an issue in the ES8 as it doesn't know about ES 9 what is obviously okay.

Ideally, I think, it would be much easier for Elastica to drop support for ES < 9 in the 9.x branch as it matches with Elastica [Versioning & Dependecies](https://github.com/ruflin/Elastica#versions--dependencies) rules.

But probably we could just override `buildCompatibilityHeaders` ([source code](https://github.com/elastic/elasticsearch-php/blob/v9.0.0/src/Traits/EndpointTrait.php#L171)) method and do not send 
`compatible-with` header to ES. That should disable compatibility checks, see https://www.elastic.co/docs/reference/elasticsearch/rest-apis/compatibility.

ruflin:
Will you take care of the 9.x branch or should I? I guess as soon as we have it, this will target the 8.x branch?

sidz:
> Will you take care of the 9.x branch or should I?

Could you please create it @ruflin ? As I'm not able to create branches in your repository without PRs 

> I guess as soon as we have it, this will target the 8.x branch?

Yes. It need to be created from the latest 8.x

ruflin:
Here you go: https://github.com/ruflin/Elastica/tree/9.x I have also updated the repo to make 9.x the default branch.

ruflin:
Probably worth a changelog entry ;-) Should we also mention the PHP version requirement update?

sidz:
> Probably worth a changelog entry ;-) Should we also mention the PHP version requirement update?

done

coreation:
hey @sidz @ruflin any update on this, what needs to happen to merge this into the main repo? @sidz I much appreciate your hard work, kudos!!

ruflin:
Sorry, this slipped through on my end. @coreation thx for the ping.

coreation:
@ruflin thanks so much for keeping this repository alive and kicking for such a long time!

PR #2250, opened 2025-07-21
Fix Index template to use correct API for 9.x

nyroDev:
I updated indexTemplate in order to use the correct `indexTemplate` function to work correctly in 9.x

### DIFF
--- a/src/IndexTemplate.php
+++ b/src/IndexTemplate.php
@@ -61,7 +61,7 @@ class IndexTemplate
     public function delete(): Response
     {
         return $this->_client->toElasticaResponse(
-            $this->_client->indices()->deleteTemplate(['name' => $this->getName()])
+            $this->_client->indices()->deleteIndexTemplate(['name' => $this->getName()])
         );
     }
 
@@ -81,7 +81,7 @@ class IndexTemplate
     public function create(array $args = []): Response
     {
         return $this->_client->toElasticaResponse(
-            $this->_client->indices()->putTemplate(['name' => $this->getName(), 'body' => $args])
+            $this->_client->indices()->putIndexTemplate(['name' => $this->getName(), 'body' => $args])
         );
     }
 
@@ -96,7 +96,7 @@ class IndexTemplate
      */
     public function exists(): bool
     {
-        $response = $this->_client->indices()->existsTemplate(['name' => $this->getName()]);
+        $response = $this->_client->indices()->existsIndexTemplate(['name' => $this->getName()]);
 
         return 200 === $response->getStatusCode();
     }


sidz:
Hi @nyroDev.

Thanks for this PR! Nice catch!

Could we please add tests to cover each case?

nyroDev:
Hi @sidz, I think evyrthing is already covered by [tests/IndexTemplateTest.php](https://github.com/ruflin/Elastica/blob/9.x/tests/IndexTemplateTest.php)

sidz:
Ah okay, I see now. 
The API are just deprecated: https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-indices-delete-template

I was a bit confused by the PR title and description

ruflin:
The problem here is that legacy index templates and composable index templates are not the same. Even though the legacy ones are deprecated, these are still available in Elasticsearch 9 and user must be able to manage these.

The miss here is that we don't support the composable template and we should add the functionality.

nyroDev:
> The problem here is that legacy index templates and composable index templates are not the same. Even though the legacy ones are deprecated, these are still available in Elasticsearch 9 and user must be able to manage these.
> 
> The miss here is that we don't support the composable template and we should add the functionality.

I'm not sure to follow you.

What should be the prupose of `IndexTemplate`?
Write into _index_template or _template?

And depending of the response, what should be the name of the class to write into the other one?

ruflin:
The purpose of IndexTemplate was / is to write to [_template](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-indices-get-template). It existed for a long time and before composable templates came along and changing it to write to _index_template instead would be breaking change because as far as I know, these two are not fully compatible with each other, that is why there is two different endpoints. Now "unfortunately" we already call this "Index template" which sounds like the in 7.9 introduced `_index_template` API endpoint: https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-indices-get-index-template 

If we would just make this change, I would expected to break quite a few deployments. Now the question is how we get around the naming problem. Some options:

* Rename the existing Class to LegacyIndexTemplate. Unfortunately this would be a breaking change
* Add IndexTemplatev2 Class: Not ideal as this is not what it is called in Elasticsearch
* Add ComposableIndexTemplate: That was my initial proposal for the name to call it in Elasticsearch as I was pretty close to the effort but we sticked to just call it Index Template 😇 

As inspiration, the key of the "new" index template is that these are composable with different component templates: https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-cluster-put-component-template In any case, Elastic should support adding component templates too.

@nyroDev As you see, I think our main problem is naming (one of the hardest engineering challenges). As I was too deeply involved with "Composable" Index Templates, it would be great to hear your more outside opinion on what would be a good name. 

The good news is, we have not tagged 9.0 yet, so we can still make breaking change if needed.

nyroDev:
I used ES back in version 2 to 6 in PHP.  
I came back only lately and started with ES 9 to use the last version possible.  

I read documentation about all ES new features, such as [Templates](https://www.elastic.co/docs/manage-data/data-store/templates) (index and component templates)

When trying to implement it with Elastica, I simply used the existing `IndexTemplate` class because it's name match other naming conventions:
`Search` to use `_search` or `Bulk` for `_bulk`

That's only after digging into the code that I found out it was using the `_template` API endpoint.  
And because 9.0 is not already released as stable, I simply created this PR which solved my problem, without any worries about legacy implications...

So, to answer your question, my newbie recommendation will be to be as close as possible to ElasticSearch naming endpoints.
In my point of view, it's easier to understand and found out relation between ES doc and Elastica implementation.

- Template -> _template : will use the code from 8, and we should probably mark it as deprecated too?
- ComponentTemplate -> _component_template : will be new, nothing special here I guess
- IndexTemplate -> _index_template : will be a breaking change as you mentioned

ruflin:
The reason that IndexTemplate is already called IndexTemplate is because historically it was the "Index Template" API endpoint. Overall I like your suggestion to move forward. Code wise this would mean:

* Rename current IndexTemplate class to Template and directly deprecate it
* Introduce a new IndexTemplate class which will heavily overlap with the current but has likely additional methods / new tests like adding component templates
* Introduce component templates

Implementation wise I think we can either pack this into 1 PR (my preference) or split up the rename (breaking change) and follow up with "new" index and component templates. For me index and component templates belong closely together as the composable index templates without component templates would not have the full features set.

@nyroDev Intersted / would you have time to take this one? Fully understand if you say this is out of scope of the initial quick fix you had planned. In any case, appreciate your contributions!

sidz:
Just a thought.

We could do it bit user-friendly and do all what is discussed here in both **8.x** and **9.x** branches as ES 8 is still maintained, i.e:
- _Deprecate_ current IndexTemplate in the 8.x version but leave current behavior to avoid BC breaks.
- Create **IndexTemplate2** (just an example) which would work with the _index_template API
- Introduce Template
- Introduce Component templates

And in the next PR into 9.x we could remove deprecated **IndexTemplate** class and rename **IndexTemplate2** into **IndexTemplate**.

p.s. I understand that it would take more time but at the end everybody win

nyroDev:
> Just a thought.
> 
> We could do it bit user-friendly and do all what is discussed here in both **8.x** and **9.x** branches as ES 8 is still maintained, i.e:
> 
> * _Deprecate_ current IndexTemplate in the 8.x version but leave current behavior to avoid BC breaks.
> * Create **IndexTemplate2** (just an example) which would work with the _index_template API
> * Introduce Template
> * Introduce Component templates
> 
> And in the next PR into 9.x we could remove deprecated **IndexTemplate** class and rename **IndexTemplate2** into **IndexTemplate**.
> 
> p.s. I understand that it would take more time but at the end everybody win

What about :  
- Add an internal boolean `useLegacy` (default to true) to IndexTemplate to indicate if the `_template` or `_index_template` should be used
- If the useLgeacy is set to true, we could use the `Template` class internally
- If the useLgeacy is set to false, use `_index_template` API
- Introduce Template -> it should be depreacated
- Introduce Component templates

When moving to 9.x, we could remove the useLegacy code that was in IndexTemplate and note 2 BC:
- If you used IndexTemplate (without param or with useLegacy=true), you should use Template (depracated)
- If you used IndexTemplate with useLegacy=false, you should simply remove the useLegacy set and keep using IndexTemplate as before

This solution has the benefit to not create a temporary class that will be deleted anyway.
It also have the deprecation feature already in place and no duplication code.
It will add a small overhead on IndexTemplate, but it's only for 8.x branch and on a feature that might not ave been used many times I guess, not bug report about that already...

ruflin:
I like where you both are heading here especially as it provides a more seamless migration path for our users. Two things that bother me:

* Should users already have in 8.x a default way to use the new templates (without having to set the config to true). As I expect less experience users would just keep using the default.
* Code complexity: How much complexity will it create in 8.x?

The nice part about the plan is that we can also split it up and first do the breaking change in 9.x. Introduce the new classes, potentially introduce the legacy flag and if needed backport it to 8.x and change the default.

nyroDev:
> I like where you both are heading here especially as it provides a more seamless migration path for our users. Two things that bother me:
> 
> * Should users already have in 8.x a default way to use the new templates (without having to set the config to true). As I expect less experience users would just keep using the default.
> * Code complexity: How much complexity will it create in 8.x?
> 
> The nice part about the plan is that we can also split it up and first do the breaking change in 9.x. Introduce the new classes, potentially introduce the legacy flag and if needed backport it to 8.x and change the default.

I'm sorry, but I'm not sure to follow you on this comment.
What is your recommandation?
Work on 8.x or only on 9.x?

Regarding your question about a default way to use indexTemplate:
The idea here is to not introduce a BC in 8.x.
If a user wants to use _index_template in 8.x, he will have to use : 
```php
new IndexTemplate($client, $name, false);
```

Regarding your question about code complexity:  
IndexTemplate only have 5 functions (+__construct), and does not extends anything.

the __construct could looks like:
```php
public function __construct(Client $client, $name, bool $useLegacy = true)
    {
        $this->_useLegacy = $useLegacy;
        if ($useLegacy) {
                // @todo trigger a deprecation note here? how should it be done?
                $this->_legacyTemplate = new Template($client, $name);
                return;
        }
        $this->_client = $client;
        if (!\is_scalar($name)) {
            throw new InvalidException('Index template should be a scalar type');
        }
        $this->_name = (string) $name;
    }
```

Each of the functions will have something like:  
```php
if ($this->useLegacy) {
        return $this->_legacyTemplate->callToFunction();
}
```

ruflin:
I have second thoughts about the legacy option in 9.x. Do we really need or is it enough if we introduce the option in 8.x and have people move over to `Template` which will still exist in 9.x? Ideally I would like to keep the 9.x code without conditions to keep it simple. I'm aware, this is slightly different from what I proposed before. It means:

* 9.x
  * IndexTemplate class -> no conditions, no legacy flag
  * Template class -> no conditions
* 8.x
  * IndexTemplate class -> conditions, legacy flag, deprecation notice to move to Template
  * Template class -> no conditions

This also ensures that if we want to exend IndexTemplate in 9.x we don't have to worry about Template compatibility. I expect the new IndexTemplate to also have for example a function addComponentTemplate(s). In case someone we would support `legacy` option, each of these functions would have to throw an exception.

Work wise, I would start with 9.x. If we not do the 8.x part afterwards, it is a bumpy road to 9.x for users but still we have 9.x clean. If we do the work in 8.x, users are able to migrate during the 8.x release and have a smooth upgrade path. Ideally we have both PR's in parallel. 

@nyroDev Looks like you have already most code for 8.x ;-)

nyroDev:
#2257 and #2258 have been created.  
This PR could be closed.

ruflin:
@nyroDev Thank you! Will try to have a look as soon as I get to it!

ruflin:
Closing this PR as the two follow up PRs will solve the problem. @nyroDev I really appreciate your initiative here!

PR #2252, opened 2025-07-23
Update phpstan to 2.x

sidz:
This PR updates phpstan to the 2.1 and resolve some of the issues.

### DIFF
--- a/tests/Bulk/ResponseSetTest.php
+++ b/tests/Bulk/ResponseSetTest.php
@@ -18,9 +18,9 @@ use PHPUnit\Framework\Attributes\Group;
 /**
  * @internal
  */
+#[Group('unit')]
 class ResponseSetTest extends BaseTest
 {
-    #[Group('unit')]
     public function testConstructor(): void
     {
         [$responseData, $actions] = self::_getFixture();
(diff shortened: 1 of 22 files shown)


tests/Bulk/ResponseSetTest.php
```diff
@@ -18,9 +18,9 @@
 /**
  * @internal
  */
+#[Group('unit')]
```
ruflin, line 21: This implies all test in here are unit tests?

ruflin:
Nice update. Should we mention this in the changelog? Not sure TBH.

ruflin (approved):
Approving the change itself. Lets have a quick chat about the changelog entry.

tests/Bulk/ResponseSetTest.php
sidz: yup.

In my projects I usually use different test suites instead of groups. so Unit and Functional tests are placed in the different folders, ie:

tests/Unit
tests/Functional

sidz:
> Nice update. Should we mention this in the changelog? Not sure TBH.

It makes sense @ruflin. I forgot to add a new line into Changelog. 

Changelog updated

ruflin:
Thanks!

PR #2254, opened 2025-07-24
Support PHP 8.5 in 8.x branch

sidz:
This PR adds PHP 8.5 support and test Elastica against all ES 8 versions (added missed)

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -43,15 +43,19 @@ jobs:
         dependencies:
           - 'highest'
         elasticsearch:
-          - '8.18.4'
+          - '8.19.0'
         include:
+          # Test with previous version to support backward compatibility
+          - php: '8.0'
+            elasticsearch: '7.17.28'
+            experimental: false
           # Test with the lowest set of dependencies
           - php: '8.0'
             dependencies: 'lowest'
-            elasticsearch: '8.18.4'
+            elasticsearch: '8.19.0'
             experimental: false
           - php: '8.0'
-            elasticsearch: '8.17.9'
+            elasticsearch: '8.18.4'
             experimental: false
       fail-fast: false
     steps:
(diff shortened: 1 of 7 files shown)


.github/workflows/continuous-integration.yaml
```diff
@@ -75,15 +76,33 @@ jobs:
...
+          - php: '8.1'
+            elasticsearch: '8.14.3' # there are some bc in minor version https://www.elastic.co/guide/en/elasticsearch/reference/current/migrating-8.14.html
+            experimental: false
+          - php: '8.1'
+            elasticsearch: '8.15.5' # there are some bc in minor version https://www.elastic.co/guide/en/elasticsearch/reference/current/migrating-8.15.html
+            experimental: false
+          - php: '8.1'
+            elasticsearch: '8.16.6' # there are some bc in minor version https://www.elastic.co/guide/en/elasticsearch/reference/current/migrating-8.16.html
+            experimental: false
+          - php: '8.1'
```
ruflin: Should we really test against all the older minor elasticsearch versions? I would focus either only on the last minor or the two last minors.

sidz: Fine for me. Changed.

docker/docker-compose.es.yml
```diff
@@ -3,9 +3,7 @@ version: '3.8'
 services:
     es01:
         container_name: es01
-        image: &image docker.elastic.co/elasticsearch/elasticsearch:${ES_VERSION:-8.11.1}
-        command: &command >
```
sidz, line 7: No longer needed. 
> [ingest-attachment] is no longer a plugin but instead a module packaged with this distribution of Elasticsearch

.github/workflows/continuous-integration.yaml
```diff
@@ -37,53 +37,26 @@ jobs:
...
           - false
         dependencies:
           - 'highest'
         elasticsearch:
-          - '8.0.1'
+          - '8.18.4'
         include:
-          # Test with previous version to support backward compatibility
-          - php: '8.0'
-            elasticsearch: '7.17.18'
```
ruflin, line 49: Sorry, my fault. Lets also keep the last version of the previous major.

sidz: @ruflin restored

PR #2255, opened 2025-07-24
Remove unused method argument in the Elastica\SearchableInterface methods

sidz:
This PR removes:
- 3rd $method argument from the Elastica\SearchableInterface::search and Elastica\SearchableInterface::count as they are not unused anymore.
- Elastica\Request class as it is not needed anymore.

On one hand it's a BC break as signatures are changed but on other hand it is not as we basically didn't clean up code base well before releasing version 8.

`$method` argument is not used inside of the methods.
`Elastica\Request` class is not needed anymore as right methods are set inside of official elasticsearch SDK


sidz:
Seems like random fail. Can not re-run the job

ruflin:
I would prefer to get this first into 9.x and then discuss if we should also do it in 8.x. Just my way of working, first get it into main and backport from there. Overall, I'm on board with the change. My only "itch" is that even though the args don't do anything, it could break someone upgrading. Instead, can we just keep it in 8.x and do it in 9.x only?

ruflin:
Should have looked at the changes in the other order first ;-) https://github.com/ruflin/Elastica/pull/2256

ruflin:
@sidz Someone upgrading do the next minor using this and by accident having use the 3 param, their code will break, right?

sidz:
yes @ruflin. That's the reason why we have added [UPGRADE-8.2.md](https://github.com/ruflin/Elastica/pull/2255/files#diff-e022b6a1f157ac44d0354d2715ae09bf33563ec9aae23cfeb5d869f60fed0604) as was discussed in February in the original PR https://github.com/ruflin/Elastica/pull/2240#issuecomment-2661429738

ruflin:
The discussion we had in February was before we had 9.0 out. Now that we have 9.0, I wonder if we should discuss again. If you feel strongly we should do this already in 8.x, happy to move forward.

PR #2256, opened 2025-07-24
Remove unused method argument in the Elastica\SearchableInterface methods in 9.x

sidz:
This PR removes:
- 3rd `$method` argument from the `Elastica\SearchableInterface::search` and `Elastica\SearchableInterface::count` as they are not unused anymore.
- `Elastica\Request` class as it is not needed anymore.

`$method` argument is not used inside of the methods.
Elastica\Request class is not needed anymore as right methods are set inside of official elasticsearch SDK


ruflin (approved):
LGTM

PR #2257, opened 2025-07-25
Add component and legacy Template

nyroDev:
Folowing #2250 

This is the update for 8.x version which **don't** add Breaking change.

When using IndexTemplate, it has the `useLgeacy` set to true.
In this case `IndexTemplate` uses the new `Template` class.

The new `Template` is marked deprecated because [ElasticSearch did so in 7.8 version](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-indices-put-template)

This PR also add a `ComponentTemplate`.

Please note that `exists` function implemented in https://github.com/elastic/elasticsearch-php does not work correctly.
Here is the corresponding issue:
https://github.com/elastic/elasticsearch-php/issues/1458

### DIFF
--- a/src/IndexTemplate.php
+++ b/src/IndexTemplate.php
@@ -86,7 +86,7 @@ class IndexTemplate
         }
 
         return $this->_client->toElasticaResponse(
-            $this->_client->indices()->deleteTemplate(['name' => $this->getName()])
+            $this->_client->indices()->deleteIndexTemplate(['name' => $this->getName()])
         );
     }
 
@@ -110,7 +110,7 @@ class IndexTemplate
         }
 
         return $this->_client->toElasticaResponse(
-            $this->_client->indices()->putTemplate(['name' => $this->getName(), 'body' => $args])
+            $this->_client->indices()->putIndexTemplate(['name' => $this->getName(), 'body' => $args])
         );
     }
 
@@ -128,7 +128,7 @@ class IndexTemplate
         if ($this->_useLegacy) {
             return $this->_legacyTemplate->exists();
         }
-        $response = $this->_client->indices()->existsTemplate(['name' => $this->getName()]);
+        $response = $this->_client->indices()->existsIndexTemplate(['name' => $this->getName()]);
 
         return 200 === $response->getStatusCode();
     }
--- a/src/Template.php
+++ b/src/Template.php
@@ -12,13 +12,14 @@ use Elastica\Exception\ClientException;
 use Elastica\Exception\InvalidException;
 
 /**
- * Elastica index template object.
+ * Elastica legacy index template object.
  *
  * @author Dmitry Balabka <dmitry.balabka@gmail.com>
  *
+ * @deprecated _template have been deprecated in Elasticsearch 7.8 and will be removed in a future version. You should use the IndexTemplate class instead.
  * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-templates.html
  */
-class IndexTemplate
+class Template
 {
     /**
      * Index template name.
@@ -33,35 +34,14 @@ class IndexTemplate
     protected $_client;
 
     /**
-     * Indicates if we should use the legacy template API.
-     *
-     * @var bool
-     */
-    protected $_useLegacy;
-
-    /**
-     * Legacy template object if using the legacy API.
-     *
-     * @var Template
-     */
-    protected $_legacyTemplate;
-
-    /**
      * Creates a new index template object.
      *
      * @param string $name Index template name
      *
      * @throws InvalidException
      */
-    public function __construct(Client $client, $name, $useLegacy = true)
+    public function __construct(Client $client, $name)
     {
-        $this->_useLegacy = $useLegacy;
-        if ($useLegacy) {
-            $this->_legacyTemplate = new Template($client, $name);
-
-            return;
-        }
-
         $this->_client = $client;
 
         if (!\is_scalar($name)) {
@@ -81,10 +61,6 @@ class IndexTemplate
      */
     public function delete(): Response
     {
-        if ($this->_useLegacy) {
-            return $this->_legacyTemplate->delete();
-        }
-
         return $this->_client->toElasticaResponse(
             $this->_client->indices()->deleteTemplate(['name' => $this->getName()])
         );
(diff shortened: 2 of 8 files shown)


src/Template.php
```diff
@@ -0,0 +1,120 @@
...
+use Elastic\Transport\Exception\NoNodeAvailableException;
+use Elastica\Exception\ClientException;
+use Elastica\Exception\InvalidException;
+
+/**
+ * Elastica legacy index template object.
+ *
+ * @author Dmitry Balabka <dmitry.balabka@gmail.com>
+ *
+ * @deprecated _template have been deprecated in Elasticsearch 7.8 and will be removed in a future version. You should use the IndexTemplate class instead.
```
ruflin, line 19: Nit: I think the deprecation happened later that is just when the other templates were introduced but could not find quickly when. So lets keep it.

nyroDev: [Depracted since 7.8](https://www.elastic.co/guide/en/elasticsearch/reference/7.8/release-highlights.html)

> Composable index templates replace the previous version of index templates, which are now deprecated. If an index matches both a composable template and a legacy index template, Elasticsearch uses the composable template.

ruflin: Oh, I was not aware. Thanks for the link.

ruflin:
Still need to go through the PR in details. Could you add an entry to the changelog?

src/IndexTemplate.php
```diff
@@ -60,6 +81,10 @@ public function __construct(Client $client, $name)
      */
     public function delete(): Response
     {
+        if ($this->_useLegacy) {
+            return $this->_legacyTemplate->delete();
+        }
+
         return $this->_client->toElasticaResponse(
             $this->_client->indices()->deleteTemplate(['name' => $this->getName()])
```
ruflin: This should now be `deleteIndexTemplate` ?

nyroDev: Done. Sorry about that, I forgot to redo it on this PR...

nyroDev:
> Still need to go through the PR in details. Could you add an entry to the changelog?

I just pushed it. Not sure if it respect exactly what is needed here.

src/Template.php
```diff
@@ -0,0 +1,120 @@
...
+ * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-templates.html
+ */
+class Template
+{
+    /**
+     * Index template name.
+     *
+     * @var string Index pattern
```
ruflin: This is not the Index pattern but the template name. Ideally we also remove index on line 25

src/Template.php
```diff
@@ -0,0 +1,120 @@
+<?php
```
ruflin, line 1: I assume this is basically a copy of the "old" IndexTemplate file?

ruflin (approved):
Change LGTM

ruflin:
@nyroDev Some of the tests are failing because if I remember correctly the structure of the content of index templates vs legacy index templates is not 100% identical. `settings` and `mappings` are now nested under `template`. Now the question is, should this be something that the lib automatically does for the user?

ruflin:
@nyroDev One suggestion as a follow up:

* 1 PR to introduce Template.php (copy of current IndexTemplate like in 9.x): This allows users to now migrate over
* 1 PR Introduce ComponentTemplate class like in 9.x
* 1 Github issue to follow up discussion on what we do with 8.x and index template. Maybe code is even easier. Is there a "compatibility" way that does not create tons of complexity in the code?

nyroDev:
> * 1 PR to introduce Template.php (copy of current IndexTemplate like in 9.x): This allows users to now migrate over
> * 1 PR Introduce ComponentTemplate class like in 9.x
> * 1 Github issue to follow up discussion on what we do with 8.x and index template. Maybe code is even easier. Is there a "compatibility" way that does not create tons of complexity in the code?

Regarding the first PR "to introduce Template.php (copy of current IndexTemplate like in 9.x): This allows users to now migrate over", to be sure what you have in mind:  
- In 8.x, Template.php will use new API
- In 9.x, Template.php will use the old API (already done with #2274)
- In 8.x, IndexTemplate will use the old API (this is currently the case)
- In 9.x, IndexTemplate.php will use the new API (already done with #2274)

And the Breaking Change from 8.x to 9.x will basically says:
- If you use Template.php in 8.x, use IndexTemplate.php in 9.x
- If you use IndexTemplate.php in 8.x, use Template.php in 9.x (note that is an ES deprecated endpoint)


IMHO, it seems odd to me to introduce Template.php that will go into new API in 8.x and in old API in 9.x.
The current PR seems smoother to me:  
- In 8.x, Template.php will use the old API
- In 9.x, Template.php will use the old API (already done with #2274)
- In 8.x, IndexTemplate will use the old API by default (this is currently the case)
- In 8.x, IndexTemplate will use the new API if you set `$useLegacy=false` (this is what is doing this PR)
- In 9.x, IndexTemplate.php will use the new API (already done with #2274)

And the Breaking Change from 8.x to 9.x will basically says:
- If you use IndexTemplate.php in 8.x without useLegacy=false, use Template.php in 9.x (note that is an ES deprecated endpoint)
- If you use IndexTemplate.php in 8.x with useLegacy=false, remove useLegacy=false

The compatibility code I suggested was to add `bool $useLegacy` in constructor of IndexTemplate, and this is what this PR is currently doing.
The code does not seems to be too complicated for me.

I could remove the ComponentTemplate from this PR, but I'm not sure it adds much more complexity here.

Let me know your thoughts.

ruflin:
We are aligned on the approach, it is as you suggested. I can see how the `(copy of current IndexTemplate like in 9.x):` part was confusing. With current I meant 8.x and "a copy like it is a copy in 9.x" :-) 

The only piece currently missing is I think that the path inside the template changes if `$useLegacy=false`. That is also why the tests are missing.

On having many PR vs 1 larger PR: That is just my personal preference. Both works.

src/Template.php
nyroDev: Yes it is.

nyroDev:
@ruflin I updated the tests and fixed the variable name.
Only test on php8.5 is failing now, but I'm not sure if it's my fault or not.

Please let me know if you need something else here.

ruflin (approved):
I fixed the PHP 8.5 error and now all the tests pass.

ruflin:
@nyroDev Thank you!!!

PR #2258, opened 2025-07-25
Remove legacy index template

nyroDev:
Folowwing #2250 

This PR has to be merged after #2257 and after merged into 9.x branch.
(code from #2257 is merged into this PR)

The only changes that will be done after that is:  
1. Remove the `useLgeacy` parameter for `IndexTemplate`
2. Remove legacy tests for `IndexTemplateTest`
3. Rewrite tests for `IndexTemplateTest` to remove the third false argument

This PR will introduce a Breaking change when instanciating `IndexTemplate`.

Code on 8.x:  
```php
$legacyTemplate = new IndexTemplate($client, $name);
$indexTemplate = new IndexTemplate($client, $name, false);
```
Should become on 9.x:
```php
$legacyTemplate = new Template($client, $name);
$indexTemplate = new IndexTemplate($client, $name, false);
```


tests/IndexTemplateTest.php
```diff
@@ -12,8 +12,6 @@
 /**
  * IndexTemplate class tests.
  *
- * @author Dmitry Balabka <dmitry.balabka@intexsys.lv>
- *
  * @internal
  */
 class IndexTemplateTest extends BaseTest
```
ruflin, line 18: Will we need an addition of add/removeComponentTemplate to this class?

nyroDev: I didn't implement add/remove component template because as far as I understand it will imply to update the `$args` parameter from the `create` function to update the `composed_of` key.  

In this case, I don't know what is expected: 
1. Overwrite the value from `$args`?
2. Merge them (and keeping unique values)?
3. Throw an exception if `composed_of` exists on `$args`?

ruflin: What we have in other places is that we have 3 methods:

* add (append) -> adds one entry with a name
* set (overwrite) -> takes array, overwrite all

It would mean, we store the `$args` locally to be able to modify it. I would skip remove as I'm not sure what the use case here would be.

We can also add this change as a follow up.

ruflin:
Can you add a Changelog entry?

ruflin:
To make sure we are aligned: In 9.x, the third param will not exists anymore. User have to do the migration in 8.x and can't keep using with `false` legacy templates in 9.x. 

```
$indexTemplate = new IndexTemplate($client, $name); // last param does not exist anymore
```

In this PR, shouldn't there also be the changes that you had in your original PR to make sure that indexTemplates methods are called instead of the legacy templates? I might miss something here.

nyroDev:
> To make sure we are aligned: In 9.x, the third param will not exists anymore. User have to do the migration in 8.x and can't keep using with `false` legacy templates in 9.x.
> 
> ```
> $indexTemplate = new IndexTemplate($client, $name); // last param does not exist anymore
> ```
> 
> In this PR, shouldn't there also be the changes that you had in your original PR to make sure that indexTemplates methods are called instead of the legacy templates? I might miss something here.

What I did:
1. Start from 8.x branch for the other PR (changelog is clean on this one)
2. Start from 9.x branch for this PR
3. Merge branch from other PR on this branch
4. Apply the changes to remove the `useLegacy` parameter.

So after merging the other PR into 8.x, and then after merging 8.x into 9.x, this PR should be clean and have only these 2 commits:
- Removing the `useLegacy` https://github.com/nyroDev/Elastica/commit/fa17efe9c037c00dfe257ebb7c2754b7b3685d89
- Updating changelog https://github.com/nyroDev/Elastica/commit/05b820ade992c42680df4504a3ec6f47ba98587d

tests/IndexTemplateTest.php
nyroDev: If you don't mind, I'd like to this in another PR, after this one is merged.

ruflin: SGTM

ruflin:
> and then after merging 8.x into 9.x,

That is not something we usually do. We should rather pick some of the commits instead. Usually we develop against 9.x and then backport to 8.x.

ruflin:
@nyroDev Any chance we could prioritise first the pull request against 9.x and get it in. And then have the detailed discussion around the 8.x magic. Because the structure difference (see other comment) I expect we will have some more discussions there. In summary, lets make all the commits in 9.x independent.

CHANGELOG.md
```diff
@@ -4,18 +4,33 @@ All notable changes to this project will be documented in this file.
...
+* `Elastica\IndexTemplate` now works only with new `_index_template` API [#2258](https://github.com/ruflin/Elastica/pull/2258)
+* Remove `useLegacy` argument from `Elastica\IndexTemplate` ; `Elastica\IndexTemplate` now can only work with new `_index_template` API [#2258](https://github.com/ruflin/Elastica/pull/2258)
+### Added
+### Changed
+### Deprecated
+### Removed
+### Fixed
+### Security
+
 ## [Unreleased](https://github.com/ruflin/Elastica/compare/8.1.0...8.x)
```
ruflin, line 18: @sidz Looks like we missed to update this here as the changes below are against 9.x?

sidz: @ruflin I assume we would change this before we release 9.0

nyroDev:
> @nyroDev Any chance we could prioritise first the pull request against 9.x and get it in. And then have the detailed discussion around the 8.x magic. Because the structure difference (see other comment) I expect we will have some more discussions there. In summary, lets make all the commits in 9.x independent.

I just did here: https://github.com/ruflin/Elastica/pull/2274
This PR could be closed I guess.

ruflin:
Closing as we just merged https://github.com/ruflin/Elastica/pull/2274

PR #2261, opened 2025-07-25
Fix skipped testForcemerge test by enabling delete expunge

ruflin:
The test was skipped because it expected deleted documents to be expunged after forcemerge, but expunge_deletes_allowed was set to 0. Changed the setting to 10 to allow deletion expunge and removed the markTestSkipped call.

🤖 Generated with Claude Code


ruflin:
@claude The tests fail, investigate why.

ruflin:
Closing as fix does not work.

PR #2266, opened 2025-07-27
Fix Elasticsearch cluster discovery configuration to ensure green health status

ruflin:
Update docker-compose.es.yml to include both es01 and es02 nodes in cluster.initial_master_nodes and discovery.seed_hosts settings. This ensures proper cluster formation where both nodes can discover each other, preventing yellow health status that occurs when replicas cannot be properly assigned due to incomplete cluster membership.

Fixes ClusterTest::testGetHealth which expects green status but was failing with yellow.


ruflin:
Merging to see if it will make tests more stable.

PR #2274, opened 2025-08-11
Add component and legacy Template, update IndexTemplate

nyroDev:
Followin https://github.com/ruflin/Elastica/pull/2258#issuecomment-3124558900
this PR targets directly 9.x branch and depends on nothing else.

When using IndexTemplate, ity now uses the ` _index_template` API.
The `Template` class use the old API, like `IndexTemplate was working before this PR.

The new `Template` is marked deprecated because [ElasticSearch did so in 7.8 version](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-indices-put-template)

This PR also add a `ComponentTemplate`.


ruflin (approved):
LGTM

Thanks @nyroDev for getting this over the line. Lets have a follow up discussion now on what we do in 8.x.

PR #2278, opened 2025-09-12
Allow zero retries number

hotrush:
ATM it can be 1 or greater which is a bug because 0 number of retries definitely must be allowed

PR #2282, opened 2025-09-17
Fix: Allow zero retries in Client configuration

ruflin:
- Changed retries condition from `> 0` to `>= 0` to allow zero retries
- Fixed type casting issue by casting retries value to int when passing to setRetries()
- Added comprehensive tests for zero retries functionality
- Updated CHANGELOG.md to document the fix

Fixes #2278


ruflin:
This change is taken from https://github.com/ruflin/Elastica/pull/2278

PR #2293, opened 2025-11-07
Allow numeric `minimum_should_match` values for terms set query

hotrush:
Docs says that `minimum_should_match` field is also supported and it can be numeric or percentage. Added support for numbers.

Ref https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-terms-set-query

### DIFF
--- a/src/Query/TermsSet.php
+++ b/src/Query/TermsSet.php
@@ -15,8 +15,9 @@ class TermsSet extends AbstractQuery
     private string $field;
 
     /**
+     * @param string                       $field
      * @param array<bool|float|int|string> $terms
-     * @param AbstractScript|string        $minimumShouldMatch
+     * @param AbstractScript|string|int    $minimumShouldMatch
      */
     public function __construct(string $field, array $terms, $minimumShouldMatch)
     {
@@ -27,7 +28,9 @@ class TermsSet extends AbstractQuery
         $this->field = $field;
         $this->setTerms($terms);
 
-        if (\is_string($minimumShouldMatch)) {
+        if (\is_int($minimumShouldMatch)) {
+            $this->setMinimumShouldMatch($minimumShouldMatch);
+        } elseif (\is_string($minimumShouldMatch)) {
             $this->setMinimumShouldMatchField($minimumShouldMatch);
         } elseif ($minimumShouldMatch instanceof AbstractScript) {
             $this->setMinimumShouldMatchScript($minimumShouldMatch);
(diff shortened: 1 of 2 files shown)


src/Query/TermsSet.php
```diff
@@ -15,8 +15,9 @@ class TermsSet extends AbstractQuery
     private string $field;
 
     /**
+     * @param string                       $field
```
sidz: Mostly useless IMO and could be removed.

sidz (approved):

src/Query/TermsSet.php
hotrush: removed

src/Query/TermsSet.php
```diff
@@ -27,7 +27,9 @@ public function __construct(string $field, array $terms, $minimumShouldMatch)
         $this->field = $field;
         $this->setTerms($terms);
 
-        if (\is_string($minimumShouldMatch)) {
+        if (\is_int($minimumShouldMatch)) {
+            $this->setMinimumShouldMatch($minimumShouldMatch);
+        } elseif (\is_string($minimumShouldMatch)) {
             $this->setMinimumShouldMatchField($minimumShouldMatch);
         } elseif ($minimumShouldMatch instanceof AbstractScript) {
             $this->setMinimumShouldMatchScript($minimumShouldMatch);
```
ruflin, line 35: The TypeError message should be updated to include `int` in the type list. Currently it says `AbstractScript|string`, but it should say `AbstractScript|int|string` to match the updated PHPDoc and actual supported types.

ruflin (commented):
## Review Summary

This PR adds support for integer values in the `minimum_should_match` parameter for the `TermsSet` query, which aligns with Elasticsearch's documentation. The implementation looks good overall.

### ✅ Positive Aspects

1. **Correct implementation**: The logic properly handles integer values by checking for `int` type first, then falling back to string/script handling
2. **New method added**: `setMinimumShouldMatch(int)` method provides a clean API for setting numeric values
3. **Test coverage**: Unit test follows the project's testing patterns and verifies the correct query structure
4. **PHPDoc updated**: Type hints correctly reflect the new supported type

### ⚠️ Issue Found

1. **TypeError message outdated**: The error message on line 35 still references `AbstractScript|string` but should include `int` to match the updated type signature: `AbstractScript|int|string`

### Suggestions

- Consider adding a functional test similar to `testMinimumShouldMatchFieldSearch()` to verify the numeric value works correctly with Elasticsearch, though the unit test may be sufficient for this simple case.

Overall, this is a solid contribution that correctly implements the feature. Once the TypeError message is updated, this should be ready to merge.

ruflin (approved):
Change LGTM

The Changelog entry is missing but I will follow up with one. A functional tests would also have been nice, but again I'll try to get one in with the changelog.

ruflin:
Follow up PR with changelog is here: https://github.com/ruflin/Elastica/pull/2294

PR #2295, opened 2025-12-15
[6.x] fix PHP8.4 deprecations

JulienRAVIA:
I know this deprecations are handled on superior versions, but in our company we are still on Elastic Search 6.x so we need this version.

### DIFF
--- a/lib/Elastica/Aggregation/BucketSelector.php
+++ b/lib/Elastica/Aggregation/BucketSelector.php
@@ -14,7 +14,7 @@ class BucketSelector extends AbstractSimpleAggregation
      * @param array|null  $bucketsPath
      * @param string|null $script
      */
-    public function __construct(string $name, array $bucketsPath = null, string $script = null)
+    public function __construct(string $name, ?array $bucketsPath = null, ?string $script = null)
     {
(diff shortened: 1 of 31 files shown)


ruflin:
Hi @JulienRAVIA Thanks for the PR. I have learned the hard way, that I better don't get PRs into branches which are not maintained anymore. Even though it looks like probably nothing should break, it still might break something (been there). 

So I wonder what we can do here for you. Do you need the PR merged somewhere in the Elastica repo or can you pull the changes from your repository? If not, we could a branch with something like `6.x-julien` so you could pull form this branch?

JulienRAVIA:
> Hi @JulienRAVIA Thanks for the PR. I have learned the hard way, that I better don't get PRs into branches which are not maintained anymore. Even though it looks like probably nothing should break, it still might break something (been there).
> 
> So I wonder what we can do here for you. Do you need the PR merged somewhere in the Elastica repo or can you pull the changes from your repository? If not, we could a branch with something like `6.x-julien` so you could pull form this branch?

It's nice of you, thanks, but this won't be necessary, we created a fork and we are using the fork for now.
Have a nice day !

PR #2298, opened 2026-03-22
#1645: add support for "search after" pagination

klimov-paul:
Resolves #1645.

Adds support for "search after" pagination and infinite documents scroll.

See: https://www.elastic.co/guide/en/elasticsearch/reference/8.18/paginate-search-results.html#search-after

Covers "8.x" branch.
Should be merged to "9.x" separately, if PR is accepted.

## Summary by CodeRabbit

* **New Features**
  * Introduced search-after pagination support for efficient cursor-based navigation through large result sets.
  * Added methods to iterate through results with automatic pagination handling.

PR #2299, opened 2026-04-06
#1645: add support for "search after" pagination

### LINKED ISSUE #1645: Implement Search After. With an iterator ?
The Query object is missing a setter for the `search_after` param.

https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-search-after.html

I would also like to discuss about adding an iterator to do deep scrolling with `search_after`, based on the model of the `Scroll.php`. Do you think it would be a good addition ?

klimov-paul:
Resolves #1645.

Adds support for "search after" pagination and infinite documents scroll.

See: https://www.elastic.co/guide/en/elasticsearch/reference/8.18/paginate-search-results.html#search-after

Covers "9.x" branch.
Should be ported to "8.x" separately, if PR is accepted.

Migrated from #2298.

## Summary by CodeRabbit

* **New Features**
  * Added "search after" pagination and new iteration APIs to traverse results as individual items or batches with configurable batch sizes.

* **Documentation**
  * Updated changelog to note the new pagination support.

* **Tests**
  * Added functional tests validating search-after pagination, batch/each iteration behavior, ordering, and error conditions for invalid iteration parameters.

### DIFF
--- a/src/Document.php
+++ b/src/Document.php
@@ -235,11 +235,9 @@ class Document extends AbstractUpdateAction
     }
 
     /**
-     * @see \Elastica\Query::setSearchAfter()
-     *
      * Returns the sort position for this document, which can be used as starting offset to search next hits page.
      *
-     * @return array
+     * @see \Elastica\Query::setSearchAfter()
      */
     public function getSort(): array
     {
--- a/src/Document.php
+++ b/src/Document.php
@@ -235,6 +235,18 @@ class Document extends AbstractUpdateAction
     }
 
     /**
+     * @see \Elastica\Query::setSearchAfter()
+     *
+     * Returns the sort position for this document, which can be used as starting offset to search next hits page.
+     *
+     * @return array
+     */
+    public function getSort(): array
+    {
+        return $this->getParam('sort');
+    }
+
+    /**
      * Returns the document as an array.
      */
     public function toArray(): array
--- a/src/Index.php
+++ b/src/Index.php
@@ -551,17 +551,7 @@ class Index implements SearchableInterface
      */
     public function each($query = '', int $batchSize = 100, ?array $options = null): \Generator
     {
-        if ($batchSize < 1) {
-            throw new InvalidException('Batch size must be greater than 0.');
-        }
-
-        $query = clone Query::create($query);
-
-        if (!$query->hasParam('sort')) {
-            throw new InvalidException('Query must have "sort" parameter in order to use "search after" based iteration.');
-        }
-
-        $query->setSize($batchSize);
+        $query = $this->prepareSearchAfterIteratorQuery($query, $batchSize);
 
         while (true) {
             $resultSet = $this->search($query, $options);
@@ -589,17 +579,7 @@ class Index implements SearchableInterface
      */
     public function batch($query = '', int $batchSize = 100, ?array $options = null): \Generator
     {
-        if ($batchSize < 1) {
-            throw new InvalidException('Batch size must be greater than 0.');
-        }
-
-        $query = clone Query::create($query);
-
-        if (!$query->hasParam('sort')) {
-            throw new InvalidException('Query must have "sort" parameter in order to use "search after" based iteration.');
-        }
-
-        $query->setSize($batchSize);
+        $query = $this->prepareSearchAfterIteratorQuery($query, $batchSize);
 
         while (true) {
             $resultSet = $this->search($query, $options);
@@ -619,6 +599,27 @@ class Index implements SearchableInterface
 
             $query->setSearchAfter($lastDocument->getSort());
         }
+    }
+
+    private function prepareSearchAfterIteratorQuery(mixed $query, int $batchSize): Query
+    {
+        if ($batchSize < 1) {
+            throw new InvalidException('Batch size must be greater than 0.');
+        }
+
+        $query = clone Query::create($query); // if original query is object - keep it intact
+
+        if (!$query->hasParam('sort')) {
+            throw new InvalidException('Query must have "sort" parameter in order to use "search after" based iteration.');
+        }
+
+        if ($query->hasParam('from') && 0 !== $query->getParam('from')) {
+            throw new InvalidException('Query must not specify "from" parameter in order to use "search after" based iteration.');
+        }
+
+        $query->setSize($batchSize);
+
+        return $query;
     }
 
     /**
--- a/src/Index.php
+++ b/src/Index.php
@@ -540,6 +540,88 @@ class Index implements SearchableInterface
     }
 
     /**
+     * Iterates over all documents, matching given query, using "search after" based pagination.
+     *
+     * @see \Elastica\Query::setSearchAfter()
+     *
+     * @param mixed $query search query with sort by unique document field.
+     * @param int $batchSize the number of rows to be returned in each batch (e.g. each query size).
+     * @param array|null $options search request options.
+     * @return \Generator|\Elastica\Document[] list of all documents matched the given query as iterator.
+     */
+    public function each($query = '', int $batchSize = 100, ?array $options = null): \Generator
+    {
+        if ($batchSize < 1) {
+            throw new InvalidException('Batch size must be greater than 0.');
+        }
+
+        $query = clone Query::create($query);
+
+        if (!$query->hasParam('sort')) {
+            throw new InvalidException('Query must have "sort" parameter in order to use "search after" based iteration.');
+        }
+
+        $query->setSize($batchSize);
+
+        while (true) {
+            $resultSet = $this->search($query, $options);
+            foreach ($resultSet->getDocuments() as $document) {
+                yield $document;
+            }
+
+            if (count($resultSet->getDocuments()) < $batchSize) {
+                break;
+            }
+
+            $query->setSearchAfter($document->getSort());
+        }
+    }
+
+    /**
+     * Iterates over all documents in batches, matching given query, using "search after" based pagination.
+     *
+     * @see \Elastica\Query::setSearchAfter()
+     *
+     * @param mixed $query search query with sort by unique document field.
+     * @param int $batchSize the number of rows to be returned in each batch (e.g. each query size).
+     * @param array|null $options search request options.
+     * @return \Generator|\Elastica\Document[][] list of document batches matched the given query as iterator.
+     */
+    public function batch($query = '', int $batchSize = 100, ?array $options = null): \Generator
+    {
+        if ($batchSize < 1) {
+            throw new InvalidException('Batch size must be greater than 0.');
+        }
+
+        $query = clone Query::create($query);
+
+        if (!$query->hasParam('sort')) {
+            throw new InvalidException('Query must have "sort" parameter in order to use "search after" based iteration.');
+        }
+
+        $query->setSize($batchSize);
+
+        while (true) {
+            $resultSet = $this->search($query, $options);
+
+            $documents = $resultSet->getDocuments();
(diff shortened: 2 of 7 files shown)


src/Index.php
```diff
@@ -539,6 +539,89 @@ public function count($query = ''): int
...
+                yield $document;
+            }
+
+            if (count($resultSet->getDocuments()) < $batchSize) {
+                break;
+            }
+
+            $query->setSearchAfter($document->getSort());
+        }
+    }
```
ruflin, line 574: Fixes the PHPStan failure (`Variable $document might not be defined` at line 566), the two `\count` cs-fixer findings, and the duplicate `getDocuments()` call. Also reorders the PHPDoc to match the project convention (summary first, aligned `@param`s, blank line, `@see`, blank line, `@return` without a trailing period).

```suggestion
    /**
     * Iterates over all documents, matching given query, using "search after" based pagination.
     *
     * @param mixed      $query     search query with sort by unique document field
     * @param int        $batchSize the number of rows to be returned in each batch (e.g. each query size).
     * @param array|null $options   search request options
     *
     * @see \Elastica\Query::setSearchAfter()
     *
     * @return Document[]|\Generator list of all documents matched the given query as iterator
     */
    public function each($query = '', int $batchSize = 100, ?array $options = null): \Generator
    {
        $query = $this->prepareSearchAfterIteratorQuery($query, $batchSize);

        while (true) {
            $documents = $this->search($query, $options)->getDocuments();
            $count = \count($documents);
            if (0 === $count) {
                break;
            }

            foreach ($documents as $document) {
                yield $document;
            }

            if ($count < $batchSize) {
                break;
            }

            $query->setSearchAfter($documents[$count - 1]->getSort());
        }
    }
```

src/Index.php
```diff
@@ -539,6 +539,89 @@ public function count($query = ''): int
...
+
+            if (count($documents) < $batchSize) {
+                break;
+            }
+
+            $lastDocument = array_pop($documents);
+
+            $query->setSearchAfter($lastDocument->getSort());
+        }
+    }
```
ruflin, line 606: Same PHPDoc reordering, plus `\count(...)` for the cs-fixer rule, and replacing `array_pop()` (which mutates the just-yielded array) with an explicit last-index access. Functionally identical.

```suggestion
    /**
     * Iterates over all documents in batches, matching given query, using "search after" based pagination.
     *
     * @param mixed      $query     search query with sort by unique document field
     * @param int        $batchSize the number of rows to be returned in each batch (e.g. each query size).
     * @param array|null $options   search request options
     *
     * @see \Elastica\Query::setSearchAfter()
     *
     * @return Document[][]|\Generator list of document batches matched the given query as iterator
     */
    public function batch($query = '', int $batchSize = 100, ?array $options = null): \Generator
    {
        $query = $this->prepareSearchAfterIteratorQuery($query, $batchSize);

        while (true) {
            $documents = $this->search($query, $options)->getDocuments();
            $count = \count($documents);
            if (0 === $count) {
                break;
            }

            yield $documents;

            if ($count < $batchSize) {
                break;
            }

            $query->setSearchAfter($documents[$count - 1]->getSort());
        }
    }
```

src/Document.php
```diff
@@ -234,6 +234,18 @@ public function hasPipeline(): bool
...
+     * @see \Elastica\Query::setSearchAfter()
+     *
+     * Returns the sort position for this document, which can be used as starting offset to search next hits page.
+     *
+     * @return array
+     */
+    public function getSort(): array
+    {
+        return $this->getParam('sort');
+    }
```
ruflin, line 245: `@return array` is redundant with the `: array` return type and the cs-fixer flags it. Putting `@see` after the summary also matches the project's PHPDoc convention.

```suggestion
    /**
     * Returns the sort position for this document, which can be used as starting offset to search next hits page.
     *
     * @see \Elastica\Query::setSearchAfter()
     */
    public function getSort(): array
    {
        return $this->getParam('sort');
    }
```

src/Query.php
```diff
@@ -476,4 +476,20 @@ public function setTrackTotalHits($trackTotalHits = true): self
...
+     *
+     * @param array $searchAfter the sort of the last document from previous search result set.
+     * @return static self reference.
+     */
+    public function setSearchAfter(array $searchAfter): self
+    {
+        $this->setParam('search_after', $searchAfter);
+
+        return $this;
+    }
```
ruflin, line 496: Adds the missing blank line between `@param` and `@return`, removes the trailing periods, and reorders so the summary comes first (project convention: summary, blank, body text, blank, `@see`, blank, `@param`, blank, `@return`).

```suggestion
    /**
     * Allows retrieval of the next page of hits using sort values from previous page.
     *
     * The value of {@see \Elastica\Document::getSort()} should be passed as argument here.
     *
     * @see https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html#search-after
     *
     * @param array $searchAfter the sort of the last document from previous search result set
     *
     * @return static self reference
     */
    public function setSearchAfter(array $searchAfter): self
    {
        $this->setParam('search_after', $searchAfter);

        return $this;
    }
```

tests/QueryTest.php
```diff
@@ -621,4 +621,44 @@ public function testSetTrackTotalHits(): void
...
+        $lastDocument->getParam('sort');
+
+        $this->assertNotEmpty($lastDocument->getSort());
+
+        $query->setSearchAfter($lastDocument->getSort());
+
+        $secondPageResultSet = $index->search($query);
+        $documents = $secondPageResultSet->getDocuments();
+        $this->assertCount(1, $documents);
+    }
```
ruflin, line 663: Two small tightenings:
1. `\array_pop(...)` for the cs-fixer rule.
2. Drop the no-op `$lastDocument->getParam('sort');` (return value discarded).
3. Assert the second-page document's id so the test actually verifies pagination continuity instead of just count.

With `country_id desc, region_id asc` over docs `1`(1,1), `2`(1,2), `3`(2,3): page 1 returns `['3', '1']`, so search_after on `1`'s sort should yield `'2'` next.

```suggestion
        /** @var Document $lastDocument */
        $lastDocument = \array_pop($documents);

        $this->assertNotEmpty($lastDocument->getSort());

        $query->setSearchAfter($lastDocument->getSort());

        $secondPageResultSet = $index->search($query);
        $documents = $secondPageResultSet->getDocuments();
        $this->assertCount(1, $documents);
        $this->assertSame('2', $documents[0]->getId());
    }
```

tests/IndexTest.php
```diff
@@ -856,4 +857,66 @@ public function testGetEmptyAliases(): void
...
+
+        $documentIds = [];
+        foreach ($index->batch($query, 2) as $documents) {
+            foreach ($documents as $document) {
+                $documentIds[] = $document->getId();
+            }
+        }
+
+        $this->assertEquals(['3', '2', '1'], $documentIds);
+    }
```
ruflin, line 922: Optional but recommended: cover the three validation branches in `prepareSearchAfterIteratorQuery()` so we don't regress the contract.

```php
#[Group('functional')]
public function testIterateEachThrowsWhenSortIsMissing(): void
{
    $index = $this->_createIndex();

    $this->expectException(InvalidException::class);
    $this->expectExceptionMessageMatches('/sort/');

    iterator_to_array($index->each(new Query(), 10));
}

#[Group('functional')]
public function testIterateEachThrowsWhenFromIsNonZero(): void
{
    $index = $this->_createIndex();

    $query = new Query();
    $query->setSort(['_id' => 'asc']);
    $query->setFrom(5);

    $this->expectException(InvalidException::class);
    $this->expectExceptionMessageMatches('/from/');

    iterator_to_array($index->each($query, 10));
}

#[Group('functional')]
public function testIterateEachThrowsWhenBatchSizeIsZero(): void
{
    $index = $this->_createIndex();

    $query = new Query();
    $query->setSort(['_id' => 'asc']);

    $this->expectException(InvalidException::class);
    $this->expectExceptionMessageMatches('/[Bb]atch size/');

    iterator_to_array($index->each($query, 0));
}
```

The same three should also be added for `batch()`. (Add `use Elastica\Exception\InvalidException;` at the top of the file.)

ruflin (changes requested):
Thanks for tackling this long-standing request — the design is right (generators, query cloning, validation) and `each()` / `batch()` are nice ergonomics on top of `search_after`.

Before merging I'd like to get the two failing CI jobs green and tighten the new code/tests. The PHPStan failure is real, the `php-cs-fixer` failures are mechanical, and the test gaps are easy to close.

## Summary of CI failures

### 1. PHPStan — `Variable $document might not be defined` (`src/Index.php#566`)

In `Index::each()`, `$document` is the `foreach` loop variable and is read after the loop. PHPStan correctly flags that it can be undefined when the page is empty. This also calls `getDocuments()` twice per iteration, which rebuilds the documents from the response each time. The fix below (caching the array, using the last index) addresses both and matches the shape of `batch()`.

### 2. PHPStan — stale baseline entry for `tests/Bulk/Action/UpdateDocumentTest.php`

```
Ignored error pattern #^Parameter \#1 \$expected of method PHPUnit\\Framework\\Assert\:\:assertSame\(\) contains unresolvable type\.$# ... was not matched in reported errors
```

This is **not introduced by this PR** — `9.x` is green because the underlying error happens to still be reported there, but adding `Document::getSort()` apparently changes inference enough that PHPStan no longer reports it on this branch. Easiest fix: remove the now-obsolete entry from `phpstan-baseline.neon` (the `argument.unresolvableType` block for `tests/Bulk/Action/UpdateDocumentTest.php`). Happy to do this in a follow-up if you'd prefer to keep this PR scoped.

### 3. `php-cs-fixer` — missing leading backslashes and PHPDoc ordering

The fixer wants:
- `\count(...)` and `\array_pop(...)` for the new function calls (project enforces `native_function_invocation` with `@all`).
- PHPDoc blocks reordered: summary first, then `@param`s (aligned), blank line, `@see`, blank line, `@return`. Trailing periods on `@param`/`@return` text removed. `@return array` removed when redundant with the return type (Document::getSort).
- Generator `@return` should list types as `Document[]|\Generator` (alpha-ish, classes first per the repo's existing pattern).

All of those are addressed in the inline suggestions below.

## Other (non-CI) suggestions

- `Index::batch()`: `array_pop()` mutates the array unnecessarily — using the last index is clearer. (Suggestion inline.)
- `tests/QueryTest.php::testSearchAfter()`: drop the no-op `$lastDocument->getParam('sort');` and assert the second-page document's id to actually verify continuity rather than just count. (Suggestion inline.)
- Test coverage misses the three validation branches in `prepareSearchAfterIteratorQuery()`. Adding three small tests (no `sort`, non-zero `from`, `batchSize < 1`) would lock the contract in. Not blocking, but recommended.

CodeRabbit suggested adding guards in `Document::getSort()` and `Query::setSearchAfter()`. I'd skip those — `Param::getParam()` already throws `InvalidException` when the key is missing, and Elasticsearch already rejects empty `search_after`. Not worth the extra surface.

Happy to land this once the two CI jobs are green.

ruflin:
@klimov-paul Sorry about the super slow round trip time. I have directly pushed a commit to your branch with the CI fixes and small improvements. So you can ignore all the code comments. Hope that is ok for you.

ruflin (approved):

ruflin:
@klimov-paul Thank you for the contribution! Should we backport this to 8.x as proposed before? Do you want to tackle this directly?

klimov-paul:
Backport to 8.x would be nice, but I have to leave it in your hands.

PR #2314, opened 2026-05-18
feat: add top-level kNN search via Query\Knn and Query::setKnn()

Cryde:
Adds an Elastica\Query\Knn class wrapping Elasticsearch's top-level `knn`
search clause, plus Query::setKnn() to attach a single Knn or a list of
Knn (multiple kNN searches in a single request) as a sibling of `query`.

- Knn supports field, query_vector, k, num_candidates plus optional
  pre-filters (addFilter), similarity threshold and boost.
- Query::toArray() no longer auto-pads with match_all when only knn is set.

## Summary by CodeRabbit

* **New Features**
  * Top-level kNN (k-nearest neighbors) query support for vector similarity searches.
  * Support for single or multiple top-level kNN queries.
  * Options to apply pre-filters, set similarity threshold, and apply boost weights to kNN queries.

* **Tests**
  * Added unit and functional tests covering kNN serialization, filter integration, multi-kNN handling, and end-to-end vector search.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ruflin/Elastica/pull/2314?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/Query/Knn.php
+++ b/src/Query/Knn.php
@@ -1,0 +1,55 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Elastica\Query;
+
+use Elastica\Param;
+
+/**
+ * Top-level kNN search.
+ *
+ * Note: `knn` is a sibling of `query` in the search request body, not a clause inside it.
+ * Attach it via {@see \Elastica\Query::setKnn()} rather than putting it inside a BoolQuery.
+ *
+ * @see https://www.elastic.co/docs/solutions/search/vector/knn
+ */
+class Knn extends Param
+{
+    /**
+     * @param float[] $queryVector
+     */
+    public function __construct(string $field, array $queryVector, int $k, int $numCandidates)
+    {
+        $this->setParam('field', $field);
+        $this->setParam('query_vector', $queryVector);
+        $this->setParam('k', $k);
+        $this->setParam('num_candidates', $numCandidates);
+    }
+
+    /**
+     * Adds a Query DSL filter applied before the kNN search.
+     *
+     * Filters are ANDed together by Elasticsearch.
+     */
+    public function addFilter(AbstractQuery $filter): self
+    {
+        return $this->addParam('filter', $filter);
+    }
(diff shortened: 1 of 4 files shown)


src/Query/Knn.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Elastica\Query;
+
+/**
+ * Top-level kNN search.
+ *
+ * Note: `knn` is a sibling of `query` in the search request body, not a clause inside it.
```
ruflin, line 12: That is the part I stumble over. Should we make it part of Query or make it it's own top level thing extending Parms instead?

src/Query/Knn.php
```diff
@@ -0,0 +1,53 @@
...
+ * Attach it via {@see \Elastica\Query::setKnn()} rather than putting it inside a BoolQuery.
+ *
+ * @see https://www.elastic.co/docs/solutions/search/vector/knn
+ */
+class Knn extends AbstractQuery
```
ruflin: see comment above

ruflin (commented):
Thanks for the contribution. What is your take on the KNN being it's own object not extending Query (as it isn't one 🤔 )

src/Query/Knn.php
Cryde: You're right, knn is a sibling of query in the search body.

Nothing today prevents a user from passing a Knn to Query::setQuery() or inside a BoolQuery, where it would silently produce an invalid request.

I'll switch to extending Param indeed. Pushing the change shortly !

Cryde:
Also maybe I should move to `Elastica\Knn` rather than `Elastica\Query\Knn` ?

ruflin:
Argh, I missed the ping.

> Also maybe I should move to Elastica\Knn rather than Elastica\Query\Knn ?

I was thinking the same. i would like to limit the high level objects but likely makes more sense.

